PR #13, opened 2020-06-23
Documentation and unit tests for Variable

adisr04:
- Update Variable documentation and fix minor semantic errors
- Fix typo in _energy function in potential.py
- Add all unit tests for Variable in test_core

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -57,6 +57,8 @@ class test_Variable(unittest.TestCase):
 
     def test_init(self):
         """Test construction with different bounds."""
+
+        #test with no bounds and non-default value of `const`
         v = relentless.core.Variable(value=1.0, const=True)
         self.assertAlmostEqual(v.value, 1.0)
         self.assertEqual(v.const, True)
@@ -66,6 +68,7 @@ class test_Variable(unittest.TestCase):
         self.assertEqual(v.atlow(), False)
         self.assertEqual(v.athigh(), False)
 
+        #test in between low and high bounds
         v = relentless.core.Variable(value=1.2, low=0.0, high=2.0)
         self.assertAlmostEqual(v.value, 1.2)
         self.assertAlmostEqual(v.low, 0.0)
@@ -74,6 +77,7 @@ class test_Variable(unittest.TestCase):
         self.assertEqual(v.atlow(), False)
         self.assertEqual(v.athigh(), False)
 
+        #test below low bound
         v = relentless.core.Variable(value=-1, low=0.5)
         self.assertAlmostEqual(v.value, 0.5)
         self.assertAlmostEqual(v.low, 0.5)
@@ -82,6 +86,7 @@ class test_Variable(unittest.TestCase):
         self.assertEqual(v.atlow(), True)
         self.assertEqual(v.athigh(), False)
 
+        #test above high bound
         v = relentless.core.Variable(value=2.2, high=2.0)
         self.assertAlmostEqual(v.value, 2.0)
         self.assertEqual(v.high, 2.0)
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -57,12 +59,78 @@ class test_Variable(unittest.TestCase):
...
+        self.assertEqual(v.const, True)
+
+        v = relentless.core.Variable(value=1.2, low=0.0, high=2.0)
+        self.assertAlmostEqual(v.value, 1.2)
+        self.assertAlmostEqual(v.low, 0.0)
+        self.assertAlmostEqual(v.high, 2.0)
+
+        v = relentless.core.Variable(value=-1.2, low=0.5)
+        self.assertAlmostEqual(v.value, 0.5)
+        self.assertAlmostEqual(v.low, 0.5)
```
mphoward, line 73: Could you also test the `isfree`, `atlow`, and `athigh` methods on all these variables?

tests/test_core.py
```diff
@@ -57,12 +59,78 @@ class test_Variable(unittest.TestCase):
...
+
+        v = relentless.core.Variable(value=2.2, high=2.0)
+        self.assertAlmostEqual(v.value, 2.0)
+        self.assertAlmostEqual(v.high, 2.0)
+
+        with self.assertRaises(ValueError):
+            v = relentless.core.Variable(value=4)
```
mphoward, line 80: This test is totally correct but indicates bad design on my part. Even though we think of these as being floats, the user will sometimes type an int (e.g., `4`) instead of a float (`4.0`), and we should allow both. Let's fix it in the Variable class and convert this to a test that passes using an int.

We still want to restrict the types of the variables to those we know how to work with, so in the `isinstance` call inside `value`, check `(float,int)` instead of just `float`. This should cover the two main data types. Please also update the documentation for `__init__` to allow float or int.

You can still keep a test that checks if this raises, but use some silly thing like a `str` and/or a tuple.

tests/test_core.py
```diff
@@ -57,12 +59,78 @@ class test_Variable(unittest.TestCase):
...
 
     def test_clamp(self):
         """Test methods for clamping values with bounds."""
-        raise NotImplementedError()
+        v = relentless.core.Variable(value=0.0, low=2.0)
+
+        x = v.clamp(1.0)
```
mphoward, line 86: If you want to make your tests a little more readable, you can catch both values in variables like:
```py
val,state = v.clamp(1.0)
```

tests/test_core.py
```diff
@@ -57,12 +59,78 @@ class test_Variable(unittest.TestCase):
...
+
+        v.value = -1.5
+        self.assertAlmostEqual(v.value, -1.0)
+        self.assertEqual(v.state, v.State.LOW)
+
+        v.value = 3.0
+        self.assertAlmostEqual(v.value, 1.0)
+        self.assertEqual(v.state, v.State.HIGH)
+
+        with self.assertRaises(ValueError):
```
mphoward, line 132: See my comment above, let's convert this to a test that will pass.

tests/test_core.py
```diff
@@ -57,12 +57,94 @@ class test_Variable(unittest.TestCase):
...
+        self.assertEqual(v.atlow(), False)
+        self.assertEqual(v.athigh(), True)
+
+        with self.assertRaises(ValueError):
+            v = relentless.core.Variable(value='4')
 
     def test_clamp(self):
         """Test methods for clamping values with bounds."""
-        raise NotImplementedError()
+        v = relentless.core.Variable(value=0.0, low=2.0)
```
mphoward, line 98: To make these tests easier to read later, can you add a comment before each block of tests? Like
```py
# clamp with lower bound
...

# clamp with lower and upper bound
...

# clamp with no bounds
```

PR #14, opened 2020-06-24
Documentation and unit tests for Interpolator

adisr04:
- Update Interpolator documentation
- Add all unit tests for Interpolator in test_core

### DIFF
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -15,10 +15,19 @@ class Interpolator:
 
     Parameters
     ----------
-    x : array_like
+    x : float or array_like
         1-d array of x coordinates that must be continually increasing.
     y : array_like
         1-d array of y coordinates.
+
+    Raises
+    ------
+    ValueError
+        If x and y are scalars
+    ValueError
+        If x and y are not 1-dimensional
+    ValueError
+        If x is not strictly increasing
 
     Examples
     --------
@@ -40,6 +49,13 @@ class Interpolator:
 
     """
     def __init__(self, x, y):
+        x = np.atleast_1d(x)
+        if isinstance(x, (float,int)) or isinstance(y, (float,int)):
+            raise ValueError('x and y must not be scalars')
+        if isinstance(x[0], (tuple,list,np.ndarray)) or isinstance(y[0], (tuple,list,np.ndarray)):
+            raise ValueError('x and y must be 1-dimensional')
+        if not np.all(np.diff(x) > 0):
+            raise ValueError('x must be strictly increasing')
         self._domain = (x[0],x[-1])
         self._spline = scipy.interpolate.Akima1DInterpolator(x=x, y=y)
 
@@ -48,10 +64,8 @@ class Interpolator:
 
         Parameters
         ----------
-        x : array_like
+        x : float or array_like
             1-d array of x coordinates to evaluate.
-
-            If `x` is a scalar, it is promoted to a NumPy array.
 
         Returns
         -------
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -82,7 +82,7 @@ class Interpolator:
 
     @property
     def domain(self):
-        """tuple The valid domain for interpolation."""
+        """tuple: The valid domain for interpolation."""
         return self._domain
 
 class PairMatrix:
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -53,9 +53,9 @@ class Interpolator:
     def __init__(self, x, y):
         x = np.atleast_1d(x)
         y = np.atleast_1d(y)
-        if x.ndim > 1:
-            raise ValueError('x cannot be a scalar')
         if x.shape[0] == 1:
+            raise ValueError('x cannot be a scalar')
+        if x.ndim > 1:
             raise ValueError('x must be 1-dimensional')
         if x.shape != y.shape:
             raise ValueError('x and y must be the same shape')
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -15,7 +15,7 @@ class Interpolator:
 
     Parameters
     ----------
-    x : float or array_like
+    x : array_like
         1-d array of x coordinates that must be continually increasing.
     y : array_like
         1-d array of y coordinates.
@@ -23,9 +23,11 @@ class Interpolator:
     Raises
     ------
     ValueError
-        If x and y are scalars
+        If x is a scalar
     ValueError
-        If x and y are not 1-dimensional
+        If x is not 1-dimensional
+    ValueError
+        If y is not the same shape as x
     ValueError
         If x is not strictly increasing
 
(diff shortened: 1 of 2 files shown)


relentless/core.py
```diff
@@ -82,7 +82,7 @@ def __call__(self, x):
```
mphoward, line 82: I'm tagging a few other doc changes for lines above:
- [ ] Add explicit test in `__init__` that x is increasing before setting the domain. You can use something like:
```py
x = np.atleast_1d(x)
if not np.all(x[1:] > x[:-1]):
    raise ValueError('x must be increasing')
```
You might also want to check that x is only a 1d array using `len(x.shape) == 1`.
- [ ] Update the parameters for `__init__` to document that x can be a float or `array_like`
- [ ] Update the parameters for `__call__` to document that x can be a float or `array_like`. Remove comment about being promoted to an array.

tests/test_core.py
```diff
@@ -8,15 +10,52 @@ class test_Interpolator(unittest.TestCase):
 
     def test_init(self):
         """Test creation from data."""
-        raise NotImplementedError()
+
+        #test construction with tuple input
+        f = relentless.core.Interpolator(x=(-1,0,1), y=(-2,0,2))
+        self.assertEqual(f._domain, (-1,1))
+        self.assertIsInstance(f._spline, Akima1DInterpolator)
```
mphoward, line 17: This is a good thought, but we don't want to be this specific about what the exact implementation of `_spline` is, in case it changes in future. We jalso ust want to test the exposed interface. So, you should check `f.domain` rather than `f._domain`, and not check the instance of `_spline`.

tests/test_core.py
```diff
@@ -8,15 +10,52 @@ class test_Interpolator(unittest.TestCase):
...
+        #test construction with list input
+        f = relentless.core.Interpolator(x=[-1,0,1], y=[-2,0,2])
+        self.assertEqual(f._domain, (-1,1))
+        self.assertIsInstance(f._spline, Akima1DInterpolator)
+
+        #test construction with numpy array input
+        f = relentless.core.Interpolator(x=np.array([-1,0,1]), y=np.array([-2,0,2]))
+        self.assertEqual(f._domain, (-1,1))
+        self.assertIsInstance(f._spline, Akima1DInterpolator)
+
```
mphoward, line 28: Can you add a test for mixed input, like a list for x and a tuple for y?

tests/test_core.py
```diff
@@ -8,15 +10,52 @@ class test_Interpolator(unittest.TestCase):
...
+        #test construction with non-increasing domain
+        with self.assertRaises(ValueError):
+            f = relentless.core.Interpolator(x=(1,0,-1), y=(2,0,-2))
```
mphoward, line 31: Good idea. Are there other tests that might fail? What about if x and y are only scalars? (Not actually sure what happens there. If it works, then no need to test this.)

tests/test_core.py
```diff
@@ -8,15 +10,52 @@ class test_Interpolator(unittest.TestCase):
...
         """Test calls, both scalar and array."""
-        raise NotImplementedError()
+        f = relentless.core.Interpolator(x=(-1,0,1), y=(-2,0,2))
+
+        #test scalar call
+        self.assertAlmostEqual(f(-0.5), -1.0)
+        self.assertAlmostEqual(f(0.5), 1.0)
+
+        #test array call
+        np.testing.assert_almost_equal(f([-0.5,0.5]), [-1.0,1.0])
```
mphoward, line 42: NumPy recommends using `assert_allclose` instead of this assertion, so please replace throughout.

relentless/core.py
```diff
@@ -15,11 +15,20 @@ class Interpolator:
 
     Parameters
     ----------
-    x : array_like
+    x : float or array_like
```
mphoward, line 18: Sorry, this is my fault. Clearly the interpolation fails if x is a scalar, so revert this to `array_like`.

relentless/core.py
```diff
@@ -40,6 +49,13 @@ class Interpolator:
 
     """
     def __init__(self, x, y):
+        x = np.atleast_1d(x)
+        if isinstance(x, (float,int)) or isinstance(y, (float,int)):
+            raise ValueError('x and y must not be scalars')
+        if isinstance(x[0], (tuple,list,np.ndarray)) or isinstance(y[0], (tuple,list,np.ndarray)):
+            raise ValueError('x and y must be 1-dimensional')
```
mphoward, line 56: Since you have already cast x to an array, just use the array shape to simplify these tests. (Usually enumerating types is a bad idea.) First, you should coerce both x and y into arrays. Then, you should test the dimension and shape of x, then check that x and y have the same shape. Like this:
```py
x = np.atleast_1d(x)
y = np.atleast_1d(y)
# 1d
if x.ndim > 1:
    raise
# non-scalar
if x.shape[0] == 1:
    raise
# x and y are same shape
if x.shape != y.shape:
    raise
```

relentless/core.py
```diff
@@ -40,6 +49,13 @@ class Interpolator:
...
+        if not np.all(np.diff(x) > 0):
```
mphoward, line 57: You should use something like the other version that I suggested:
```py
if not np.all(x[1:] > x[:-1]):
    raise
```
Your version needs to perform an arithmetic difference between elements, then test that the result is positive. My version performs only a logical comparison between elements. Hence, it's probably about 2x faster if you had a large array.

You have already guaranteed that x has at least 2 elements, so this slice will not fail.

tests/test_core.py
```diff
@@ -1,5 +1,7 @@
 """Unit tests for core module."""
 import unittest
+import numpy as np
+from scipy.interpolate import Akima1DInterpolator
```
mphoward, line 4: You don't need this import any more.

tests/test_core.py
```diff
@@ -8,15 +10,63 @@ class test_Interpolator(unittest.TestCase):
...
+
+        #test construction with scalar input
+        with self.assertRaises(ValueError):
+            f = relentless.core.Interpolator(x=1, y=2)
+
+        #test construction with 2d-array input
+        with self.assertRaises(ValueError):
+            f = relentless.core.Interpolator(x=np.array([[-1,0,1], [-2,2,4]]),
+                                             y=np.array([[-1,0,1], [-2,2,4]]))
+
```
mphoward, line 39: Can you add a test where x and y are both 1d arrays but do not have the same length?

relentless/core.py
```diff
@@ -40,6 +51,16 @@ class Interpolator:
 
     """
     def __init__(self, x, y):
+        x = np.atleast_1d(x)
+        y = np.atleast_1d(y)
+        if x.ndim > 1:
+            raise ValueError('x cannot be a scalar')
+        if x.shape[0] == 1:
+            raise ValueError('x must be 1-dimensional')
```
mphoward, line 59: These error messages are backwards, right?

adisr04: yes

PR #15, opened 2020-06-24
Documentation and unit tests for PairMatrix

adisr04:
- Update PairMatrix documentation
- Add all unit tests for PairMatrix in test_core

### DIFF
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -200,7 +200,7 @@ class PairMatrix:
 
     @property
     def pairs(self):
-        """tuple All unique pairs in the matrix."""
+        """tuple: All unique pairs in the matrix."""
         return tuple(self._data.keys())
 
 class TypeDict:
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -104,7 +104,12 @@ class PairMatrix:
     Parameters
     ----------
     types : array_like
-        List of types (typically, a type is a `str`).
+        List of types (A type must be a `str`).
+
+    Raises
+    ------
+    ValueError
+        If types does not consist of only strings
 
     Examples
     --------
@@ -143,13 +148,15 @@ class PairMatrix:
 
     """
     def __init__(self, types):
+        if not all(isinstance(t, str) for t in types):
+            raise TypeError('All types must be strings')
         self.types = tuple(types)
 
         # flood data with type pairs
         self._data = {}
         for i in self.types:
             for j in self.types:
-                if str(j) >= str(i):
+                if j >= i:
                     self._data[i,j] = {}
 
     def _check_key(self, key):
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -149,7 +149,7 @@ class PairMatrix:
         self._data = {}
         for i in self.types:
             for j in self.types:
-                if j >= i:
+                if str(j) >= str(i):
                     self._data[i,j] = {}
 
     def _check_key(self, key):
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -25,42 +25,34 @@ class test_PairMatrix(unittest.TestCase):
     def test_init(self):
         """Test construction with different list types."""
 
-        types_real = ('A','B')
-        pairs_real  = (('A','B'), ('B','B'), ('A','A'))
+        types = ('A','B')
+        pairs  = (('A','B'), ('B','B'), ('A','A'))
 
         #test construction with tuple input
         m = relentless.core.PairMatrix(types=('A','B'))
-        self.assertEqual(m.types, types_real)
-        self.assertCountEqual(m.pairs, pairs_real)
+        self.assertEqual(m.types, types)
+        self.assertCountEqual(m.pairs, pairs)
 
         #test construction with list input
         m = relentless.core.PairMatrix(types=['A','B'])
-        self.assertEqual(m.types, types_real)
-        self.assertCountEqual(m.pairs, pairs_real)
+        self.assertEqual(m.types, types)
+        self.assertCountEqual(m.pairs, pairs)
 
-        types_real = ('A',)
-        pairs_real = (('A','A'),)
+        types = ('A',)
+        pairs = (('A','A'),)
 
         #test construction with single type tuple
         m = relentless.core.PairMatrix(types=('A',))
-        self.assertEqual(m.types, types_real)
-        self.assertCountEqual(m.pairs, pairs_real)
-
-        types_real = (1,2)
-        pairs_real = ((1,2), (2,2), (1,1))
+        self.assertEqual(m.types, types)
+        self.assertCountEqual(m.pairs, pairs)
 
         #test construction with int type input
-        m = relentless.core.PairMatrix(types=(1,2))
-        self.assertEqual(m.types, types_real)
-        self.assertCountEqual(m.pairs, pairs_real)
-
-        types_real = ('1',2)
-        pairs_real = (('1',2), (2,2), ('1','1'))
+        with self.assertRaises(TypeError):
+            m = relentless.core.PairMatrix(types=(1,2))
 
         #test construction with mixed type input
-        m = relentless.core.PairMatrix(types=('1',2))
-        self.assertEqual(m.types, types_real)
-        self.assertCountEqual(m.pairs, pairs_real)
+        with self.assertRaises(TypeError):
+            m = relentless.core.PairMatrix(types=('1',2))
 
     def test_accessors(self):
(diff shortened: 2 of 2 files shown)


tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
+
+        #test construction with list input
+        m = relentless.core.PairMatrix(types=['A','B'])
+        self.assertEqual(m.types, pair_types)
+        self.assertEqual(m._data, pair_dict)
+
+        #test construction with numpy array input
+        m = relentless.core.PairMatrix(types=np.array(['A','B']))
+        self.assertEqual(m.types, pair_types)
+        self.assertEqual(m._data, pair_dict)
```
mphoward, line 44: NumPy arrays with strings are a little squiggy, so let's not test / support this.

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
+        #test construction with tuple input
+        m = relentless.core.PairMatrix(types=('A','B'))
+        self.assertEqual(m.types, pair_types)
+        self.assertEqual(m._data, pair_dict)
+
+        #test construction with list input
+        m = relentless.core.PairMatrix(types=['A','B'])
+        self.assertEqual(m.types, pair_types)
+        self.assertEqual(m._data, pair_dict)
+
```
mphoward, line 40: Can we do at least one test where the type is not a string? Perhaps an int? This might be a little silly for the user to do, but the class should still work. You can test all ints, and then also mixes of ints and strings. You might want to add it as a separate test method.

Another test that would be good is to use a string int vs. a real int, which are technically different keys, i.e., `"1"` vs. `1`.

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
     def test_init(self):
         """Test construction with different list types."""
-        raise NotImplementedError()
+
+        pair_types = ('A','B')
+        pair_dict = {('A','B'):{}, ('B','B'):{}, ('A','A'):{}}
+
+        #test construction with tuple input
+        m = relentless.core.PairMatrix(types=('A','B'))
+        self.assertEqual(m.types, pair_types)
```
mphoward, line 33: Can you also test `m.pairs`? It should include all the keys in `pair_dict`.

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
+        self.assertEqual(m._data, pair_dict)
```
mphoward, line 34: I wouldn't test `m._data` directly in case the internal storage type changes later. Just check `m.pairs` for now, and then the `test_accessors` will cover the internals.

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
+        self.assertEqual(m['A','B']['mass'], 3.0)
+        self.assertEqual(m['A','B']['energy'], -1.5)
+        self.assertEqual(m['A','B'], {'energy':-1.5, 'mass':3.0})
+
+        m['B','B']['mass'] = 5.0
+        self.assertEqual(m['B','B']['mass'], 5.0)
+        self.assertEqual(m['B','B']['energy'], 0.0)
+        self.assertEqual(m['B','B'], {'energy':0.0, 'mass':5.0})
+
+        m_real = {('A','B'):{'energy':-1.5, 'mass':3.0}, ('A','A'):{'energy':2.0, 'mass':1.0}, ('B','B'):{'energy':0.0, 'mass':5.0}}
```
mphoward, line 107: Your tests above are sufficient, you can remove this check---again, because it is looking at an internal "private" variable, beginning with `_`, so it should be allowed to be silently changeable.

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
         """Test iteration on the matrix."""
-        raise NotImplementedError()
+        m = relentless.core.PairMatrix(types=('A','B'))
+
+        #test iteration for initialization
+        for pair in m:
+            m[pair]['mass'] = 2.0
+            m[pair]['energy'] = 1.0
+        m_real = {('A','B'):{'energy':1.0, 'mass':2.0}, ('A','A'):{'energy':1.0, 'mass':2.0}, ('B','B'):{'energy':1.0, 'mass':2.0}}
+        self.assertEqual(m._data, m_real)
```
mphoward, line 119: It will be a pain, but please do these tests using each entry like `m['A','B']` directly.

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
+
+        m = relentless.core.PairMatrix(types=('A','B'))
+
+        #test set and get for each pair type
+        m['A','A']['energy'] = 1.0
+        self.assertEqual(m['A','A']['energy'], 1.0)
+        self.assertEqual(m['A','B'], {})
+        self.assertEqual(m['B','B'], {})
+
+        m['A','B']['energy'] = -1.0
```
mphoward, line 65: Can you do a test with key `m['B','A']`, which should give the exact same value as `m['A','B']`?

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
+        m['A','B']['mass'] = 3.0
+        self.assertEqual(m['A','B']['mass'], 3.0)
+        self.assertEqual(m['A','B']['energy'], -1.5)
+        self.assertEqual(m['A','B'], {'energy':-1.5, 'mass':3.0})
+
+        m['B','B']['mass'] = 5.0
+        self.assertEqual(m['B','B']['mass'], 5.0)
+        self.assertEqual(m['B','B']['energy'], 0.0)
+        self.assertEqual(m['B','B'], {'energy':0.0, 'mass':5.0})
+
```
mphoward, line 106: Can you also write some tests where you try to read / set keys that are not in the valid types? For example, (C,C) or (A,C). A `KeyError` should be raised in those cases.

tests/test_core.py
```diff
@@ -23,15 +24,112 @@ class test_PairMatrix(unittest.TestCase):
...
+        #test resetting values manually
+        m['A','B']['mass'] = 2.5
+        m['A','A']['energy'] = 1.5
+        m_real = {('A','B'):{'energy':1.0, 'mass':2.5}, ('A','A'):{'energy':1.5, 'mass':2.0}, ('B','B'):{'energy':1.0, 'mass':2.0}}
+        self.assertEqual(m._data, m_real)
+
+        #test re-iteration for setting values
+        for pair in m:
+            m[pair]['mass'] = 3.0
```
mphoward, line 129: Maybe only reset one of 'mass' or 'energy' here? This would be slightly different usage than the test before.

relentless/core.py
```diff
@@ -149,7 +149,7 @@ def __init__(self, types):
         self._data = {}
         for i in self.types:
             for j in self.types:
-                if j >= i:
+                if str(j) >= str(i):
```
mphoward, line 152: This will force all types into strings, which we don't necessarily want to do. Let's revert it back, and just check that an error is raised if two types are mixed that cannot be compared.

tests/test_core.py
```diff
@@ -25,31 +25,42 @@ class test_PairMatrix(unittest.TestCase):
     def test_init(self):
         """Test construction with different list types."""
 
-        pair_types = ('A','B')
-        pair_dict = {('A','B'):{}, ('B','B'):{}, ('A','A'):{}}
+        types_real = ('A','B')
+        pairs_real  = (('A','B'), ('B','B'), ('A','A'))
```
mphoward, line 29: For clarity, let's just call these "types" and "pairs", since "real" doesn't mean much.

tests/test_core.py
```diff
@@ -25,31 +25,42 @@ class test_PairMatrix(unittest.TestCase):
...
+        self.assertEqual(m.types, types_real)
+        self.assertCountEqual(m.pairs, pairs_real)
+
+        types_real = ('1',2)
+        pairs_real = (('1',2), (2,2), ('1','1'))
+
+        #test construction with mixed type input
+        m = relentless.core.PairMatrix(types=('1',2))
+        self.assertEqual(m.types, types_real)
+        self.assertCountEqual(m.pairs, pairs_real)
```
mphoward, line 63: On thinking further, let's make this an error to mix two types that don't have comparators implemented. Just check that an error is raised.

PR #16, opened 2020-06-25
Documentation and unit tests for TypeDict

adisr04:
- Update TypeDict documentation
- Add all unit tests for TypeDict in test_core

### DIFF
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -302,7 +302,7 @@ class TypeDict:
         key = self._check_key(key)
         return self._data[key]
 
-    def setitem__(self, key, value):
+    def __setitem__(self, key, value):
         key = self._check_key(key)
         self._data[key] = value
 
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -215,7 +215,7 @@ class TypeDict:
 
     Raises
     ------
-    ValueError
+    TypeError
         If types does not consist only of strings
 
     Examples
@@ -276,7 +276,7 @@ class TypeDict:
         key = self._check_key(key)
         return self._data[key]
 
-    def __setitem__(self, key, value):
+    def setitem__(self, key, value):
         key = self._check_key(key)
         self._data[key] = value
 
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -295,7 +295,7 @@ class TypeDict:
 
     @property
     def types(self):
-        """tuple All types in the dictionary."""
+        """tuple: All types in the dictionary."""
         return self._types
 
 class Variable:
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -213,6 +213,11 @@ class TypeDict:
     default
         Initial value to fill in the dictionary, defaults to `None`.
 
+    Raises
+    ------
+    ValueError
+        If types does not consist only of strings
+
     Examples
     --------
     Create a type dictionary::
@@ -242,6 +247,8 @@ class TypeDict:
 
     """
     def __init__(self, types, default=None):
+        if not all(isinstance(t, str) for t in types):
+            raise TypeError('All types must be strings')
         self._types = tuple(types)
         self._data = {}
         for i in self.types:
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -104,7 +104,12 @@ class PairMatrix:
     Parameters
     ----------
     types : array_like
-        List of types (typically, a type is a `str`).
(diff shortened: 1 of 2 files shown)


relentless/core.py
```diff
@@ -213,6 +213,11 @@ class TypeDict:
     default
         Initial value to fill in the dictionary, defaults to `None`.
 
+    Raises
+    ------
+    ValueError
```
mphoward, line 218: ValueError -> TypeError

tests/test_core.py
```diff
@@ -38,19 +39,106 @@ class test_TypeDict(unittest.TestCase):
...
+        d = relentless.core.TypeDict(types=('A','B'))
+        self.assertEqual(d.types, types)
+
+        #test construction with list input
+        d = relentless.core.TypeDict(types=['A','B'])
+        self.assertEqual(d.types, types)
+
+        #test construction with numpy array input
+        d = relentless.core.TypeDict(types=np.array(['A','B']))
+        self.assertEqual(d.types, types)
```
mphoward, line 55: Drop numpy array test.

tests/test_core.py
```diff
@@ -38,19 +39,106 @@ class test_TypeDict(unittest.TestCase):
 
     def test_init(self):
         """Test construction with different list types."""
-        raise NotImplementedError()
+
```
mphoward, line 42: Don't leave a whitespace between docstrings and code. (You can delete here and below.)

tests/test_core.py
```diff
@@ -38,19 +39,106 @@ class test_TypeDict(unittest.TestCase):
...
+        types = ('A','B')
+
+        #test construction with tuple input
+        d = relentless.core.TypeDict(types=('A','B'))
+        self.assertEqual(d.types, types)
+
+        #test construction with list input
+        d = relentless.core.TypeDict(types=['A','B'])
+        self.assertEqual(d.types, types)
+
```
mphoward, line 52: Can we add a test where the `default` parameter is set to something other than None?

relentless/core.py
```diff
@@ -295,7 +302,7 @@ def __getitem__(self, key):
         key = self._check_key(key)
         return self._data[key]
 
-    def __setitem__(self, key, value):
+    def setitem__(self, key, value):
```
mphoward, line 305: This looks like a typo, I doubt the tests work correctly without fixing this.

tests/test_core.py
```diff
@@ -199,26 +196,110 @@ class test_TypeDict(unittest.TestCase):
 
     def test_init(self):
         """Test construction with different list types."""
-        raise NotImplementedError()
+        types = ('A','B')
+        default = {'A':1.0, 'B':1.0}
+
+        #test construction with tuple input
+        d = relentless.core.TypeDict(types=('A','B'))
+        self.assertEqual(d.types, types)
```
mphoward, line 204: Can you put back the checks here that the dict is filled with `None` if `default` is not specified? Also, let's not access `_data` directly. Use either `todict` or access by iterator. You could, for example, do this in a one-liner with list comprehension:
```py
self.assertEqual([d[t] for t in d.types], (None, None))
```

tests/test_core.py
```diff
@@ -199,26 +196,110 @@ class test_TypeDict(unittest.TestCase):
...
+        #test construction with list input
+        d = relentless.core.TypeDict(types=['A','B'])
+        self.assertEqual(d.types, types)
+
+        #test construction with defined default input
+        d = relentless.core.TypeDict(types=('A','B'), default=1.0)
+        self.assertEqual(d.types, types)
+        self.assertEqual(d._data, default)
+
+        types = ('A',)
```
mphoward, line 215: This variable looks out of place here, can you place it with the test below your comment?

tests/test_core.py
```diff
@@ -199,26 +196,110 @@ class test_TypeDict(unittest.TestCase):
...
+        self.assertEqual(d.types, types)
+
+        #test construction with list input
+        d = relentless.core.TypeDict(types=['A','B'])
+        self.assertEqual(d.types, types)
+
+        #test construction with defined default input
+        d = relentless.core.TypeDict(types=('A','B'), default=1.0)
+        self.assertEqual(d.types, types)
+        self.assertEqual(d._data, default)
```
mphoward, line 213: Similarly to above, don't access `_data` directly.

PR #19, opened 2020-06-29
Documentation, features, and unit tests for CoefficientMatrix

adisr04:
- Add CoefficientMatrix documentation
- Complete load, save, from_file, setitem functions in CoefficientMatrix
- Add all unit tests for CoefficientMatrix in test_potential
- Add properties and setters for low and high in Variable
- Rename TypeDict class to FixedKeyDict
- Add update method to FixedKeyDict

### DIFF
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -1,4 +1,4 @@
-__all__ = ['Interpolator','Variable']
+__all__ = ['Interpolator','Variable','FixedKeyDict']
 
 from enum import Enum
 
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -353,10 +353,10 @@ class FixedKeyDict:
         if len(data) > 1:
             raise TypeError('More than one positional argument is given')
         d  = data[0] if len(data) == 1 else values
-        for key,value in d.items():
+        for key in d:
             if str(key) not in self.keys:
                 raise KeyError('Only the known keys can be set.')
-            self[key] = value
+            self[key] = d[key]
 
     def todict(self):
         """Convert the fixed-key dictionary to a standard dictionary.
@@ -492,6 +492,8 @@ class Variable:
         if low is not None and not isinstance(low, (float,int)):
             raise ValueError('Low bound must be a float or int')
         self._low = low
+        if hasattr(self, '_value'):
+            self._value, self._state = self.clamp(float(self._value))
 
     @property
     def high(self):
@@ -503,6 +505,8 @@ class Variable:
         if high is not None and not isinstance(high, (float,int)):
             raise ValueError('High bound must be a float or int')
         self._high = high
+        if hasattr(self, '_value'):
+            self._value, self._state = self.clamp(float(self._value))
 
     @property
     def state(self):
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -404,7 +404,6 @@ class Variable:
         self.const = const
         self.low = low
         self.high = high
-
         self.value = value
 
     def clamp(self, value):
@@ -445,6 +444,28 @@ class Variable:
         if not isinstance(value,(float,int)):
             raise ValueError('Variable must be a float or int')
         self._value, self._state = self.clamp(float(value))
+
+    @property
+    def low(self):
+        """float: The low bound of the variable"""
+        return self._low
+
+    @low.setter
+    def low(self, low):
+        if not isinstance(low, (float,int)):
+            raise ValueError('Low bound must be a float or int')
+        self._low = low
+
+    @property
+    def high(self):
+        """float: The high bound of the variable"""
+        return self._high
+
+    @high.setter
+    def high(self, high):
+        if not isinstance(high, (float,int)):
+            raise ValueError('High bound must be a float or int')
+        return self._low
 
     @property
     def state(self):
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -452,7 +452,7 @@ class Variable:
 
     @low.setter
     def low(self, low):
-        if not isinstance(low, (float,int)):
+        if low is not None and not isinstance(low, (float,int)):
             raise ValueError('Low bound must be a float or int')
         self._low = low
 
@@ -463,9 +463,9 @@ class Variable:
 
     @high.setter
     def high(self, high):
-        if not isinstance(high, (float,int)):
+        if high is not None and not isinstance(high, (float,int)):
             raise ValueError('High bound must be a float or int')
-        return self._low
+        self._high = high
 
     @property
     def state(self):
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -352,11 +352,15 @@ class FixedKeyDict:
         """
         if len(data) > 1:
             raise TypeError('More than one positional argument is given')
-        d  = data[0] if len(data) == 1 else values
-        for key in d:
+        if len(data) == 1:
+            for key in data[0]:
+                if str(key) not in self.keys:
+                    raise KeyError('Only the known keys can be set.')
+                self[key] = data[0][key]
+        for key in values:
             if str(key) not in self.keys:
                 raise KeyError('Only the known keys can be set.')
-            self[key] = d[key]
+            self[key] = values[key]
 
     def todict(self):
         """Convert the fixed-key dictionary to a standard dictionary.
@@ -480,7 +484,7 @@ class Variable:
     def value(self, value):
         if not isinstance(value,(float,int)):
             raise ValueError('Variable must be a float or int')
-        self._value, self._state = self.clamp(float(value))
+        self._value, self._state = self.clamp(value)
 
     @property
     def low(self):
@@ -491,9 +495,20 @@ class Variable:
     def low(self, low):
         if low is not None and not isinstance(low, (float,int)):
             raise ValueError('Low bound must be a float or int')
-        self._low = low
-        if hasattr(self, '_value'):
-            self._value, self._state = self.clamp(float(self._value))
+        try:
+            if low is None or self._high is None:
+                self._low = low
+            elif low <= self._high:
+                self._low = low
+            else:
+                raise ValueError('The low bound must be less than the high bound')
+        except AttributeError:
+            self._low = low
+
+        try:
+            self._value, self._state = self.clamp(self._value)
+        except AttributeError:
+            pass
 
     @property
     def high(self):
@@ -504,9 +519,20 @@ class Variable:
     def high(self, high):
         if high is not None and not isinstance(high, (float,int)):
             raise ValueError('High bound must be a float or int')
-        self._high = high
-        if hasattr(self, '_value'):
-            self._value, self._state = self.clamp(float(self._value))
+        try:
+            if high is None or self._high is None:
+                self._high = high
+            elif high >= self._low:
+                self._high = high
+            else:
+                raise ValueError('The high bound must be greater than the low bound')
+        except AttributeError:
+            self._high = high
+
+        try:
+            self._value, self._state = self.clamp(self._value)
+        except AttributeError:
+            pass
 
     @property
     def state(self):
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -335,6 +335,9 @@ class FixedKeyDict:
     def update(self, *data, **values):
         """Partially reassigns key values.
 
+        If both positional argument (data) and keyword arguments (values)
+        are given as parameters, any keys in values will take precedence over data.
+
         Parameters
         ----------
         data : `dict`
@@ -345,21 +348,15 @@ class FixedKeyDict:
         Raises
         ------
         TypeError
-            If more than one positional argument is given
-        KeyError
-            If the key to be updated is not in self.
+            If more than one positional argument is given.
 
         """
         if len(data) > 1:
             raise TypeError('More than one positional argument is given')
-        if len(data) == 1:
+        elif len(data) == 1:
             for key in data[0]:
-                if str(key) not in self.keys:
-                    raise KeyError('Only the known keys can be set.')
                 self[key] = data[0][key]
         for key in values:
-            if str(key) not in self.keys:
-                raise KeyError('Only the known keys can be set.')
             self[key] = values[key]
 
     def todict(self):
@@ -498,7 +495,7 @@ class Variable:
         try:
             if low is None or self._high is None:
                 self._low = low
-            elif low <= self._high:
+            elif low < self._high:
                 self._low = low
             else:
                 raise ValueError('The low bound must be less than the high bound')
@@ -522,7 +519,7 @@ class Variable:
         try:
             if high is None or self._high is None:
                 self._high = high
-            elif high >= self._low:
+            elif high > self._low:
                 self._high = high
             else:
                 raise ValueError('The high bound must be greater than the low bound')
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -1,4 +1,4 @@
-__all__ = ['Interpolator','Variable','FixedKeyDict']
+__all__ = ['Interpolator','Variable']
 
 from enum import Enum
 
@@ -235,9 +235,9 @@ class FixedKeyDict:
     Parameters
     ----------
     keys : array_like
-        List of keys to be fixed (this must be a `str`).
-    default
-        Initial value to fill in the dictionary, defaults to `None`.
+        List of keys to be fixed (a key must be a `str`).
+    default : `dict`
+        Initial values to fill in the dictionary, defaults to `None`.
 
     Raises
     ------
@@ -254,6 +254,12 @@ class FixedKeyDict:
 
         >>> print(d)
         {'A': None, 'B': None}
+
+    Set default values::
+
+        d = FixedKeyDict(keys=('A','B'), default={'A':1.0, 'B':2.0})
+        >>> print(d)
+        {'A':1.0, 'B':2.0}
 
     Iterate as a dictionary::
 
@@ -272,13 +278,13 @@ class FixedKeyDict:
         FixedKeyDict(keys=('A',))
 
     """
-    def __init__(self, keys, default=None):
+    def __init__(self, keys, default={}):
         if not all(isinstance(k, str) for k in keys):
             raise TypeError('All keys must be strings')
         self._keys = tuple(keys)
         self._data = {}
         for i in self.keys:
-            self._data[i] = default
+            self._data[i] = default[i] if i in default else None
 
     def _check_key(self, key):
         """Check that a type is in the dictionary.
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -127,6 +127,8 @@ class PairMatrix:
 
     Raises
     ------
+    ValueError
+        If initialization occurs with empty types
     TypeError
         If types does not consist of only strings
 
@@ -167,6 +169,8 @@ class PairMatrix:
 
     """
     def __init__(self, types):
+        if len(types) == 0:
+            raise ValueError('Cannot initialize with empty types')
         if not all(isinstance(t, str) for t in types):
             raise TypeError('All types must be strings')
         self.types = tuple(types)
@@ -236,7 +240,7 @@ class FixedKeyDict:
     ----------
     keys : array_like
         List of keys to be fixed (a key must be a `str`).
-    default : float or int
+    default
         Initial values to fill in the dictionary, defaults to `None`.
 
     Raises
@@ -272,6 +276,13 @@ class FixedKeyDict:
         1.0
         >>> d['B']
         1.0
+
+    Partially reassign/update values::
+
+        d.update({'A':0.5})
+        d.update(A=0.5)  #equivalent statement
+        >>> print(d)
+        {'A':0.5, 'B':1.0}
 
     Single-key dictionary still needs `keys` as a tuple::
 
@@ -320,6 +331,32 @@ class FixedKeyDict:
 
     def __str__(self):
         return str(self._data)
+
+    def update(self, *data, **values):
+        """Partially reassigns key values.
+
+        Parameters
+        ----------
+        data : `dict`
+            The keys and values to be updated/over-written, in a dictionary form.
+        values : kwargs
+            The keys and values to be updated/over-written.
+
+        Raises
+        ------
+        TypeError
+            If more than one positional argument is given
+        KeyError
+            If the key to be updated is not in self.
+
+        """
+        if len(data) > 1:
+            raise TypeError('More than one positional argument is given')
+        d  = data[0] if len(data) == 1 else values
+        for key,value in d.items():
+            if str(key) not in self.keys:
+                raise KeyError('Only the known keys can be set.')
+            self[key] = value
 
     def todict(self):
         """Convert the fixed-key dictionary to a standard dictionary.
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -236,7 +236,7 @@ class FixedKeyDict:
     ----------
     keys : array_like
         List of keys to be fixed (a key must be a `str`).
-    default : `dict`
+    default : float or int
         Initial values to fill in the dictionary, defaults to `None`.
 
     Raises
@@ -257,9 +257,9 @@ class FixedKeyDict:
 
     Set default values::
 
-        d = FixedKeyDict(keys=('A','B'), default={'A':1.0, 'B':2.0})
+        d = FixedKeyDict(keys=('A','B'), default=0.0)
         >>> print(d)
-        {'A':1.0, 'B':2.0}
+        {'A':0.0, 'B':0.0}
 
     Iterate as a dictionary::
 
@@ -278,13 +278,13 @@ class FixedKeyDict:
         FixedKeyDict(keys=('A',))
 
     """
-    def __init__(self, keys, default={}):
+    def __init__(self, keys, default=None):
         if not all(isinstance(k, str) for k in keys):
             raise TypeError('All keys must be strings')
         self._keys = tuple(keys)
         self._data = {}
         for i in self.keys:
-            self._data[i] = default[i] if i in default else None
+            self._data[i] = default
 
     def _check_key(self, key):
         """Check that a type is in the dictionary.
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -229,26 +229,26 @@ class PairMatrix:
         """tuple: All unique pairs in the matrix."""
         return tuple(self._data.keys())
 
-class TypeDict:
-    """Dictionary with fixed keys set by type.
+class FixedKeyDict:
+    """Dictionary with fixed keys.
 
     Parameters
     ----------
-    types : array_like
-        List of types (typically, a type is a `str`).
+    keys : array_like
+        List of keys to be fixed (this must be a `str`).
     default
         Initial value to fill in the dictionary, defaults to `None`.
 
     Raises
     ------
     TypeError
-        If types does not consist only of strings
+        If keys does not consist only of strings
 
     Examples
     --------
-    Create a type dictionary::
+    Create a keyed dictionary::
 
-        d = TypeDict(types=('A','B'))
+        d = FixedKeyDict(keys=('A','B'))
 
     Default values::
 
@@ -257,8 +257,8 @@ class TypeDict:
 
     Iterate as a dictionary::
 
-        for t in d:
-            d[t] = 1.0
+        for k in d:
+            d[k] = 1.0
 
     Access by key::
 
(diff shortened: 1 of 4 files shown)


relentless/potential/potential.py
```diff
@@ -1,16 +1,63 @@
...
+    Create a coefficient matrix with default values::
+
+        m = CoefficientMatrix(types=('A','B'), params=('energy','mass'),
+                              default={'energy':1.0, 'mass':2.0})
+
+    Set coefficient matrix values::
+
+        m['A','A']['energy'] = 2.0
+        m['A,'B']['mass'] = 1.5
+        m['A','B']['energy'] = 3.0
```
mphoward, line 45: We should discuss how users are allowed to interact with parameters in the `CoefficientMatrix` tomorrow. The original design I had in mind was to require parameters always be given as dictionaries:
```py
m['A','B'] = {'mass': 1.5, 'energy': 3.0}
```
These dictionaries currently work in an "append" mode, which may not be ideal because it is not intuitive that parameters that are unset stay at their current value. (i.e., if you set a parameter to something, you must explicitly reset it.) A user might expect that values not set in the dict get reset to their default.

It would also be very natural for someone to try to set a key individually, as Adithya has done:
```py
m['A','B']['mass'] = 1.5
m['A','B']['energy'] = 3.0
```
We should decide if this is something we should allow. I believe that currently there is no way to prevent the user from injecting a parameter into the dict because they get access to the underlying data, and then they can do whatever they want to it.

relentless/core.py
```diff
@@ -1,4 +1,4 @@
-__all__ = ['Interpolator','Variable']
+__all__ = ['Interpolator','Variable','FixedKeyDict']
```
mphoward, line 1: `relentless/__init__.py` will import everything from `core.__all__` into its base namespace, so we only want to expose classes the user should actually be working with. We should omit containers like `FixedKeyDict` or `PairMatrix`.

relentless/core.py
```diff
@@ -229,26 +229,26 @@ def pairs(self):
...
-    """Dictionary with fixed keys set by type.
+class FixedKeyDict:
+    """Dictionary with fixed keys.
 
     Parameters
     ----------
-    types : array_like
-        List of types (typically, a type is a `str`).
+    keys : array_like
+        List of keys to be fixed (this must be a `str`).
```
mphoward, line 238: "this must be" -> "a key must be"

relentless/potential/potential.py
```diff
@@ -40,29 +41,33 @@ class CoefficientMatrix(PairMatrix):
 
     Set coefficient matrix values::
 
-        m['A','A']['energy'] = 2.0
-        m['A,'B']['mass'] = 1.5
-        m['A','B']['energy'] = 3.0
+        m['A','A']['energy'] = 2.0               #access param directly
+        m['A','B'] = {'energy':0.0, 'mass':1.5}  #reset default in full
+        m['A','A'] = {'mass':0.5}                #reset default partially
+        for p in self.params:                    #iteratively acesss params
```
mphoward, line 47: Each of these is a different example (with the description given in the comment), so please separate them out.

relentless/potential/potential.py
```diff
@@ -40,29 +41,33 @@ class CoefficientMatrix(PairMatrix):
...
+            m['B','B'][p] = 0.1
```
mphoward, line 48: Can we add an example using a `Variable`?

relentless/potential/potential.py
```diff
@@ -40,29 +41,33 @@ class CoefficientMatrix(PairMatrix):
...
 
     Evaluate (retrieve) pair parameters::
```
mphoward, line 50: Similarly, can you also show an example of what happens when you do `m.evaluate(('A','A'))` vs. accessing `m['A','A']`?

relentless/potential/potential.py
```diff
@@ -40,29 +41,33 @@ class CoefficientMatrix(PairMatrix):
...
         self.params = tuple(params)
+
+        d = FixedKeyDict(keys=self.params)
+        for i in d:
+            d[i] = default[i] if i in default else None
         for key in self:
-            for p in self.params:
-                v = default[p] if p in default else None
-                self[key][p] = v
+            self[key] = d
```
mphoward, line 70: I think you need to create `FixedKeyDict` here inside the loop, rather than once above. As it reads now, I am worried that python might assign the _reference_ to the same object to `self[key]`, which means a change to one will change them all. Your unit tests look like they check for this, so maybe I am mistaken Alternatively, we could implement a `copy` method for `FixedKeyDict`.

tests/potential/test_potential.py
```diff
@@ -60,38 +60,33 @@ def test_accessors(self):
         self.assertEqual(m['B','B']['energy'], None)
         self.assertEqual(m['B','B']['mass'], None)
 
-        m['A','B']['energy'] = 1.0
```
mphoward: There is a lot going on in these tests, so let's split them into two methods (two `test_` methods). First, do a test with just one type 'A', and check all these different scenarios. Then, you can do a test with multiple types for a limited scenario or two.

tests/potential/test_potential.py
```diff
@@ -60,38 +60,33 @@ def test_accessors(self):
...
         self.assertEqual(m['A','A']['energy'], None)
         self.assertEqual(m['A','A']['mass'], None)
         self.assertEqual(m['B','B']['energy'], None)
         self.assertEqual(m['B','B']['mass'], None)
 
-        m['A','A']['energy'] = 1.0
+        #test setting key values partially
+        m['A','A'] = {'energy':1.0}
         self.assertEqual(m['A','B']['energy'], 1.0)
         self.assertEqual(m['A','B']['mass'], 2.0)
```
mphoward: This is not really the behavior we want to have here, I think. When setting using the `=` (assignment) operator, we want to just set the values to exactly what's in the dict. If something is not in the dict, it should default to the `default` value specified in the constructor. If this is also not specified, it should be `None`. You will need to modify some of the methods above to get this to work.

Then, we will support this case of partial update with a new method later.

adisr04: I don't really understand what you're saying here, in your second sentence. Which `dict` should the values be set to? Should we just disallow partial reassignment?

mphoward: Yes, we don't want to allow "partial" reassignment by the `=` operator, just complete reassignment. I'll try to give an example.

Suppose we start with defaults
```py
m = CoefficientMatrix(types=('A',), params=('energy','mass'), default={'energy': 0.0, 'mass': 1.0})
```
If I assign on m, then I should get:
```py
>>> m['A','A'] = {'energy': 1.0, 'mass': 2.0}
>>> m['A','A']['energy'] == 1.0
True
>>> m['A','A']['mass'] == 2.0
True
```
If I started with the initial m and only assigned partially, I should leave the mass at its default:
```py
>>> m['A','A'] = {'energy': 1.0}
>>> m['A','A']['energy'] == 1.0
True
>>> m['A','A']['mass'] == 1.0
True
```
Now, if I combined the two (execute these blocks one after the other), I would expect the same behavior because the `=` assignment should completely overwrite the entry. I believe as implemented the mass stays at 2.0, and the last check would fail, right?

We will want to implement a separate `update` method (or something like that), which allows for partial assignment of parameters over existing ones.

relentless/potential/potential.py
```diff
@@ -34,28 +123,52 @@ def evaluate(self, pair):
...
+        filename : `str`
+            The name of the file from which to load data.
+
+        """
+        with open(filename, 'r') as f:
+            data = json.load(f)
+        for key in data:
+            key_tup = tuple(str(key).strip('"'))
+            if key_tup in self:
+                self[key_tup] = data[key]
```
mphoward, line 160: This is heading in the right direction! A few things that will make this more robust:
1. Can you check that the keys of data are _exactly_ the `pairs` of this matrix? That is, they include no more and no less. Otherwise, the JSON schema is invalid.
2. Can you check that the params in each `data[pair]` are exactly those that are in this object? You do check this indirectly with `__setitem__`, but the error message might be confusing to a user calling `load`. Instead, you should raise an exception here that alerts the user that the parameters loaded are inconsistent with the schema of this object.
3. When the parameters are dumped by save, they are evaluated first. This means that we lose information about whether they were stored in a float or a `Variable` inside this class. When you load, you need to detect whether you are working with a scalar or a `Variable`, and then save appropriately. To address this, we probably need to modify the `save` method, so let's discuss tomorrow.
4. Does that combination of tuple str and strip actually explode the key into two strings? I didn't know that! If not, you can go back to using that other tool to process the key.

relentless/potential/potential.py
```diff
@@ -1,29 +1,118 @@
...
+
+    Create a coefficient matrix with a default value::
+
+        m = CoefficientMatrix(types=('A','B'), params=('energy','mass'), default=0.0)
+
+    Set coefficient matrix values by accessing parameter directly::
+
+        m['A','A']['energy'] = 2.0
+
+    Assigning to a pair using `update()` overwrites parameters::
```
mphoward, line 44: Let's be a little more specific, it "overwrites only the specified parameters".

relentless/potential/potential.py
```diff
@@ -1,29 +1,118 @@
...
         super().__init__(types)
+        if not all(isinstance(p, str) for p in params):
+            raise TypeError('All parameters must be strings')
         self.params = tuple(params)
+        self.default = default
         for key in self:
-            for p in self.params:
-                v = default[p] if p in default else None
-                self[key][p] = v
+            self[key] = FixedKeyDict(keys=self.params,default=self.default)
```
mphoward, line 87: Just noting here, per our discussion, change to `self[key].update(self.default)`.

relentless/potential/potential.py
```diff
@@ -1,29 +1,118 @@
...
 
     def evaluate(self, pair):
         """Evaluate pair parameters.
 
+        Returns a dictionary of the parameters and values for the specified pair.
+
+        Parameters
+        ----------
+        pair : array_like
```
mphoward, line 96: Can `pair` be array_like, or does it need to be a tuple? I forget how the indexing is defined.

relentless/potential/potential.py
```diff
@@ -34,28 +123,91 @@ def evaluate(self, pair):
...
+        if isinstance(values, dict):
+            for key in values:
+                self[key] = values[key]
+        else:
+            for key,value in values.items():
+                key = str(key)
+                if key not in self.params:
+                    raise KeyError('Only the known parameters can be set in the coefficient matrix.')
+                self[key] = value
+
```
mphoward, line 156: Sorry I didn't notice this earlier. This method needs to be defined for the `FixedKeyDict`, not the `CoefficientMatrix` since we want to be able to do
```
m['A','B'].update(coeffs)
```
which is calling the `update` method of `m['A','B']`. Can you move it over to that class?

Also, I'm not sure if this call signature is quite right. I think you need to accept an optional positional argument and optional keyword args, like this:
```py
def update(self, *other, **values):
    if len(other) > 1:
        # raise an error if more than one positional arg is given

    # use other[0] if given, otherwise try to use the optional kwargs
    d = other[0] if len(other) == 1 else values

    # iterate over d and assign values using the second code path
    for key,value in d.items():
        # set stuff as before
```

relentless/core.py
```diff
@@ -440,6 +482,28 @@ def value(self, value):
...
+    def high(self):
+        """float: The high bound of the variable"""
+        return self._high
+
+    @high.setter
+    def high(self, high):
+        if high is not None and not isinstance(high, (float,int)):
+            raise ValueError('High bound must be a float or int')
+        self._high = high
+
```
mphoward, line 506: Similarly here for high.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+        data = {}
         for key in self:
-            all_params[str(key)] = self.evaluate(key)
+            all_params = {}
+            for param in self[key]:
+                var = self[key][param]
+                if isinstance(var, Variable):
+                    all_params[param] = {'type':'variable', 'value':var.value,
+                                         'const':var.const, 'low':var.low, 'high':var.high}
+                elif isinstance(var, (float,int)):
```
mphoward, line 159: This should be `np.isscalar` for consistency with evaluate method.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+                    all_params[param] = {'type':'scalar', 'value':var}
```
mphoward, line 160: There should be an else branch here that raises an error if the type is unknown, as in `evaluate`.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+        Parameters
+        ----------
+        filename : `str`
+            The name of the file to save data in.
+
+        """
+        data = {}
         for key in self:
-            all_params[str(key)] = self.evaluate(key)
+            all_params = {}
```
mphoward, line 153: To simplify this, you can make this `data[str(key)] = {}`, then set the parameters directly in this dict rather than using `all_params` and then copying.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+        KeyError
+            If the pairs in the data are not exactly the pairs in self
+        KeyError
+            If the params in the data are not exactly the params in self
+        TypeError
+            If the parameter value type in the data and in self do not match
+
+        """
+        with open(filename, 'r') as f:
+            data = json.load(f)
```
mphoward, line 184: You end up coercing the keys twice, which is not ideal. Better would be to do something like
```py
data_ = json.load(f)
data = {eval(key): data_[key] for key in data_}
```
so that `data` is just keyed by tuples.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+        pairs = []
+        for key in data:
+            pairs.append(eval(key))
+        if sorted(self.pairs) != sorted(tuple(pairs)):
+            raise KeyError('The type pairs in the data are not exactly the pairs in self')
```
mphoward, line 189: This should generate the correct result but is kind of confusing. Maybe you want to do a more straightforward check, like:
```py
if not (all(p in self.pairs for p in data) and all(p in data for p in self.pairs)):
    # raise
```
You could also use a `set`, like
```py
if not set(self.pairs) == set(data.keys()):
    # raise
```

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+        for key in data:
+            key_tup = eval(key)
+            if sorted(self.params) != sorted(data[key].keys()):
+                raise KeyError('The parameters in the data are not exactly the parameters in self')
```
mphoward, line 193: Similarly here.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+            for param in data[key]:
+                var_data = data[key][param]
+                var_self = self[key_tup][param]
```
mphoward, line 196: If you make the change I suggested above, then `key` will be the same as `key_tup`, and you should update this section. Also, I think you probably don't need a `var_self` since you also use `self[key_tup][param]` repeatedly below for setting, and I would prefer to just use the second.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+                if var_self is None:
+                    if var_data['type'] == 'variable':
+                        self[key_tup][param] = Variable(value=var_data['value'], const=var_data['const'],
+                                                        low=var_data['low'], high=var_data['high'])
+                    elif var_data['type'] == 'scalar':
+                        self[key_tup][param] = var_data['value']
+                    else:
+                        raise TypeError('The parameter value types in the data and self do not match')
```
mphoward, line 204: I think this error should just be "Unknown variable type", right?

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+                elif var_data['type'] == 'scalar' and isinstance(var_self, (float,int)):
```
mphoward, line 205: `np.isscalar`

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+                    self[key_tup][param] = var_data['value']
+                elif var_data['type'] == 'variable' and isinstance(var_self, Variable):
```
mphoward, line 207: For consistency with above, can you please flip this `elif` with the one above, so that you check for variable first, then scalar?

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+                    self[key_tup][param] = Variable(value=var_data['value'], const=var_data['const'],
```
mphoward, line 208: We don't want to overwrite the reference to the original variable, so instead set everything by property:
```py
self[key][param].value = var_data['value']
```

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+            The name of the file from which to load data.
+
+        Returns
+        -------
+        :py:class:`CoefficientMatrix`
+            Object initialized with the data in the file.
+
+        """
+        with open(filename, 'r') as f:
+            data = json.load(f)
```
mphoward, line 229: Use the same method I suggested above to get the tuples as the keys of `data`.

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
+        t_all = data.keys()
+        p = data[list(t_all)[0]].keys()
+        for par in t_all:
+            par = eval(par)
+            if len(par) == len(set(par)):
+                t = par
+                break
```
mphoward, line 236: This is a little confusing what it's doing. Maybe better would be:
```py
types = set()
for i,j in data:
    types.add(i)
    types.add(j)
types = list(types)
```

relentless/potential/potential.py
```diff
@@ -34,28 +131,123 @@ def evaluate(self, pair):
...
         for p in value:
             if p not in self.params:
-                raise KeyError('Only the known parameters can be set in coefficient matrix.')
-            self[key][p] = value[p]
+                raise KeyError('Only the known parameters can be set in the coefficient matrix.')
+        for p in self.params:
+            if p in value:
+                self[key][p] = value[p]
+            else:
+                self[key][p] = self.default[p] if p in self.default else None
```
mphoward, line 250: I think it might be better to simplify this as two update calls:
```py
self[key].update(self.default)
self[key].update(value)
```
The first resets everything to its default, then the second overwrites whatever was set in the values.

relentless/core.py
```diff
@@ -440,6 +482,32 @@ def value(self, value):
...
+    def low(self):
+        """float: The low bound of the variable"""
+        return self._low
+
+    @low.setter
+    def low(self, low):
+        if low is not None and not isinstance(low, (float,int)):
+            raise ValueError('Low bound must be a float or int')
+        self._low = low
+        if hasattr(self, '_value'):
```
mphoward, line 495: Could you do this with `try except` instead? I think you want:
```py
try:
    self._value, self._state self.clamp(self._value)
except AttributeError:
    pass
```

relentless/core.py
```diff
@@ -440,6 +482,32 @@ def value(self, value):
             raise ValueError('Variable must be a float or int')
         self._value, self._state = self.clamp(float(value))
```
mphoward, line 483: I just noticed: we should not cast `value` to float here, it should just be as given.

relentless/potential/potential.py
```diff
@@ -34,28 +135,135 @@ def evaluate(self, pair):
...
+        Raises
+        ------
+        TypeError
+            If a value is of an unrecognizable type
+
+        """
+        data = {}
         for key in self:
-            all_params[str(key)] = self.evaluate(key)
+            data[str(key)] = {}
```
mphoward, line 162: Since the key will get reused a couple times below, let's save the string version once and recycle it, like
```py
dkey = str(key)
data[dkey] = {}
```

relentless/potential/potential.py
```diff
@@ -34,28 +135,135 @@ def evaluate(self, pair):
...
+                var_data = data[key][param]
+                if self[key][param] is None:
+                    if var_data['type'] == 'variable':
+                        self[key][param] = Variable(value=var_data['value'], const=var_data['const'],
+                                                        low=var_data['low'], high=var_data['high'])
+                    elif var_data['type'] == 'scalar':
+                        self[key][param] = var_data['value']
+                    else:
+                        raise TypeError('Unrecognized value type in the data')
+
```
mphoward, line 216: Remove this blank line for readability.

relentless/potential/potential.py
```diff
@@ -34,28 +135,135 @@ def evaluate(self, pair):
...
+        -------
+        :py:class:`CoefficientMatrix`
+            Object initialized with the data in the file.
+
+        """
+        with open(filename, 'r') as f:
+            data_ = json.load(f)
+        data = {eval(key): data_[key] for key in data_}
+
+        par = None
```
mphoward, line 246: Could you add a short comment here, like `# get the keys and parameters from the JSON file`

tests/potential/test_potential.py
```diff
@@ -0,0 +1,294 @@
...
+        #test reset and get manually
+        m['A','B'] = {'energy':1.0, 'mass':2.0}
+        self.assertEqual(m['A','B']['energy'], 1.0)
+        self.assertEqual(m['A','B']['mass'], 2.0)
+        self.assertEqual(m['A','A']['energy'], 0.0)
+        self.assertEqual(m['A','A']['mass'], 0.0)
+        self.assertEqual(m['B','B']['energy'], 0.0)
+        self.assertEqual(m['B','B']['mass'], 0.0)
+
+        m['A','A'] = {'energy':1.0, 'mass':2.0}
```
mphoward, line 120: To get better coverage, could you make this an `update` using the dict and also use different values?

tests/potential/test_potential.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertEqual(m['A','B']['energy'], 1.0)
+        self.assertEqual(m['A','B']['mass'], 2.0)
+        self.assertEqual(m['A','A']['energy'], 1.0)
+        self.assertEqual(m['A','A']['mass'], 2.0)
+        self.assertEqual(m['B','B']['energy'], 0.0)
+        self.assertEqual(m['B','B']['mass'], 0.0)
+
+        m['B','B'] = {'energy':1.0, 'mass':2.0}
```
mphoward, line 128: And here, could you do this by update using the keyword version, and again different values?

tests/potential/test_potential.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertEqual(m['A','A']['energy'], 1.0)
+        self.assertEqual(m['A','A']['mass'], 2.0)
+
+        #test reset and get via iteration
+        params = ('energy', 'mass')
+        for j in params:
+            m['A','A'][j] = 1.5
+        self.assertEqual(m['A','A']['energy'], 1.5)
+        self.assertEqual(m['A','A']['mass'], 1.5)
+
```
mphoward, line 98: Can you add some tests here where you both get / set on an invalid (1) type pair and (2) parameter name?

tests/potential/test_potential.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertEqual(m.evaluate(('B','A')), m.evaluate(('A','B')))
+
+        #test evaluation with initialized parameter values as unrecognized types
+        m = potential.CoefficientMatrix(types=('A','B'), params=('energy','mass'),
+                                        default={'energy':core.Interpolator(x=(-1,0,1), y=(-2,0,2)),
+                                                 'mass':core.Interpolator(x=(-1,0,1), y=(-2,0,2))})
+        with self.assertRaises(TypeError):
+            x = m.evaluate(('A','B'))
+
+    def test_parse(self):
```
mphoward, line 186: Can we call this `test_saveload` or `test_io`?

tests/potential/test_potential.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertEqual(m['A','A']['mass'].low, x['A','A']['mass'].low)
+        self.assertEqual(m['B','B']['energy'].low, x['B','B']['energy'].low)
+        self.assertEqual(m['B','B']['mass'].low, x['B','B']['mass'].low)
+
+        self.assertEqual(m['A','B']['energy'].high, x['A','B']['energy'].high)
+        self.assertEqual(m['A','B']['mass'].high, x['A','B']['mass'].high)
+        self.assertEqual(m['A','A']['energy'].high, x['A','A']['energy'].high)
+        self.assertEqual(m['A','A']['mass'].high, x['A','A']['mass'].high)
+        self.assertEqual(m['B','B']['energy'].high, x['B','B']['energy'].high)
+        self.assertEqual(m['B','B']['mass'].high, x['B','B']['mass'].high)
```
mphoward, line 289: Can we make a separate test method for `from_file`? This one is starting to get kinda bulky to read.

tests/test_core.py
```diff
@@ -250,17 +250,26 @@ def test_accessors(self):
         self.assertEqual(d['A'], 2.0)
         self.assertEqual(d['B'], 1.5)
 
+        #test using `update()`
```
mphoward, line 253: Can you make a separate test for the `update` method? It would also be good to check: updating for only a subset of the keys (rather than both), and to check that the appropriate behavior happens if both *args and **kwargs are used.

tests/test_core.py
```diff
@@ -341,7 +350,13 @@ def test_init(self):
...
         #test invalid value initialization
         with self.assertRaises(ValueError):
-            v = relentless.core.Variable(value='4')
+            v.value = '4'
+        #test invalid low initialization
+        with self.assertRaises(ValueError):
+            v.low = '4'
+        #test invalid high initialization
+        with self.assertRaises(ValueError):
+            v.high = '4'
```
mphoward, line 359: Can you add a test or two checking that you can set low and high to real values, and that the clamped values are updated appropriately? This should probably a separate test called `test_bounds`.

relentless/potential/potential.py
```diff
@@ -1,29 +1,130 @@
...
+    Raises
+    ------
+    ValueError
+        If params is initialized as empty
+    TypeError
+        If params does not consist of only strings
+
+    Examples
+    --------
+
```
mphoward, line 34: Delete this blank line.

relentless/potential/potential.py
```diff
@@ -34,28 +135,136 @@ def evaluate(self, pair):
...
+        data = {}
         for key in self:
-            all_params[str(key)] = self.evaluate(key)
+            dkey = str(key)
+            data[dkey] = {}
+            for param in self[key]:
+                var = self[key][param]
+                if isinstance(var, Variable):
+                    data[dkey][param] = {'type':'variable', 'value':var.value,
+                                             'const':var.const, 'low':var.low, 'high':var.high}
```
mphoward, line 168: Fix whitespace alignment here.

relentless/potential/potential.py
```diff
@@ -34,28 +135,136 @@ def evaluate(self, pair):
...
+        for key in data:
+            if set(self.params) != set(data[key].keys()):
+                raise KeyError('The parameters in the data are not exactly the parameters in self')
+
+            for param in data[key]:
+                var_data = data[key][param]
+                if self[key][param] is None:
+                    if var_data['type'] == 'variable':
+                        self[key][param] = Variable(value=var_data['value'], const=var_data['const'],
+                                                        low=var_data['low'], high=var_data['high'])
```
mphoward, line 212: Fix indent here.

tests/potential/test_potential.py
```diff
@@ -0,0 +1,304 @@
...
+
+        m['A','A'].update({'energy':1.5, 'mass':2.5})
+        self.assertEqual(m['A','B']['energy'], 1.0)
+        self.assertEqual(m['A','B']['mass'], 2.0)
+        self.assertEqual(m['A','A']['energy'], 1.5)
+        self.assertEqual(m['A','A']['mass'], 2.5)
+        self.assertEqual(m['B','B']['energy'], 0.0)
+        self.assertEqual(m['B','B']['mass'], 0.0)
+
+        m['B','B'] = {'energy':3.0, 'mass':4.0}
```
mphoward, line 138: Can you make this test:
```py
m['B','B'].update(energy=3.0, mass=4.0)?
```
just to test that this also works correctly?

PR #20, opened 2020-07-08
Documentation, features, and unit tests for PairPotential

adisr04:
- Add PairPotential documentation
- Implement `@abstractmethod` decoration for base class energy/force/derivative functions
- Modify load and save functions, and default parameter initialization in PairPotential
- Add all unit tests for PairPotential in test_potential, using Linear Potential example

### DIFF
--- a/relentless/potential/potential.py
+++ b/relentless/potential/potential.py
@@ -1,8 +1,8 @@
 __all__ = ['PairPotential','Tabulator']
 
+import abc
 import json
 import numpy as np
-import abc
 
 from relentless.core import PairMatrix
 from relentless.core import FixedKeyDict
@@ -282,10 +282,6 @@ class PairPotential(abc.ABC):
     default : dict
         Initial value for any or all parameters. Any value not specified defaults to `None`.
 
-    Examples
-    --------
-    ****TODO****
-
     Todo
     ----
     1. Inspect _energy() call signature for parameters.
@@ -315,6 +311,13 @@ class PairPotential(abc.ABC):
     def energy(self, pair, r):
         """Evaluate energy for a (i,j) pair.
 
+        If an `rmin` or `rmax` value is set, then any energy evaluated at an `r`
+        value greater than `rmax` or less than `rmin` is set to the value of energy
+        evaluated or `rmax` or `rmin`, respectively.
+
+        Additionally, if the `shift` parameter is set to be `True`, all energy values
+        are shifted so that the energy is `rmax` is 0.
+
         Parameters
         ----------
         pair : array_like
@@ -324,8 +327,10 @@ class PairPotential(abc.ABC):
 
         Returns
         -------
-        array_like
+        scalar or array_like
             The energy at the specified location(s).
+            The returned quantity will be a scalar if `r` is scalar
+            or a numpy array if `r` is array_like.
 
         Raises
         ------
@@ -335,10 +340,10 @@ class PairPotential(abc.ABC):
             If the potential is to be shifted without setting `rmax`.
 
         """
-        if np.min(r) < 0:
-            raise ValueError('r cannot be negative')
         params = self.coeff.evaluate(pair)
         r,u,scalar_r = self._zeros(r)
+        if any(r < 0):
+            raise ValueError('r cannot be negative')
 
         # evaluate at points below rmax (if set) first, including rmin cutoff (if set)
         flags = np.ones(r.shape[0], dtype=bool)
@@ -369,6 +374,8 @@ class PairPotential(abc.ABC):
     def force(self, pair, r):
         """Evaluate force for a (i,j) pair.
 
+        The force is only evaluated for `r` values between `rmin` and `rmax`, if set.
+
         Parameters
         ----------
         pair : array_like
@@ -378,8 +385,10 @@ class PairPotential(abc.ABC):
 
         Returns
         -------
-        array_like
+        scalar or array_like
             The force at the specified location(s).
+            The returned quantity will be a scalar if `r` is scalar
+            or a numpy array if `r` is array_like.
 
         Raises
         ------
@@ -387,10 +396,10 @@ class PairPotential(abc.ABC):
             If any value in `r` is negative.
 
         """
-        if np.min(r) < 0:
-            raise ValueError('r cannot be negative')
         params = self.coeff.evaluate(pair)
         r,f,scalar_r = self._zeros(r)
+        if any(r < 0):
+            raise ValueError('r cannot be negative')
 
         # only evaluate at points inside [rmin,rmax], if specified
         flags = np.ones(r.shape[0], dtype=bool)
@@ -408,6 +417,8 @@ class PairPotential(abc.ABC):
     def derivative(self, pair, param, r):
         """Evaluate derivative for a (i,j) pair with respect to a parameter.
 
+        The derivative is only evaluated for `r` values between `rmin` and `rmax`, if set.
+
         Parameters
         ----------
         pair : array_like
@@ -419,8 +430,10 @@ class PairPotential(abc.ABC):
 
         Returns
         -------
-        array_like
-            The derivative at the specified locations.
+        scalar or array_like
+            The derivative at the specified location(s).
+            The returned quantity will be a scalar if `r` is scalar
+            or a numpy array if `r` is array_like.
 
         Raises
         ------
@@ -428,10 +441,10 @@ class PairPotential(abc.ABC):
             If any value in `r` is negative.
 
         """
-        if np.min(r) < 0:
-            raise ValueError('r cannot be negative.')
         params = self.coeff.evaluate(pair)
         r,deriv,scalar_r = self._zeros(r)
+        if any(r < 0):
+            raise ValueError('r cannot be negative')
 
         # only evaluate at points inside [rmin,rmax], if specified
         flags = np.ones(r.shape[0], dtype=bool)
@@ -447,7 +460,7 @@ class PairPotential(abc.ABC):
         return deriv
 
     def _zeros(self, r):
-        """Coerces input shape and creates zeros for output.
+        """Force input to a 1-dimensional array and make matching array of zeros.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


tests/potential/test_potential.py
```diff
@@ -308,5 +309,281 @@ def test_fromfile(self):
...
+        p = test_PairPotential.LinPot(types=('1',), params=('m',),
+                                      default={'m':2})
+        u = p.energy(pair=(('1','1')), r=0.5)
+        self.assertEqual(u, 1.0)
+        u = p.energy(pair=(('1','1')), r=[0.25,0.75])
+        np.testing.assert_allclose(u, [0.5,1.5])
+
+        #test with rmin set
+        p = test_PairPotential.LinPot(types=('1',), params=('m','rmin'),
+                                      default={'m':2,'rmin':0.0})
```
mphoward, line 447: Rather than creating new pair potentials each time, could you set the `coeff`? This would give us some extra tests that `evaluate` is being called properly, etc. Creating new objects is probably not necessary since you have good coverage in the `test_init`.

tests/potential/test_potential.py
```diff
@@ -308,5 +309,281 @@ def test_fromfile(self):
...
+        u = p.energy(pair=(('1','1')), r=0.0)
```
mphoward, line 448: Make `rmin` something bigger than 0.0 like 0.1, it isn't a good test because the potential is trivially `u(0) = 0` even if you don't use `rmin` correctly.

tests/potential/test_potential.py
```diff
@@ -308,5 +309,281 @@ def test_fromfile(self):
...
+        self.assertEqual(u, 0.0)
+        u = p.energy(pair=(('1','1')), r=[-0.25,0.75])
```
mphoward, line 450: `r` should not be a negative value. Maybe you should add an error check for this in the `PairPotential`, and then update this test.

tests/potential/test_potential.py
```diff
@@ -308,5 +309,281 @@ def test_fromfile(self):
...
+        with self.assertRaises(TypeError):
+            r_copy, u, s = p._zeros(r)
+
+    def test_energy(self):
+        """Test energy method"""
+        #test with no cutoffs
+        p = test_PairPotential.LinPot(types=('1',), params=('m',),
+                                      default={'m':2})
+        u = p.energy(pair=(('1','1')), r=0.5)
+        self.assertEqual(u, 1.0)
```
mphoward, line 441: All of these checks should probably be `assertAlmostEqual` because you are getting floats back.

tests/potential/test_potential.py
```diff
@@ -308,5 +309,281 @@ def test_fromfile(self):
...
+
+        #test with shift set
+        p = test_PairPotential.LinPot(types=('1',), params=('m','rmax','shift'),
+                                      default={'m':2,'rmax':1.0,'shift':True})
+        u = p.energy(pair=(('1','1')), r=0.5)
+        self.assertEqual(u, -1.0)
+        u = p.energy(pair=(('1','1')), r=[1.0,1.5])
+        np.testing.assert_allclose(u, [0.0,0.0])
+
+    def test_force(self):
```
mphoward, line 477: Make similar changes here as in `test_energy`.

tests/potential/test_potential.py
```diff
@@ -308,5 +309,281 @@ def test_fromfile(self):
...
+
+        #test with shift set
+        p = test_PairPotential.LinPot(types=('1',), params=('m','rmax','shift'),
+                                      default={'m':2,'rmax':1.0,'shift':True})
+        f = p.force(pair=(('1','1')), r=0.5)
+        self.assertEqual(f, -2.0)
+        f = p.force(pair=(('1','1')), r=[1.0,1.5])
+        np.testing.assert_allclose(f, [-2.0,0.0])
+
+    def test_derivative(self):
```
mphoward, line 519: Same here.

tests/potential/test_potential.py
```diff
@@ -308,5 +309,281 @@ def test_fromfile(self):
...
+
+        #test with shift set
+        p = test_PairPotential.LinPot(types=('1',), params=('m','rmax','shift'),
+                                      default={'m':2,'rmax':1.0,'shift':True})
+        d = p.derivative(pair=(('1','1')), param='m', r=0.5)
+        self.assertEqual(d, 0.5)
+        d = p.derivative(pair=(('1','1')), param='m', r=[1.0,1.5])
+        np.testing.assert_allclose(d, [1.0,0.0])
+
+    def test_iteration(self):
```
mphoward, line 561: What is the goal of this test? It seems like it's duplicating the CoefficientMatrix tests since everything is testing on `p.coeff` rather than `p` itself?

relentless/potential/potential.py
```diff
@@ -294,48 +312,91 @@ def __init__(self, types, params, default={}):
...
+        ----------
+        pair : array_like
+            The pair for which to calculate the energy.
+        r : array_like
+            The location(s) at which to evaluate the energy.
+
+        Returns
+        -------
+        array_like
+            The energy at the specified location(s).
```
mphoward, line 328: You might want to clarify that the returned quantity will have the same form as `r`. So, if r is a scalar, it will be a scalar, but if r is array_like, it will be a numpy array.

relentless/potential/potential.py
```diff
@@ -294,48 +312,91 @@ def __init__(self, types, params, default={}):
...
+        ValueError
+            If any value in `r` is negative.
+        ValueError
+            If the potential is to be shifted without setting `rmax`.
+
+        """
+        if np.min(r) < 0:
+            raise ValueError('r cannot be negative')
         params = self.coeff.evaluate(pair)
         r,u,scalar_r = self._zeros(r)
```
mphoward, line 341: I would move the check on r here, and just use:
```py
any(r < 0)
```
since it is easier to read.

You can move the `evaluate` after this check, if you only want to do this after you do the error checking.

relentless/potential/potential.py
```diff
@@ -294,48 +312,91 @@ def __init__(self, types, params, default={}):
...
                 u[r > params['rmax']] = self._energy(params['rmax'], **params)
+        elif params['shift'] is True:
+            raise ValueError('Cannot shift potential without rmax')
 
         # coerce u back into shape of the input
         if scalar_r:
             u = u.item()
         return u
 
     def force(self, pair, r):
```
mphoward, line 369: The comments on `energy` also apply to `force` and `derivative`.

relentless/potential/potential.py
```diff
@@ -363,27 +447,67 @@ def derivative(self, pair, param, r):
         return deriv
 
     def _zeros(self, r):
-        # coerce input shape and create zeros for output
+        """Coerces input shape and creates zeros for output.
```
mphoward, line 450: This was a silly description by me, let's make it:
"Force input to a one-dimensional array and make matching array of zeros."

tests/potential/test_potential.py
```diff
@@ -308,5 +309,272 @@ def test_fromfile(self):
...
+                f = f.item()
+            return f
+
+        def _derivative(self, param, r, **params):
+            r,d,s = self._zeros(r)
+            if param == 'm':
+                d = r
+            if s:
+                d = d.item()
+            return d
```
mphoward, line 340: You're probably going to need this class to test the `Tabulator` later. You can refer to it there as `test_PairPotential.LinPot`, or you can move this class outside the unittest, so it is accessible directly as `LinPot`.

tests/potential/test_potential.py
```diff
@@ -308,5 +309,272 @@ def test_fromfile(self):
...
+        #test with rmin set
+        p.coeff['1','1']['rmin'] = 0.5
+        u = p.energy(pair=('1','1'), r=0.6)
+        self.assertAlmostEqual(u, 1.2)
+        u = p.energy(pair=('1','1'), r=[0.25,0.75])
+        np.testing.assert_allclose(u, [1.0,1.5])
+
+        #test with rmax set
+        p.coeff['1','1']['rmin'] = False
+        p.coeff['1','1']['rmax'] = 1.5
```
mphoward, line 457: Just to mess with the methods as much as possible, let's do this with update:
```suggestion
p.coeff['1','1'].update(rmin=False, rmax=1.5)
```

tests/potential/test_potential.py
```diff
@@ -308,5 +309,272 @@ def test_fromfile(self):
...
+        u = p.energy(pair=('1','1'), r=[0.25,0.5,1.5,1.75])
+        np.testing.assert_allclose(u, [1.0,1.0,3.0,3.0])
+
+        #test with shift set
+        p.coeff['1','1']['shift'] = True
+        u = p.energy(pair=('1','1'), r=0.5)
+        self.assertAlmostEqual(u, -2.0)
+        u = p.energy(pair=('1','1'), r=[1.0,1.5])
+        np.testing.assert_allclose(u, [-1.0,0.0])
+
```
mphoward, line 476: Could you add a short test that an error is raised if `shift == True and rmax==False`?

tests/potential/test_potential.py
```diff
@@ -308,5 +309,272 @@ def test_fromfile(self):
...
+        u = p.energy(pair=('1','1'), r=0.75)
+        self.assertAlmostEqual(u, 1.5)
+        u = p.energy(pair=('1','1'), r=[0.25,0.5,1.5,1.75])
+        np.testing.assert_allclose(u, [1.0,1.0,3.0,3.0])
+
+        #test with shift set
+        p.coeff['1','1']['shift'] = True
+        u = p.energy(pair=('1','1'), r=0.5)
+        self.assertAlmostEqual(u, -2.0)
+        u = p.energy(pair=('1','1'), r=[1.0,1.5])
```
mphoward, line 474: Could you test this for a few more r values, including below `rmin` and above `rmax` to make sure these cases are done correctly?

relentless/potential/potential.py
```diff
@@ -2,6 +2,7 @@
 
 import json
 import numpy as np
+import abc
```
mphoward, line 5: Sorry, forgot to also mention: please add imports in alphabetical order, so `abc` would come before `json`.

PR #21, opened 2020-07-13
Documentation, features, and unit tests for Tabulator

adisr04:
- Add Tabulator documentation
- Simplify construction of Tabulator objects/attributes
- Modify regularize function in Tabulator
- Add all unit tests for Tabulator in test_potential

### DIFF
--- a/relentless/potential/potential.py
+++ b/relentless/potential/potential.py
@@ -3,6 +3,7 @@ __all__ = ['PairPotential','Tabulator']
 import abc
 import json
 import numpy as np
+import warnings
 
 from relentless.core import PairMatrix
 from relentless.core import FixedKeyDict
@@ -529,7 +530,44 @@ class PairPotential(abc.ABC):
         return next(self.coeff)
 
 class Tabulator:
+    """Creates an object that organizes multiple potential functions together.
+
+    Bins accumulated energy and force values for multiple potential functions,
+    allows "regularization" of the force (and energy).
+
+    Parameters
+    ----------
+    nbins : int
+        The number of bins in which to place the values for the potential.
+    rmin : float
+        The minimum r value at which to calculate energy and force.
+    rmax : float
+        The maximum r value at which to calculate energy and force.
+    fmax : float
+        (Optional) The value of the force at which to truncate the maximum force at small r values.
+    fcut : float
+        (Optional) The value of the force at which to truncate the minimum force at large r values.
+    edges : bool
+        Whether to include the edges (rmin and rmax) in the bins, defaults to `True`.
+
+    Raises
+    ------
+    ValueError
+        If nbins is not a positive integer.
+    ValueError
+        If rmin is greater than or equal to rmax.
+    ValueError
+        If rmin is not positive.
+
+    """
     def __init__(self, nbins, rmin, rmax, fmax=None, fcut=None, edges=True):
+        if not isinstance(nbins, int) and nbins > 0:
+            raise ValueError('nbins must be a positive integer')
+        if rmin >= rmax:
+            raise ValueError('rmin must be less than rmax')
+        if rmin < 0:
+            raise ValueError('rmin must be positive')
+
         self._nbins = nbins
         self._rmin = rmin
         self._rmax = rmax
@@ -545,13 +583,30 @@ class Tabulator:
 
     @property
     def dr(self):
+        """float: The interval for r between each bin."""
         return self._dr
 
     @property
     def r(self):
+        """array_like: The r values at each bin."""
         return self._r
 
     def energy(self, pair, potentials):
+        """Accumulates and bins energy values for all potentials, for the specified pair.
+
+        Parameters
+        ----------
+        pair : tuple
+            The type pair (i,j) for which to calculate the energy.
+        potentials : array_like
+            All the potential functions for which to calculate and accumulate the energy.
+
+        Returns
+        -------
+        array_like
+            Returns the array of binned, accumulated energy values.
+
+        """
         u = np.zeros_like(self.r)
         for pot in potentials:
             try:
@@ -561,6 +616,21 @@ class Tabulator:
         return u
 
     def force(self, pair, potentials):
+        """Accumulates and bins force values for all potentials, for the specified pair.
+
+        Parameters
+        ----------
+        pair : tuple
+            The type pair (i,j) for which to calculate the force.
+        potentials : array_like
+            All the potential functions for which to calculate and accumulate the force.
+
+        Returns
+        -------
+        array_like
+            Returns the array of binned, accumulated force values.
+
+        """
         f = np.zeros_like(self.r)
         for pot in potentials:
             try:
@@ -569,7 +639,40 @@ class Tabulator:
                 pass
         return f
 
-    def regularize(self, u, f, trim=True):
+    def regularize_force(self, u, f, trim=True):
+        """Adjusts the accumulated force values based on the values of fmax and fcut.
+
+        Based on the adjustments made to the force, the energy values are also adjusted.
+
+        Parameters
+        ----------
+        u : array_like
+            The binned energy values.
+        f : array_like
+            The binned force values.
+        trim : bool
+            Whether to trim off trailing zeros from the regularized force/energy values,
+            defaults to `True`.
+
+        Returns
+        -------
+        float or `None`
+            The value of r at which trimming trailing zeros is implemented or could be
+            implemented (if trim is False). Returns `None` if trimming cannot be performed.
+        array_like
+            Stacks the r, adjusted u, and adjusted f as columns into a `nx3` array for n bins.
+
+        Raises
+        ------
+        IndexError
+            If the energy array is not the same length as the array of r values.
+        IndexError
+            If the force array is not the same length as the array of r values.
+        UserWarning
+            If rmax is too small to cutoff the potential, i.e. if the force at the end
+            of the potential is larger than fcut.
+
+        """
         if len(u) != len(self.r):
             raise IndexError('Potential must have the same length as r.')
         if len(f) != len(self.r):
@@ -577,28 +680,34 @@ class Tabulator:
 
         # find first point from beginning that is within energy tolerance
         if self.fmax is not None:
-            cut = np.argmax(np.abs(f) <= self.fmax)
+            cut = np.argmax(f <= self.fmax)
             if cut > 0:
                 u[:cut] = u[cut] - f[cut]*(self.r[:cut] - self.r[cut])
                 f[:cut] = f[cut]
 
         # find first point from end with sufficient force and cutoff the potential after it
         if self.fcut is not None:
-            flags = np.abs(np.flip(f)) >= self.fcut
+            flags = np.flip(f) >= self.fcut
             cut = len(f)-1 - np.argmax(flags)
-            u -= u[cut]
+            u -= u[cut]                         #would this be considered 'regularizing' the energy?
+            if cut == len(f)-1:
+                if f[-1] > self.fcut:
+                    warnings.warn('rmax is too small to cutoff the potential.', UserWarning)
             if cut < len(f)-1:
                 u[(cut+1):] = 0.
                 f[(cut+1):] = 0.
 
         # trim off trailing zeros
         r = self.r.copy()
-        if trim:
-            flags = np.abs(np.flip(f)) > 0
-            cut = len(f) - np.argmax(flags)
-            if cut < len(f)-1:
+        flags = np.abs(np.flip(f)) > 0
+        cut = len(f) - np.argmax(flags)
+        if cut < len(f)-1:
+            rcut = r[cut]
+            if trim:
                 r = r[:(cut+1)]
                 u = u[:(cut+1)]
                 f = f[:(cut+1)]
+        else:
+            rcut = None
 
-        return np.column_stack((r,u,f))
+        return rcut, np.column_stack((r,u,f))
--- a/relentless/potential/potential.py
+++ b/relentless/potential/potential.py
@@ -2,8 +2,9 @@ __all__ = ['PairPotential','Tabulator']
 
 import abc
 import json
-import numpy as np
 import warnings
+
+import numpy as np
 
 from relentless.core import PairMatrix
 from relentless.core import FixedKeyDict
@@ -530,10 +531,10 @@ class PairPotential(abc.ABC):
         return next(self.coeff)
 
 class Tabulator:
-    """Creates an object that organizes multiple potential functions together.
+    """Combines and tabulates multiple potentials together.
 
-    Bins accumulated energy and force values for multiple potential functions,
-    allows "regularization" of the force (and energy).
+    Evaluates accumulated energy and force values for multiple potential functions at different r values,
+    allows regularization of the force and shifting of the energy.
 
     Parameters
     ----------
@@ -544,11 +545,13 @@ class Tabulator:
     rmax : float
         The maximum r value at which to calculate energy and force.
     fmax : float
-        (Optional) The value of the force at which to truncate the maximum force at small r values.
+        (Optional) The maximum magnitude of the force for regularization at small r values.
     fcut : float
-        (Optional) The value of the force at which to truncate the minimum force at large r values.
+        (Optional) The magnitude of the force for truncation at large r values.
     edges : bool
-        Whether to include the edges (rmin and rmax) in the bins, defaults to `True`.
+        If `True`, evaluate r at edges of the bins; otherwise, evaluate at the centers (defaults to `True`).
+    shift : bool
+        If 'True', shift the potential (according to value of fcut) (defaults to `True`).
 
     Raises
     ------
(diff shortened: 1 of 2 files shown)


relentless/potential/potential.py
```diff
@@ -3,6 +3,7 @@
 import abc
 import json
 import numpy as np
+import warnings
```
mphoward, line 6: PEP8 (https://www.python.org/dev/peps/pep-0008/#imports) recommends that you import in the following order:
1. Standard library
2. Third-party
3. Local
with a blank line between each of them.

So here:
```suggestion
import warnings

import numpy as np
```

relentless/potential/potential.py
```diff
@@ -529,7 +530,44 @@ def __next__(self):
         return next(self.coeff)
 
 class Tabulator:
+    """Creates an object that organizes multiple potential functions together.
```
mphoward, line 533:
```suggestion
    """Combines and tabulates multiple potentials together.
```

relentless/potential/potential.py
```diff
@@ -529,7 +530,44 @@ def __next__(self):
...
+
+    Bins accumulated energy and force values for multiple potential functions,
```
mphoward, line 535: Bins -> evaluates throughout this code. ("Binning" is used for histogramming. You are just evaluating at a set of r values.)

relentless/potential/potential.py
```diff
@@ -529,7 +530,44 @@ def __next__(self):
...
+    Parameters
+    ----------
+    nbins : int
+        The number of bins in which to place the values for the potential.
+    rmin : float
+        The minimum r value at which to calculate energy and force.
+    rmax : float
+        The maximum r value at which to calculate energy and force.
+    fmax : float
+        (Optional) The value of the force at which to truncate the maximum force at small r values.
```
mphoward, line 547:
```suggestion
        (Optional) The maximum magnitude of the force for regularization at small r values.
```

relentless/potential/potential.py
```diff
@@ -529,7 +530,44 @@ def __next__(self):
...
+    fcut : float
+        (Optional) The value of the force at which to truncate the minimum force at large r values.
```
mphoward, line 549:
```suggestion
        (Optional) The magnitude of the force for truncation at large r values.
```

relentless/potential/potential.py
```diff
@@ -529,7 +530,44 @@ def __next__(self):
...
+    edges : bool
+        Whether to include the edges (rmin and rmax) in the bins, defaults to `True`.
```
mphoward, line 551:
```suggestion
        If `True`, evaluate `r` at edges of the bins; otherwise, evaluate at the centers (defaults to `True`).
```

relentless/potential/potential.py
```diff
@@ -545,13 +583,30 @@ def __init__(self, nbins, rmin, rmax, fmax=None, fcut=None, edges=True):
...
+        """float: The interval for r between each bin."""
         return self._dr
 
     @property
     def r(self):
+        """array_like: The r values at each bin."""
         return self._r
 
     def energy(self, pair, potentials):
+        """Accumulates and bins energy values for all potentials, for the specified pair.
```
mphoward, line 595:
```suggestion
        """Evaluates and accumulates energy for all potentials, for the specified pair.
```

relentless/potential/potential.py
```diff
@@ -545,13 +583,30 @@ def __init__(self, nbins, rmin, rmax, fmax=None, fcut=None, edges=True):
...
+        ----------
+        pair : tuple
+            The type pair (i,j) for which to calculate the energy.
+        potentials : array_like
+            All the potential functions for which to calculate and accumulate the energy.
+
+        Returns
+        -------
+        array_like
+            Returns the array of binned, accumulated energy values.
```
mphoward, line 607:
```suggestion
            Total energy at each `r`.
```

relentless/potential/potential.py
```diff
@@ -561,6 +616,21 @@ def energy(self, pair, potentials):
         return u
 
     def force(self, pair, potentials):
```
mphoward, line 618: Apply the same doc changes here as above.

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
                 pass
         return f
 
-    def regularize(self, u, f, trim=True):
+    def regularize_force(self, u, f, trim=True):
+        """Adjusts the accumulated force values based on the values of fmax and fcut.
```
mphoward, line 643:
```suggestion
        """Regularizes and truncates the accumulated energies and forces.
```

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
+
+        Based on the adjustments made to the force, the energy values are also adjusted.
+
+        Parameters
+        ----------
+        u : array_like
+            The binned energy values.
```
mphoward, line 650:
```suggestion
            Energies, must have same shape as `r`.
```

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
+        f : array_like
+            The binned force values.
```
mphoward, line 652:
```suggestion
            Forces, must have same shape as `r`.
```

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
+        IndexError
+            If the energy array is not the same length as the array of r values.
+        IndexError
+            If the force array is not the same length as the array of r values.
+        UserWarning
+            If rmax is too small to cutoff the potential, i.e. if the force at the end
+            of the potential is larger than fcut.
+
+        """
         if len(u) != len(self.r):
```
mphoward, line 676: Probably here, we should first do:
```py
u = np.atleast_1d(u)
f = np.atleast_1d(f)
```
Then replace these checks by:
```py
if u.shape != self.r.shape:
   ...
```

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
             raise IndexError('Potential must have the same length as r.')
         if len(f) != len(self.r):
             raise IndexError('Force must have the same length as r.')
 
         # find first point from beginning that is within energy tolerance
         if self.fmax is not None:
-            cut = np.argmax(np.abs(f) <= self.fmax)
+            cut = np.argmax(f <= self.fmax)
```
mphoward, line 683: Is there a reason you remove the absolute value? I think we want to keep this here so that we regularize both large repulsive and large attractive forces (i.e., large |F| is bad, not just F >> 0).

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
             if cut > 0:
                 u[:cut] = u[cut] - f[cut]*(self.r[:cut] - self.r[cut])
                 f[:cut] = f[cut]
 
         # find first point from end with sufficient force and cutoff the potential after it
         if self.fcut is not None:
-            flags = np.abs(np.flip(f)) >= self.fcut
+            flags = np.flip(f) >= self.fcut
```
mphoward, line 690: Similarly here, we also want to be able to cutoff when |F| is small, regardless of sign.

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
             cut = len(f)-1 - np.argmax(flags)
-            u -= u[cut]
+            u -= u[cut]                         #would this be considered 'regularizing' the energy?
```
mphoward, line 692: This is "shifting" the energy, so you should probably document that above somewhere.

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
+            if cut == len(f)-1:
+                if f[-1] > self.fcut:
+                    warnings.warn('rmax is too small to cutoff the potential.', UserWarning)
             if cut < len(f)-1:
                 u[(cut+1):] = 0.
                 f[(cut+1):] = 0.
```
mphoward, line 698:
```suggestion
            if cut < len(f)-1:
                u[(cut+1):] = 0.
                f[(cut+1):] = 0.
            elif f[-1] > self.fcut:
                warnings.warn('Last tabulated force exceeds fcut, rmax may be too small.', UserWarning)
```

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
+        flags = np.abs(np.flip(f)) > 0
+        cut = len(f) - np.argmax(flags)
+        if cut < len(f)-1:
+            rcut = r[cut]
+            if trim:
                 r = r[:(cut+1)]
                 u = u[:(cut+1)]
                 f = f[:(cut+1)]
+        else:
+            rcut = None
```
mphoward, line 711: Should this return `None`, or the last value tabulated?

relentless/potential/potential.py
```diff
@@ -569,36 +639,75 @@ def force(self, pair, potentials):
...
 
-        return np.column_stack((r,u,f))
+        return rcut, np.column_stack((r,u,f))
```
mphoward, line 713: Let's flip the return order
```suggestion
        return np.column_stack((r,u,f)), rcut
```
It's a little easier to read if the user wants to ignore the second value like,
```py
pot,_ = tab.regularize_force(...)
```

tests/potential/test_potential.py
```diff
@@ -2,7 +2,8 @@
 import numpy as np
 import tempfile
 import unittest
-
+import sys
+sys.path.append('../../')
```
mphoward, line 6: Not needed, remove.

tests/potential/test_potential.py
```diff
@@ -231,10 +232,6 @@ def test_saveload(self):
 
         self.assertEqual(m['A','B']['energy'].low, x['A','B']['energy'].low)
         self.assertEqual(m['A','B']['mass'].low, x['A','B']['mass'].low)
-        self.assertEqual(m['A','A']['energy'].low, x['A','A']['energy'].low)
-        self.assertEqual(m['A','A']['mass'].low, x['A','A']['mass'].low)
-        self.assertEqual(m['B','B']['energy'].low, x['B','B']['energy'].low)
-        self.assertEqual(m['B','B']['mass'].low, x['B','B']['mass'].low)
```
mphoward, line 237: Why were these checks removed?

tests/potential/test_potential.py
```diff
@@ -581,5 +577,195 @@ def test_saveload(self):
...
+        f = t.force(pair=('1','1'), potentials=p_all)
+        rcut,stack = t.regularize_force(u, f)
+        self.assertEqual(rcut, None)
+        np.testing.assert_allclose(stack, np.array([[1, 0,  0],
+                                                    [2, 3, -6],
+                                                    [3,12,-12],
+                                                    [4,27,-18],
+                                                    [5,48,-24]]))
+        #test with only fmax
+        t = potential.Tabulator(nbins=4, rmin=1.0, rmax=5.0, fmax=-10)
```
mphoward, line 701: I sort of noted this above, but we will want to make `fmax` and `fcut` >= 0. Can you modify these tests accordingly, and also add an error check for this in `__init__`?

relentless/potential/potential.py
```diff
@@ -569,10 +652,46 @@ def force(self, pair, potentials):
...
+            Whether to trim off trailing zeros from the regularized force/energy values,
+            defaults to `True`.
+
+        Returns
+        -------
+        float or `None`
+            The value of r at which trimming trailing zeros is implemented or could be
+            implemented (if trim is False). Returns the last r value if trimming cannot be performed.
+        array_like
+            Stacks the r, adjusted u, and adjusted f as columns into a `nx3` array for n bins.
```
mphoward, line 677: These docs need to be reordered and updated.

relentless/potential/potential.py
```diff
@@ -584,21 +703,31 @@ def regularize(self, u, f, trim=True):
...
-            if cut < len(f)-1:
-                u[(cut+1):] = 0.
-                f[(cut+1):] = 0.
+            flags = np.abs(np.flip(f)) <= self.fcut
+            cut = len(f) - np.argmax(flags)
+            if cut < len(f):
+                if self._shift:
+                    u -= u[cut]
+                u[cut:] = 0.
+                f[cut:] = 0.
```
mphoward, line 712: What errors were you running into that you needed to change this logic so much? I see why you needed to change the indexes so that u is actually 0 _at_ cut, instead of being 0 at the _next_ index. I can't remember why I did it the other way before, but this seems like it should work just fine.

Based on what we discussed, though, I think this should actually be `>` not `<=` for the flags, as I think argmax is always going to return 0 if `<=` is used.

relentless/potential/potential.py
```diff
@@ -543,15 +592,34 @@ def __init__(self, nbins, rmin, rmax, fmax=None, fcut=None, edges=True):
         else:
             self._r = rmin + self._dr*(np.arange(nbins, dtype=np.float64)+0.5)
 
+        self._shift = shift
```
mphoward, line 595: Let's make this publicly accessible as `self.shift`. It's OK if the user wants to switch it between True or False directly.

relentless/potential/potential.py
```diff
@@ -529,7 +531,54 @@ def __next__(self):
...
-    def __init__(self, nbins, rmin, rmax, fmax=None, fcut=None, edges=True):
+    """Combines and tabulates multiple potentials together.
+
+    Evaluates accumulated energy and force values for multiple potential functions at different r values,
+    allows regularization of the force and shifting of the energy.
+
+    Parameters
+    ----------
+    nbins : int
+        The number of bins in which to place the values for the potential.
```
mphoward, line 542: I realize now why you used "bins" throughout... this was a bad name choice on my part! I probably borrowed it from an RDF code or numpy.histogram. I think we should actually give the user more flexibility here. Rather than specifying `nbins`, `rmin`, `rmax`, and `edges`, let's just let the user give an array like `r`. Internally, copy it into a numpy array, exposed using getters and setters:
```py
@property
def r(self):
    return self._r
@r.setter
def r(self, points):
    points = np.array(points)
    # error check shape and order here
    self._r = points
```
The error check should be that `r` is a 1d array and that its entries are continuously increasing. In the constructor, you just need:
```py
self.r = r
```
Then when constructing the `Tabulator`, the user can use `linspace` or whatever method they like (for example, if they want irregular spaced points).

relentless/potential/potential.py
```diff
@@ -543,15 +592,34 @@ def __init__(self, nbins, rmin, rmax, fmax=None, fcut=None, edges=True):
         else:
             self._r = rmin + self._dr*(np.arange(nbins, dtype=np.float64)+0.5)
 
+        self.shift = shift
+
     @property
```
mphoward, line 597: With the proposed change, this property should be removed.

relentless/potential/potential.py
```diff
@@ -529,7 +531,54 @@ def __next__(self):
...
+        if rmin < 0:
+            raise ValueError('rmin must be positive')
+        if fmax is not None and fmax <= 0:
+            raise ValueError('fmax must be positive')
+        if fcut is not None and fcut < 0:
+            raise ValueError('fcut must be positive')
+
         self._nbins = nbins
         self._rmin = rmin
         self._rmax = rmax
```
mphoward, line 584: With the proposed change, these can be removed.

relentless/potential/potential.py
```diff
@@ -529,7 +531,54 @@ def __next__(self):
...
+        If fcut is set, and it is not positive.
+
+    """
+    def __init__(self, nbins, rmin, rmax, fmax=None, fcut=None, edges=True, shift=True):
+        if not isinstance(nbins, int) and nbins > 0:
+            raise ValueError('nbins must be a positive integer')
+        if rmin >= rmax:
+            raise ValueError('rmin must be less than rmax')
+        if rmin < 0:
+            raise ValueError('rmin must be positive')
```
mphoward, line 576: With the proposed change, these can be removed.

relentless/potential/potential.py
```diff
@@ -586,19 +705,29 @@ def regularize(self, u, f, trim=True):
...
+            u -= u[-1]
 
         # trim off trailing zeros
         r = self.r.copy()
-        if trim:
-            flags = np.abs(np.flip(f)) > 0
-            cut = len(f) - np.argmax(flags)
-            if cut < len(f)-1:
+        flags = np.abs(np.flip(f)) > 0
+        cut = len(f)-1 - np.argmax(flags)
```
mphoward, line 723: I think this might need this to go back to
```suggestion
        cut = len(f) - np.argmax(flags)
```
So that we always include the last point where the force and energy are both zero.

Right now, I think this will throw out all of the points where the force is zero, right? By retaining one point where the force is zero, we ensure that the tabulated force and energy both interpolate to zero at the end point.

The rest should then work as is.

@zeesherman what do you think?

tests/potential/test_potential.py
```diff
@@ -581,5 +581,211 @@ def test_saveload(self):
...
+
+    def _derivative(self, param, r, **params):
+        r,d,s = self._zeros(r)
+        if param == 'm':
+            d = (3-r)**2
+        if s:
+            d = d.item()
+        return d
+
+class test_Tabulator(unittest.TestCase):
```
mphoward, line 612: If you make the change I proposed, these tests will need to be cleaned up a bit, so I will defer looking through them carefully until then. They look good overall though!

tests/potential/test_potential.py
```diff
@@ -581,5 +581,211 @@ def test_saveload(self):
...
+
+    def test_potential(self):
+        """Test energy and force methods"""
+        p1 = QuadPot(types=('1',), params=('m',), default={'m':2})
+        p2 = QuadPot(types=('1','2'), params=('m',), default={'m':1})
+        p_all = [p1, p2]
+
+        #test energy method
+        #test with edges=True
+        t = potential.Tabulator(nbins=4, rmin=1.0, rmax=5.0)
```
mphoward, line 661: With the proposed changes, it should be possible for you to make just one instance of the tabulator (you don't need to test `edges` anymore), and then reuse it for the tests. I'd like to see this tried, since it is a more realistic use case then making a new one each time.

tests/potential/test_potential.py
```diff
@@ -581,5 +581,211 @@ def test_saveload(self):
...
+        np.testing.assert_allclose(f, np.array([3,1,-1,-3]))
+
+    def test_regularize(self):
+        """Test regularize_force method cases"""
+        p1 = QuadPot(types=('1',), params=('m',), default={'m':2})
+        p2 = QuadPot(types=('1','2'), params=('m',), default={'m':1})
+        p_all = [p1, p2]
+
+        #test without fmax or fcut
+        t = potential.Tabulator(nbins=4, rmin=0.5, rmax=2.5)
```
mphoward, line 694: Similarly down here, can you try making just one tabulator and then setting its properties directly?

PR #22, opened 2020-07-15
Documentation, features, and unit tests for LennardJones

adisr04:
- Add LennardJones documentation
- Fix input sanitization in _zeros method of PairPotential, energy/force methods of Tabulator
- Add all unit tests for LennardJones in test_pair

### DIFF
--- a/relentless/potential/pair.py
+++ b/relentless/potential/pair.py
@@ -6,9 +6,22 @@ from scipy.interpolate import Akima1DInterpolator
 from .potential import PairPotential
 
 class LennardJones(PairPotential):
-    """Models particle pair interactions using the Lennard-Jones potential.
+    """Lennard-Jones 12-6 pair potential.
 
-    Two parameters are used: epsilon and sigma.
+    .. math::
+
+    u(r) = 4 \varepsilon\left[\left(\frac{\sigma}{r}\right)^12 - \left(\frac{\sigma}{r}\right)^12 \right]
+
+    The required coefficients per pair are:
+
+    - :math:`\varepsilon`: interaction energy
+    - :math:`\sigma`: interaction length scale (e.g., particle diameter)
+
+    The optional coefficients per pair are:
+
+    - ``rmin``: minimum radius, energy and force are 0 for ``r < rmin``. Ignored if ``False`` (default).
+    - ``rmax``: maximum radius, energy and force are 0 for ``r > rmax`` Ignored if ``False`` (default).
+    - ``shift``: If ``True``, shift potential to zero at ``rmax`` (default is ``False``).
 
     Parameters
     ----------
@@ -20,7 +33,7 @@ class LennardJones(PairPotential):
         super().__init__(types=types, params=('epsilon','sigma'))
 
     def _energy(self, r, epsilon, sigma, **params):
-        """Evaluates the energy using the L-J potential at a certain r-value.
+        """Evaluates the Lennard-Jones potential energy.
 
         Parameters
         ----------
@@ -57,7 +70,7 @@ class LennardJones(PairPotential):
         return u
 
     def _force(self, r, epsilon, sigma, **params):
-        """Evaluates the force using the L-J potential at a certain r-value.
+        """Evaluates the Lennard-Jones force.
 
         Parameters
         ----------
@@ -95,14 +108,14 @@ class LennardJones(PairPotential):
         return f
 
     def _derivative(self, param, r, epsilon, sigma, **params):
-        """Evaluates the derivative of the L-J potential at a certain r-value, for a specified parameter.
+        """Evaluates the Lennard-Jones parameter derivative.
 
         Parameters
         ----------
-        r : array_like
-            The value or values of r at which to evaluate the derivative.
         param : `str`
             The parameter with respect to which to take the derivative.
+        r : array_like
+            The value or values of r at which to evaluate the derivative.
         epsilon : float or int
             The epsilon parameter for the potential function.
         sigma : float or int
@@ -133,7 +146,7 @@ class LennardJones(PairPotential):
         elif param == 'sigma':
             d[flags] = (48.*epsilon/sigma)*(r6_inv**2 - 0.5*r6_inv)
         else:
-            raise ValueError('The parameter with respect to which to take the derivative must be "sigma" or "epsilon"')
+            raise ValueError('The Lennard-Jones parameters are sigma and epsilon.')
         d[~flags] = np.inf
 
         if s:
--- a/relentless/potential/pair.py
+++ b/relentless/potential/pair.py
@@ -6,10 +6,44 @@ from scipy.interpolate import Akima1DInterpolator
 from .potential import PairPotential
 
 class LennardJones(PairPotential):
+    """Models particle pair interactions using the Lennard-Jones potential.
+
+    Two parameters are used: epsilon and sigma.
+
+    Parameters
+    ----------
+    types : array_like
+        List of types (A type must be a `str`).
+
+    """
     def __init__(self, types):
         super().__init__(types=types, params=('epsilon','sigma'))
 
     def _energy(self, r, epsilon, sigma, **params):
+        """Evaluates the energy using the L-J potential at a certain r-value.
+
+        Parameters
+        ----------
+        r : array_like
+            The value or values of r at which to evaluate the energy.
+        epsilon : float or int
+            The epsilon parameter for the potential function.
+        sigma : float or int
+            The sigma parameter for the potential function.
+
+        Returns
+        -------
+        array_like
+            Returns the value or values of the energy evaluated at r.
+
+        Raises
+        ------
+        ValueError
+            If sigma is not a positive number.
+
+        """
+        if sigma < 0:
+            raise ValueError('sigma must be positive')
         r,u,s = self._zeros(r)
         flags = ~np.isclose(r,0)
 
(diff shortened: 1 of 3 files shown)


relentless/potential/pair.py
```diff
@@ -6,10 +6,44 @@
 from .potential import PairPotential
 
 class LennardJones(PairPotential):
+    """Models particle pair interactions using the Lennard-Jones potential.
```
mphoward, line 9:
```suggestion
    """Lennard-Jones 12-6 pair potential.
```

relentless/potential/pair.py
```diff
@@ -6,10 +6,44 @@
...
+
+    Two parameters are used: epsilon and sigma.
```
mphoward, line 11: We should expand this a bit more to include the equation and some more details. Maybe something like:

```py
.. math::

    u(r) = 4 \varepsilon\left[\left(\frac{\sigma}{r}\right)^12 - \left(\frac{\sigma}{r}\right)^12 \right]

The required coefficients per pair are:

- :math:`\varepsilon`: interaction energy
- :math:`\sigma`: interaction length scale (e.g., particle diameter)

The optional coefficients per pair are:

- ``rmin``: minimum radius, energy and force are 0 for ``r < rmin``. Ignored if ``False`` (default).
- ``rmax``: maximum radius, energy and force are 0 for ``r > rmax`` Ignored if ``False`` (default).
- ``shift``: If ``True``, shift potential to zero at ``rmax`` (default is ``False``).
```

relentless/potential/pair.py
```diff
@@ -6,10 +6,44 @@
...
+    ----------
+    types : array_like
+        List of types (A type must be a `str`).
+
+    """
     def __init__(self, types):
         super().__init__(types=types, params=('epsilon','sigma'))
 
     def _energy(self, r, epsilon, sigma, **params):
+        """Evaluates the energy using the L-J potential at a certain r-value.
```
mphoward, line 23:
```suggestion
        """Evaluates the Lennard-Jones potential energy.
```

relentless/potential/pair.py
```diff
@@ -23,6 +57,30 @@ def _energy(self, r, epsilon, sigma, **params):
         return u
 
     def _force(self, r, epsilon, sigma, **params):
+        """Evaluates the force using the L-J potential at a certain r-value.
```
mphoward, line 60:
```suggestion
        """Evaluates the Lennard-Jones force.
```

relentless/potential/pair.py
```diff
@@ -37,15 +95,45 @@ def _force(self, r, epsilon, sigma, **params):
         return f
 
     def _derivative(self, param, r, epsilon, sigma, **params):
+        """Evaluates the derivative of the L-J potential at a certain r-value, for a specified parameter.
```
mphoward, line 98:
```suggestion
        """Evaluates the Lennard-Jones parameter derivative.
```

relentless/potential/pair.py
```diff
@@ -37,15 +95,45 @@ def _force(self, r, epsilon, sigma, **params):
...
+
+        Parameters
+        ----------
+        r : array_like
+            The value or values of r at which to evaluate the derivative.
+        param : `str`
+            The parameter with respect to which to take the derivative.
```
mphoward, line 105: Flip the order of these args.

relentless/potential/pair.py
```diff
@@ -37,15 +95,45 @@ def _force(self, r, epsilon, sigma, **params):
...
 
         # evaluate derivative
         r6_inv = np.power(sigma/r[flags], 6)
         if param == 'epsilon':
             d[flags] = 4*(r6_inv**2 - r6_inv)
-        elif param == 'sigma' and sigma > 0:
+        elif param == 'sigma':
             d[flags] = (48.*epsilon/sigma)*(r6_inv**2 - 0.5*r6_inv)
+        else:
+            raise ValueError('The parameter with respect to which to take the derivative must be "sigma" or "epsilon"')
```
mphoward, line 136:
```suggestion
            raise ValueError('The Lennard-Jones parameters are sigma and epsilon.')
```

relentless/potential/potential.py
```diff
@@ -598,6 +598,7 @@ def energy(self, pair, potentials):
 
         """
         u = np.zeros_like(self.r)
+        u = u.astype(dtype=np.float64)
```
mphoward, line 601: This will result in a double-copy of `r`, so we should do the longer form:
```suggestion
        u = np.zeros(self.r.shape, dtype=np.float64)
```

relentless/potential/potential.py
```diff
@@ -622,6 +623,7 @@ def force(self, pair, potentials):
 
         """
         f = np.zeros_like(self.r)
+        f = f.astype(dtype=np.float64)
```
mphoward, line 626: Same thing here.

tests/potential/test_pair.py
```diff
@@ -0,0 +1,147 @@
...
+        self.assertAlmostEqual(lj.coeff['1','1']['sigma'], 1.5)
+        self.assertAlmostEqual(lj.coeff['1','2']['epsilon'], 2.5)
+        self.assertEqual(lj.coeff['1','2']['sigma'], None)
+
+        #test setitem with default reset
+        lj.coeff['1','2'] = {'sigma':0.5}
+        self.assertAlmostEqual(lj.coeff['1','1']['epsilon'], 2.0)
+        self.assertAlmostEqual(lj.coeff['1','1']['sigma'], 1.5)
+        self.assertEqual(lj.coeff['1','2']['epsilon'], None)
+        self.assertAlmostEqual(lj.coeff['1','2']['sigma'], 0.5)
```
mphoward, line 45: You can leave this test if you want (since it's already written), but it is probably not necessary because you already have good coverage of the `CoefficientMatrix` elsewhere and you've already checked that `lj.coeff` is getting the right parameters in `test_init`.

tests/potential/test_pair.py
```diff
@@ -0,0 +1,147 @@
...
+        s11 = lj.coeff['1','1']['sigma']
+        e11 = lj.coeff['1','1']['epsilon']
+
+        lj.coeff['1','2'].update(epsilon=1.0, sigma=0.5)
+        s12 = lj.coeff['1','2']['sigma']
+        e12 = lj.coeff['1','2']['epsilon']
+
+        #test scalar r
+        r_input = 11
+        u_actual = -0.00005144
```
mphoward, line 61: I like this test for single `r` input, but the values you are checking are really small. The LJ potential is basically zero for `r > 3*sigma`. You might want to test a smaller r so that the energy is numerically significant.

tests/potential/test_pair.py
```diff
@@ -0,0 +1,147 @@
...
+        #test setitem with default reset
+        lj.coeff['1','2'] = {'sigma':0.5}
+        self.assertAlmostEqual(lj.coeff['1','1']['epsilon'], 2.0)
+        self.assertAlmostEqual(lj.coeff['1','1']['sigma'], 1.5)
+        self.assertEqual(lj.coeff['1','2']['epsilon'], None)
+        self.assertAlmostEqual(lj.coeff['1','2']['sigma'], 0.5)
+
+    def test_energy(self):
+        """Test _energy method"""
+        lj = pair.LennardJones(types=('1','2'))
```
mphoward, line 49: I think it's fine to just do this test with one type, since we have good coverage of the `PairPotential` elsewhere. Technically, you don't even need to set the coeffs here because you can invoke the `_energy` method directly, but it's OK if you want to set them up.

tests/potential/test_pair.py
```diff
@@ -0,0 +1,147 @@
...
+        r_input = np.array([0,0.5,1,1.5])
+        u_actual = np.array([np.inf,0,-0.061523438,-0.0054794417])
+        u = lj._energy(r=r_input, epsilon=e12, sigma=s12)
+        np.testing.assert_allclose(u, u_actual)
+
+        #test negative sigma
+        with self.assertRaises(ValueError):
+            u = lj._energy(r=r_input, epsilon=e12, sigma=-1.0)
+
+    def test_force(self):
```
mphoward, line 75: Same comments as `_energy` apply here

tests/potential/test_pair.py
```diff
@@ -0,0 +1,147 @@
...
+        r_input = np.array([0,0.5,1,1.5])
+        f_actual = np.array([np.inf,48,-0.36328125,-0.02188766])
+        f = lj._force(r=r_input, epsilon=e12, sigma=s12)
+        np.testing.assert_allclose(f, f_actual)
+
+        #test negative sigma
+        with self.assertRaises(ValueError):
+            u = lj._force(r=r_input, epsilon=e12, sigma=-1.0)
+
+    def test_derivative(self):
```
mphoward, line 103: And here too.

relentless/potential/pair.py
```diff
@@ -6,10 +6,57 @@
 from .potential import PairPotential
 
 class LennardJones(PairPotential):
+    """Lennard-Jones 12-6 pair potential.
+
+    .. math::
+
+    u(r) = 4 \varepsilon\left[\left(\frac{\sigma}{r}\right)^12 - \left(\frac{\sigma}{r}\right)^12 \right]
```
mphoward, line 13: Indent one level here.

relentless/potential/pair.py
```diff
@@ -6,10 +6,57 @@
...
+
+    - :math:`\varepsilon`: interaction energy
+    - :math:`\sigma`: interaction length scale (e.g., particle diameter)
+
+    The optional coefficients per pair are:
+
+    - ``rmin``: minimum radius, energy and force are 0 for ``r < rmin``. Ignored if ``False`` (default).
+    - ``rmax``: maximum radius, energy and force are 0 for ``r > rmax`` Ignored if ``False`` (default).
+    - ``shift``: If ``True``, shift potential to zero at ``rmax`` (default is ``False``).
+
```
mphoward, line 25: I forgot to add this earlier, but could you add the following comment here:
```py
Setting ``rmax = sigma*2**(1./6.)`` and ``shift = True`` will give the purely repulsive Weeks-Chandler-Andersen potential, which can model nearly hard spheres.
```
(separate from the list above by a blank line)

relentless/potential/potential.py
```diff
@@ -621,7 +621,8 @@ def force(self, pair, potentials):
             Total force at each r value.
 
         """
-        f = np.zeros_like(self.r)
+        f = np.zeros(self.r.shape, dtype=np.float64)
+        f = f.astype(dtype=np.float64)
```
mphoward, line 625: You don't need this line anymore.

PR #25, opened 2020-07-17
Documentation, features, and tests for Interpolator.derivative

adisr04:
- Add Interpolator derivative method in core
- Add documentation
- Add unit test in test_core

### DIFF
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -44,6 +44,13 @@ class Interpolator:
         >>> f([-0.5,0.5])
         (-1.0, 1.0)
 
+    Evaluate the n-th derivative of the function::
+
+        >>> f.derivative(x=0.5, n=1)
+        2.0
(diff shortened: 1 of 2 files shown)


relentless/core.py
```diff
@@ -99,6 +106,50 @@ def __call__(self, x):
...
+        x = np.atleast_1d(x)
+        result = np.zeros(len(x))
+
+        # clamp lo
+        lo = x < self.domain[0]
+        result[lo] = self._spline(self.domain[0], nu=n)
+
+        # clamp hi
+        hi = x > self.domain[1]
+        result[hi] = self._spline(self.domain[1], nu=n)
```
mphoward, line 142: We define the extrapolation to be a constant outside the domain, so the derivative should be zero for `x < self.domain[0]` and `x > self.domain[1]`.

PR #28, opened 2020-07-20
Documentation, features, and unit tests for Spline

adisr04:
- Add Spline documentation
- Implement Spline methods/features
- Change 2-point Interpolator from Akima spline to linear interpolation
- Add all unit tests for Spline in test_pair

### DIFF
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -69,7 +69,12 @@ class Interpolator:
         if not np.all(x[1:] > x[:-1]):
             raise ValueError('x must be strictly increasing')
         self._domain = (x[0],x[-1])
-        self._spline = scipy.interpolate.Akima1DInterpolator(x=x, y=y)
+        if x.shape[0] > 2:
+            self._spline = scipy.interpolate.Akima1DInterpolator(x=x, y=y)
+        else:
+            self._spline = scipy.interpolate.interp1d(x=x, y=y)
+            self._y = y
+            self._x = x
 
     def __call__(self, x):
         """Evaluate the interpolating function.
@@ -143,7 +148,10 @@ class Interpolator:
 
         # evaluate in between
         flags = np.logical_and(~lo,~hi)
-        result[flags] = self._spline(x[flags], nu=n)
+        if isinstance(self._spline, scipy.interpolate.Akima1DInterpolator):
+            result[flags] = self._spline(x[flags], nu=n)
+        else:
+            result[flags] = (self._y[1] - self._y[0])/(self._x[1] - self._x[0])
 
         if scalar_x:
             result = result.item()
--- a/relentless/core.py
+++ b/relentless/core.py
@@ -72,9 +72,7 @@ class Interpolator:
         if x.shape[0] > 2:
             self._spline = scipy.interpolate.Akima1DInterpolator(x=x, y=y)
         else:
-            self._spline = scipy.interpolate.interp1d(x=x, y=y)
-            self._y = y
-            self._x = x
+            self._spline = scipy.interpolate.InterpolatedUnivariateSpline(x=x, y=y, k=1)
 
     def __call__(self, x):
         """Evaluate the interpolating function.
@@ -148,10 +146,7 @@ class Interpolator:
 
         # evaluate in between
         flags = np.logical_and(~lo,~hi)
-        if isinstance(self._spline, scipy.interpolate.Akima1DInterpolator):
-            result[flags] = self._spline(x[flags], nu=n)
-        else:
-            result[flags] = (self._y[1] - self._y[0])/(self._x[1] - self._x[0])
+        result[flags] = self._spline.derivative(n)(x[flags])
 
         if scalar_x:
             result = result.item()
--- a/relentless/potential/pair.py
+++ b/relentless/potential/pair.py
@@ -1,4 +1,5 @@
 __all__ = ['LennardJones','Spline','Yukawa']
+import copy
 
 import numpy as np
 
@@ -157,6 +158,26 @@ class LennardJones(PairPotential):
         return d
 
 class Spline(PairPotential):
+    """Spline potential using Akima interpolation.
+
+    Parameters
+    ----------
+    types : array_like
+        List of types (A type must be a `str`).
+    num_knots : int
+        The number of knots with which to interpolate.
+    mode : `str`
+        Describes the type of design parameter used in the optimization. Setting the mode to 'value' uses the knot amplitudes,
+        while 'diff' uses the differences between neighboring knot amplitudes. (Defaults to 'diff').
+
+    Raises
+    ------
+    ValueError
+        If the number of knots is not an integer >= 2.
+    ValueError
+        If the mode is not set to either 'value' or 'diff'.
+
+    """
     valid_modes = ('value','diff')
 
     def __init__(self, types, num_knots, mode='diff'):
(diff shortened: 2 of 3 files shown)


relentless/potential/pair.py
```diff
@@ -156,6 +157,213 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        rs = np.asarray(r, dtype=np.float64)
+        ks = np.asarray(u, dtype=np.float64)
+        if self.mode == 'diff':
+            # difference is next knot minus my knot, with last knot fixed at its current value
+            ks[:-1] -= ks[1:]
+
+        # make all knots variables, but hold all r and the last knot const
+        for i in range(self.num_knots):
+            self.coeff[pair]['r-{}'.format(i)] = core.Variable(rs[i],const=True)
+            self.coeff[pair]['knot-{}'.format(i)] = core.Variable(ks[i],const=(i==self.num_knots-1))
```
mphoward, line 233: Let's expand this so that if the knot is already a variable, we just set the value:
```py
ri = 'r-{}'.format(i)
if self.coeff[pair][ri] is None:
    self.coeff[pair][ri] = core.Variable(rs[i],const=True)
else:
    self.coeff[pair][ri].value = rs[i]
```

As a general point, it would be also be good to add a method that can spit back the knot variables
```py
def _knot_params(self, i):
    assert isinstance(i,int), "Knots are keyed by integers"
    return 'r-{}'.format(i), 'knot-{}'.format(i)
```
For use in the constructor and other methods. That way we can always rename the variables later without having to change much code.

relentless/potential/pair.py
```diff
@@ -156,6 +157,213 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+
+        Returns
+        -------
+        array_like
+            Returns the value or values of the derivative evaluated at r.
+
+        """
+        r,d,s = self._zeros(r)
+        h = 0.001
+        params = self.coeff[pair]
```
mphoward, line 301: You already have all of the knot params in `**params`, so you shouldn't be accessing the coefficients directly.

relentless/potential/pair.py
```diff
@@ -156,6 +157,213 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+
+        #perturb params[param]
+        params_high = copy.deepcopy(params.todict())
+        params_low = copy.deepcopy(params.todict())
+        params_high[param].value += h
+        params_low[param].value -= h
```
mphoward, line 307: Similarly, no need to deep copy here. You can just modify the param dict like:
```py
kcur = params[param]
params[param] = kcur + h
# evaluate
params[param] = kcur - h
# evaluate
# do derivative
```
Or, you can make a dict shallow copy like this:
```py
params_high = dict(params)
```

relentless/potential/pair.py
```diff
@@ -156,6 +157,213 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+            The parameter with respect to which to take the derivative (the knot coefficient).
+        r : array_like
+            The value or values of r at which to evaluate the derivative.
+
+        Returns
+        -------
+        array_like
+            Returns the value or values of the derivative evaluated at r.
+
+        """
```
mphoward, line 298: Let's error check that the derivative is only taken with respect to the knot height, as we may have issues with r derivatives currently.
```py
if 'knot' not in param:
    raise ValueError('Derivative can only be taken for knot values')
```

relentless/potential/pair.py
```diff
@@ -156,6 +157,213 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        params_high = copy.deepcopy(params.todict())
+        params_low = copy.deepcopy(params.todict())
+        params_high[param].value += h
+        params_low[param].value -= h
+
+        d = self._interpolate(params_high)(r) - self._interpolate(params_low)(r)
+        d /= (2*h)
+        if s:
+            d = d.item()
+        return d
```
mphoward, line 313: Anticipating that we might have other types of variables in a coefficient dictionary, do we want to force the user to leave the `Variable`s that we have created alone? We could override the parent `derivative` method to make sure everything stays that way:
```py
def derivative(self, pair, param, r):
    for i in range(self.num_knots):
        ri,ki = self._knot_params(i)
        if not isinstance(self.coeff[pair][ri], core.Variable):
            # raise error
        if not isinstance(self.coeff[pair][ki], core.Variable):
            # raise error
    return super().derivative(pair, param, r)
```
and then admonish them not to override them in the docs.

relentless/potential/pair.py
```diff
@@ -156,6 +157,213 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        params : array_like
+            The array of knot values.
+
+        Returns
+        -------
+        :py:class:`Interpolator`
+            The interpolated spline potential.
+
+        """
+        r = np.array([params['r-{}'.format(i)].value for i in range(self.num_knots)])
```
mphoward, line 329: Replace these using `_knot_params` to get the names.

tests/potential/test_pair.py
```diff
@@ -97,6 +97,150 @@ def test_derivative(self):
...
+
+    def test_force(self):
+        """Test _force method"""
+        r_arr = [1,2,3]
+        u_arr = [9,4,1]
+
+        #test diff mode
+        s = pair.Spline(types=('1',), num_knots=3)
+        s.from_array(pair=('1','1'), r=r_arr, u=u_arr)
+        f_actual = np.array([5,3,0])
```
mphoward, line 214: This might be a silly question, but how did you determine what these forces should be? For 3 points, does the interpolator just do linear interpolation?

tests/potential/test_pair.py
```diff
@@ -97,6 +97,150 @@ def test_derivative(self):
...
+        s = pair.Spline(types=('1',), num_knots=3)
+        s.from_array(pair=('1','1'), r=r_arr, u=u_arr)
+        d_actual = np.array([1.125,0.625,0])
+        d = s._derivative(pair=('1','1'), param='knot-1', r=[1.5,2.5,3.5])
+        np.testing.assert_allclose(d, d_actual)
+
+        #test value mode
+        s = pair.Spline(types=('1',), num_knots=3, mode='value')
+        s.from_array(pair=('1','1'), r=r_arr, u=u_arr)
+        d_actual = np.array([0.75,0.75,0])
```
mphoward, line 240: As I said, I think these should be different derivatives in the different modes. If we could work out analytically how they relate to each other, that would be a strong test.

tests/potential/test_pair.py
```diff
@@ -97,6 +97,150 @@ def test_derivative(self):
...
+        d = s._derivative(pair=('1','1'), param='knot-1', r=[1.5,2.5,3.5])
+        np.testing.assert_allclose(d, d_actual)
+
```
mphoward, line 243: Can we add a special test case for a spline with only 2 knots? We do want to be able to support this, but it is sort of a special case. Maybe just construct the spline with two knots and evaluate the energy and force. It should be linear interpolation (since you cannot fit any other type of polynomial to this), so it should be easy to check.

relentless/core.py
```diff
@@ -69,7 +69,12 @@ def __init__(self, x, y):
         if not np.all(x[1:] > x[:-1]):
             raise ValueError('x must be strictly increasing')
         self._domain = (x[0],x[-1])
-        self._spline = scipy.interpolate.Akima1DInterpolator(x=x, y=y)
+        if x.shape[0] > 2:
+            self._spline = scipy.interpolate.Akima1DInterpolator(x=x, y=y)
+        else:
+            self._spline = scipy.interpolate.interp1d(x=x, y=y)
+            self._y = y
+            self._x = x
```
mphoward, line 77: This works, but if you want a linear interpolation that supports the derivative method, you could use
```py
self._spline = scipy.interpolate.InterpolatedUnivariateSpline(x=x, y=y, k=1)
```
I think this class is available in most versions of scipy. If not, you can fall back on `UnivariateSpline` and additionally set `s=0`.

relentless/core.py
```diff
@@ -143,7 +148,10 @@ def derivative(self, x, n):
 
         # evaluate in between
         flags = np.logical_and(~lo,~hi)
-        result[flags] = self._spline(x[flags], nu=n)
+        if isinstance(self._spline, scipy.interpolate.Akima1DInterpolator):
```
mphoward, line 151: With the above change, you can just invoke the derivative method directly.

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
             d = d.item()
         return d
 
+class Spline(PairPotential):
+    """Spline potential using Akima interpolation.
+
+    Parameters
+    ----------
+    types : array_like
+        List of types (A type must be a `str`).
```
mphoward, line 164:
```suggestion
        List of types (A type must be a str).
```

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+    num_knots : int
+        The number of knots with which to interpolate.
+    mode : `str`
```
mphoward, line 167:
```suggestion
    mode : str
```

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+                self.coeff[pair][ri] = core.Variable(rs[i],const=True)
+            else:
+                self.coeff[pair][ri].value = rs[i]
+            if self.coeff[pair][ki] is None:
+                self.coeff[pair][ki] = core.Variable(ks[i],const=(i==self.num_knots-1))
+            else:
+                self.coeff[pair][ki].value = ks[i]
+
+    def _knot_params(self, i):
+        """Returns the knot variables.
```
mphoward, line 245:
```suggestion
        """Get the parameter names for a given knot.
```

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        ------
+        ValueError
+            If the knot key is not an integer.
+
+        """
+        if not isinstance(i, int):
+            raise ValueError('Knots are keyed by integers')
+        return 'r-{}'.format(i), 'knot-{}'.format(i)
+
+    def _energy(self, r, params, **pars):
```
mphoward, line 269: You should just accept all the keywords together. The `energy` method will just forward the entire dictionary using **params.
```suggestion
    def _energy(self, r, **params):
```

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+            The parameters (r and knot values) for the potential function.
+
+        Returns
+        -------
+        array_like
+            Returns the value or values of the energy evaluated at r.
+
+        """
+        r,u,s = self._zeros(r)
+        u = self._interpolate(params)(r)
```
mphoward, line 286: `_interpolate` is using `params` as a dict, which is why it doesn't need to be packed again. If you wanted to make the signatures the same, you could change this to `**params` and then modify it below to accept `**params` as well.

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        if s:
+            u = u.item()
+        return u
+
+    def _force(self, r, params, **pars):
```
mphoward, line 291: Same change here.

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+            Returns the value or values of the force evaluated at r.
+
+        """
+        r,f,s = self._zeros(r)
+        f = -self._interpolate(params).derivative(r, 1)
+        if s:
+            f = f.item()
+        return f
+
+    def derivative(self, pair, param, r):
```
mphoward, line 313: You can probably just omit this docstring since it doesn't add much new beyond the parent documentation. Maybe add a short comment (with `#`) that says you are extending the method to check that they are variables.

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        """
+        for i in range(self.num_knots):
+            ri,ki = self._knot_params(i)
+            if not isinstance(self.coeff[pair][ri], core.Variable):
+                raise TypeError('All r values must be Variables')
+            if not isinstance(self.coeff[pair][ki], core.Variable):
+                raise TypeError('All knot values must be Variables')
+        return super().derivative(pair, param, r)
+
+    def _derivative(self, param, r, params, **pars):
```
mphoward, line 351: Same suggested change to signature as for `_energy` and `_force`.

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+            The array of knot values.
+
+        Returns
+        -------
+        :py:class:`Interpolator`
+            The interpolated spline potential.
+
+        """
+        r = np.empty(self.num_knots)
+        u = np.empty(self.num_knots)
```
mphoward, line 408: I would recommend using `zeros`, just to be safe. `empty` gives you whatever bits are in the memory when it's allocated, which are undefined values.

tests/potential/test_pair.py
```diff
@@ -97,6 +97,173 @@ def test_derivative(self):
...
+    def test_from_array(self):
+        """Test from_array method"""
+        r_arr = [1,2,3]
+        u_arr = [9,4,1]
+
+        #test diff mode
+        s = pair.Spline(types=('1',), num_knots=3)
+        s.from_array(pair=('1','1'), r=r_arr, u=u_arr)
+
+        self.assertEqual(s.coeff['1','1']['r-0'].value, 1)
```
mphoward, line 145: I would maybe access these with the generator or with `_knot_params` so that you don't have to hardcode the parameter names.

relentless/potential/pair.py
```diff
@@ -156,6 +155,298 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        for i in range(self.num_knots):
+            ri,ki = self._knot_params(i)
+            r[i] = params[ri].value
+            u[i] = params[ki].value
```
mphoward, line 412: These should just be `params[ri]` and `params[ki]`, since `params` is intended to be passed as a dict of evaluated knots.

tests/potential/test_pair.py
```diff
@@ -97,6 +97,173 @@ def test_derivative(self):
...
+    def test_energy(self):
+        """Test _energy method"""
+        r_arr = [1,2,3]
+        u_arr = [9,4,1]
+
+        #test diff mode
+        s = pair.Spline(types=('1',), num_knots=3)
+        s.from_array(pair=('1','1'), r=r_arr, u=u_arr)
+        u_actual = np.array([6.25,2.25,1])
+        u = s._energy(r=[1.5,2.5,3.5], params=s.coeff['1','1'])
```
mphoward, line 196: I would still suggest switching this to `energy` rather than `_energy`. Otherwise, you will need to do `s.evaluate(('1','1'))` to get the params.

relentless/potential/pair.py
```diff
@@ -156,6 +155,270 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+            The keyed knot value.
+
+        Raises
+        ------
+        ValueError
+            If the knot key is not an integer.
+
+        """
+        if not isinstance(i, int):
+            raise ValueError('Knots are keyed by integers')
```
mphoward, line 266: `ValueError` -> `TypeError`

relentless/potential/pair.py
```diff
@@ -156,6 +155,270 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        Returns
+        -------
+        str
+            The keyed r value.
+        str
+            The keyed knot value.
+
+        Raises
+        ------
+        ValueError
```
mphoward, line 261: `ValueError` -> `TypeError`

relentless/potential/pair.py
```diff
@@ -156,6 +155,270 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        return f
+
+    def derivative(self, pair, param, r):
+        #Extending PairPotential method to check if r and knot values are Variables.
+        for i in range(self.num_knots):
+            ri,ki = self._knot_params(i)
+            if not isinstance(self.coeff[pair][ri], core.Variable):
+                raise TypeError('All r values must be Variables')
+            if not isinstance(self.coeff[pair][ki], core.Variable):
+                raise TypeError('All knot values must be Variables')
```
mphoward, line 320: Maybe simplify with the knot generator?
```suggestion
        for ri,ki in self.knots(pair):
            if not isinstance(ri, core.Variable):
                raise TypeError('All r values must be Variables')
            if not isinstance(ki, core.Variable):
                raise TypeError('All knot values must be Variables')
```

relentless/potential/pair.py
```diff
@@ -156,6 +155,270 @@ def _derivative(self, param, r, epsilon, sigma, **params):
...
+        ----------
+        pair : tuple
+            The type pair (i,j) for which to retrieve the knot points.
+
+        Yields
+        ------
+        `str`
+            The next r key in the coefficient array of r values.
+        `str`
+            The next knot key in the coefficient array of k values.
```
mphoward, line 413: `str` -> `Variable` (these return the variables themselves, not the keys.)

PR #31, opened 2020-07-30
Documentation, features, and unit tests for new Variable types

adisr04:
- Reorganize core module and unit tests
- Add variable types and documentation
- Add DependentVariable features for dependency graph creation and chain rule derivative calculation
- Add variable unit tests
- Make other unit tests compatible


tests/core/test_collections.py
```diff
@@ -0,0 +1,275 @@
...
+        self.assertEqual(m['B','B'], {'energy':1.0, 'mass':2.0})
+
+        #test re-iteration for setting values
+        for pair in m:
+            m[pair]['energy'] = 3.0
+        self.assertEqual(m['A','B'], {'energy':3.0, 'mass':2.5})
+        self.assertEqual(m['A','A'], {'energy':3.0, 'mass':2.0})
+        self.assertEqual(m['B','B'], {'energy':3.0, 'mass':2.0})
+
+class test_FixedKeyDict(unittest.TestCase):
```
mphoward, line 129: I reordered these alphabetically in collections, so could you move this test before the PairMatrix?

relentless/core/variable.py
```diff
@@ -0,0 +1,311 @@
...
+    low : float or None
+        Lower bound for the variable (`None` means no lower bound).
+    high : float or None
+        Upper bound for the variable (`None` means no upper bound).
+
+    Examples
+    --------
+    A variable with a lower bound::
+
+        >>> v = Variable(value=1.0, low=0.0)
```
mphoward, line 47: Can you double check for any instances of `Variable` in this class's docs?

```suggestion
        >>> v = DesignVariable(value=1.0, low=0.0)
```

relentless/core/variable.py
```diff
@@ -0,0 +1,311 @@
...
+        Parameters
+        ----------
+        value : float
+            Value to clamp within bounds.
+
+        Returns
+        -------
+        v : float
+            The clamped value.
+        b : :py:class:`Variable.State`
```
mphoward, line 102:
```suggestion
        b : :py:class:`DesignVariable.State`
```

relentless/core/variable.py
```diff
@@ -0,0 +1,311 @@
...
+
+    @property
+    def value(self):
+        """float: The value of the variable"""
+        return self._value
+
+    @value.setter
+    def value(self, value):
+        if not isinstance(value,(float,int)):
+            raise ValueError('Variable must be a float or int')
```
mphoward, line 126:
```suggestion
            raise ValueError('DesignVariable must be a float or int')
```

relentless/core/variable.py
```diff
@@ -0,0 +1,311 @@
...
+            self._high = high
+
+        try:
+            self._value, self._state = self.clamp(self._value)
+        except AttributeError:
+            pass
+
+    @property
+    def state(self):
+        """:py:class:`Variable.State`: The state of the variable"""
```
mphoward, line 179:
```suggestion
        """:py:class:`DesignVariable.State`: The state of the variable"""
```

tests/core/test_variable.py
```diff
@@ -0,0 +1,349 @@
...
+        #test variable dependent on DependentVariable
+        u = relentless.SameAs(w)
+        self.assertEqual(u.value, 1.0)
+        self.assertCountEqual(u.depends, (w,))
+
+        #test invalid variable dependence
+        with self.assertRaises(TypeError):
+            u = relentless.SameAs(1.0)
+
+    def test_accessors(self):
```
mphoward, line 168: Let's rename all of these `test_value` for each variable class, since that is closer to what you're trying.

relentless/core/variable.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    def __init__(self, *vardicts, **kwvars):
+        attrs = {}
+        for d in vardicts:
+            attrs.update(d)
+        attrs.update(**kwvars)
+
+        for k,v in attrs.items():
+            super().__setattr__(k,self._assert_variable(v))
+        self._depends = tuple(kwvars.keys())
```
mphoward, line 241: Was just looking for some code and noticed something. Does this need to be `attr.keys()`? What happens if no `kwvars` are given? It would be good to add some tests with a mock dependent variable (like you did for the pair potential) that is constructed using different combinations of these args.

mphoward:
A few notes for us:
1. Is it legal for the user to specify the same object twice as a dependency? For example, `ArithmeticMean(a,a)` is equivalent to `SameAs(a)`, and the derivative w.r.t. a should be 1.0 for both. Currently, `ArithmeticMean` does not handle this correctly, as it will return 1.0. I think we need to allow this, and fix the derivatives as a special case.
2. Should the `depends` attribute be the *unique* set of dependencies of an object? I think this should also be the case, and so we need to use a set.
3. The `derivative` method should be expanded to process the chain rule. This can be handled using a digraph.

relentless/core/variable.py
```diff
@@ -290,9 +290,15 @@ def value(self):
 
     def derivative(self, var):
         if var is self.a:
```
mphoward, line 292: I think this logic isn't quite right, since we need to make sure the objects are *exactly* the same, and not just having the same value. This should do the trick:
```py
if var is self.a and var is self.b:
    return 1.0
elif var is self.a or var is self.b:
    return 0.5
else:
    return 0.0
```
Could you also add to the unit tests a case where you have two variables that have the same value (but are different objects) in addition to the case where they are exactly the same object? Both cases will have the same value but different derivatives.

relentless/core/variable.py
```diff
@@ -304,8 +310,14 @@ def value(self):
...
-            return 0.5*np.sqrt(self.b.value/self.a.value)
```
mphoward: You'll need to do something similar as for the arithmetic mean here.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+    """Abstract base class for any variable.
+
+    At minimum, a variable must provide the interface to get its value as a property.
+
+    """
+    @property
+    @abc.abstractmethod
+    def value(self):
+        pass
+
```
mphoward, line 22: Anticipating potential future needs to add other types of "independent" variables, or maybe so that we can try incorporate scalar constants into a dependent variable, can we add a light `IndependentVariable` class? It should derive from `Variable` and its constructor should take only one parameter `value`, which is stored in `self._value`. Its `value` property should just return `self._value`, and you can add a `value` setter as well that sets `self._value`.

Then, `DesignVariable` can derive from `IndependentVariable` and override these methods as necessary.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+        if len(attrs) == 0:
+            raise AttributeError('No attributes specified for DependentVariable.')
+
+        for k,v in attrs.items():
+            super().__setattr__(k,self._assert_variable(v))
+        self._params = tuple(attrs.keys())
+
+    def __setattr__(self, name, value):
+        #Sets the value of the variable on which the specified DependentVariable depends.
+        if name != '_params' and name in self._params:
```
mphoward, line 250:
```suggestion
        if name != '_params' and name in self.params:
```

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+            value = self._assert_variable(value)
+        super().__setattr__(name,value)
+
+    @property
+    def depends(self):
+        """Generator of parameter,object pairs."""
+        for p in self._params:
```
mphoward, line 257:
```suggestion
        for p in self.params:
```

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def __setattr__(self, name, value):
+        #Sets the value of the variable on which the specified DependentVariable depends.
+        if name != '_params' and name in self._params:
+            value = self._assert_variable(value)
+        super().__setattr__(name,value)
+
+    @property
+    def depends(self):
+        """Generator of parameter,object pairs."""
```
mphoward, line 256: Can you expand this documentation a bit to explain what is returned? The first element is the string name of the parameter, and the second element is the object reference itself.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+        for p in self._params:
+            yield p,getattr(self,p)
+
+    @property
+    def params(self):
```
mphoward, line 261: Document that this returns a tuple of the named parameters that are considered dependencies.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+        return self._params
+
+    def dependency_graph(self):
```
mphoward, line 264: Document that this constructs a networkx graph of this objects on which this variable depends.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+
+        # set edges between all nodes
+        for a in g:
+            if isinstance(a,DependentVariable):
+                for p,b in a.depends:
+                    # add new edge for p if not already present, otherwise append to params list
+                    if (a,b) not in g.edges:
+                        g.add_edge(a,b,params=[p])
+                    else:
+                        g.edges[a,b]['params'].append(p)
```
mphoward, line 285: Now that unit tests are in place, I think that these two loops can probably be fused. I left them separate because adding edges to the graph automatically adds nodes, so this was making it messy to keep track of visited nodes. 

I would suggest adding a `visited` set, and using this instead:
```suggestion
        # discover all variables (nodes) by quasi-depth first search
        g = nx.DiGraph()
        stack = [self]
        visited = set()
        while stack:
            a = stack.pop()
            if a not in visited:
                g.add_node(a)
                visited.add(a)
                if isinstance(a,DependentVariable):
                    for p,b in a.depends:
                        stack.append(b)
                        # add new edge for p if not already present, otherwise append to params list
                        if (a,b) not in g.edges:
                            g.add_edge(a,b,params=[p])
                        else:
                            g.edges[a,b]['params'].append(p)
```

You can probably also remove the explicit `g.add_node(a)`, but this might be less readable.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+        return v
+
+    def _assert_acyclic(self, g):
+        # confirm dependency graph is free of cycles
+        if not nx.is_directed_acyclic_graph(g):
+            raise RuntimeError('DependentVariable has circular dependencies.')
+        return g
+
+class UnaryOperator(DependentVariable):
+    """Abstract base class for dependent variable based on one parameter value."""
```
mphoward, line 348:
```suggestion
    """Abstract base class for a value that depends on one :py:class:`Variable`."""
```

Please all document the argument `a`.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+    def __init__(self, a):
+        super().__init__(a=a)
+
+class SameAs(UnaryOperator):
+    """Unary operator for copying a variable."""
```
mphoward, line 353: Please document the constructor argument (`a`), and write a sentence or two (maybe with example) documenting that this is just a copy of the variable.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+    @property
+    def value(self):
+        return self.a.value
+
+    def _derivative(self, param):
```
mphoward, line 358: Document the argument parameter (it can only be a string `a`), and the `ValueError`.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+        if param == 'a':
+            return 1.0
+        else:
+            raise ValueError('Unknown parameter')
+
+class BinaryOperator(DependentVariable):
+    """Abstract base class for dependent variable based on two parameter values."""
```
mphoward, line 365:
```suggestion
    """Abstract base class for a value that depends on two :py:class:`Variable`s."""
```

Please also document the two arguments `a` and `b`.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+    def __init__(self, a, b):
+        super().__init__(a=a, b=b)
+
+class ArithmeticMean(BinaryOperator):
+    """Binary operator based on arithmetic mean."""
```
mphoward, line 370: Please document the constructor arguments (a and b), and write a sentence or two (maybe with example) documenting the formula. It would be good to note that this can also be used to implement mixing rules.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+    @property
+    def value(self):
+        return 0.5*(self.a.value+self.b.value)
+
+    def _derivative(self, param):
```
mphoward, line 375: Document the argument parameter (it can only be a string "a" or "b"), and the ValueError.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+        if param == 'a':
+            return 0.5
+        elif param == 'b':
+            return 0.5
+        else:
+            raise ValueError('Unknown parameter')
+
+class GeometricMean(BinaryOperator):
+    """Binary operator based on geometric mean."""
```
mphoward, line 384: Please document the constructor arguments (a and b), and write a sentence or two (maybe with example) documenting the formula. It would be good to note that this can also be used to implement mixing rules.

relentless/core/variable.py
```diff
@@ -0,0 +1,395 @@
...
+    @property
+    def value(self):
+        return np.sqrt(self.a.value*self.b.value)
+
+    def _derivative(self, param):
```
mphoward, line 389: Document the argument parameter (it can only be a string "a" or "b"), and the ValueError.

relentless/potential/potential.py
```diff
@@ -164,7 +164,7 @@ def evaluate(self, pair):
                 raise ValueError('Parameter {} is not set for ({},{}).'.format(p,pair[0],pair[1]))
 
             # evaluate the variable
-            if isinstance(v, Variable):
+            if isinstance(v, DesignVariable):
```
mphoward, line 167: This should revert to the base class `Variable`, since we now support evaluating any type.
```suggestion
            if isinstance(v, Variable):
```

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+class DepVar(relentless.DependentVariable):
+    """Mock dependent variable to test relentless.DependentVariable"""
+    def __init__(self, *a, **b):
+        super().__init__(*a, **b)
+
+    @property
+    def value(self):
+        return np.sum([v.value for p,v in self.depends])
+
+    def _derivative(self, param):
```
mphoward, line 157: Can you add a comment here that these derivatives aren't the actual derivs and that they are just used for testing? It might confused someone later.

Alternatively, do any of the tests ever trigger the first condition? You could also just modify these so that value and _derivative are compatible with each other.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        u = relentless.DesignVariable(value=2.0)
+        v = relentless.DesignVariable(value=3.0)
+
+        #test creation with only vardicts
+        w = DepVar({'t':t, 'u':u, 'v':v})
+        self.assertCountEqual(w.params, ('t','u','v'))
+        self.assertDictEqual({p:v for p,v in w.depends}, {'t':t,'u':u,'v':v})
+        self.assertAlmostEqual(w.value, 6.0)
+
+        #test creation with only kwvars
```
mphoward, line 181: I think this is missing the constructor for `w`?
```suggestion
        #test creation with only kwvars
        w = DepVar(t=t, u=u, v=v)
```

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        self.assertCountEqual(w.params, ('t','u','v'))
+        self.assertDictEqual({p:v for p,v in w.depends}, {'t':t,'u':u,'v':v})
+        self.assertAlmostEqual(w.value, 6.0)
+
+        #test creation with vardicts and kwvars
+        w = DepVar({'t':t, 'u':u}, v=v)
+        self.assertCountEqual(w.params, ('t','u','v'))
+        self.assertDictEqual({p:v for p,v in w.depends}, {'t':t,'u':u,'v':v})
+        self.assertAlmostEqual(w.value, 6.0)
+
```
mphoward, line 191: Can we add a couple more tests here:
1. The same attribute is repeated twice (like, it's in vardicts and kwvars), making sure that the last instance takes precedence.
2. There are two vardicts specified (it should iterate over them all).
3. There are no parameters specified at all, which should raise an error.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        u = relentless.DesignVariable(value=2.0)
+        v = relentless.DesignVariable(value=3.0)
+
+        #test 1 level of dependency
+        w = DepVar(t=t, u=u, v=v)
+        g = w.dependency_graph()
+        self.assertCountEqual(g.nodes, [t,u,v,w])
+        self.assertCountEqual(g.edges, [(w,t),(w,u),(w,v)])
+        self.assertCountEqual([g.edges[a,b]['params'] for (a,b) in g.edges],
+                              [['t'],['u'],['v']])
```
mphoward, line 204: Let's test these one by one so that we make sure the right edges has the right parameter. E.g.,
```suggestion
        self.assertEqual(g.edges[w,t]['params'],['t'])
```

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+
+        #test 2 levels of dependency
+        x = DepVar(t=t, v=v)
+        y = DepVar(w=w, x=x)
+        g = y.dependency_graph()
+        self.assertCountEqual(g.nodes, [t,u,v,w,x,y])
+        self.assertCountEqual(g.edges, [(w,t),(w,u),(w,v),(x,t),(x,v),(y,w),(y,x)])
+        self.assertCountEqual([g.edges[a,b]['params'] for (a,b) in g.edges],
```
mphoward, line 212: Same here, check the parameters manually.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+                              [['t'],['u'],['v'],['t'],['v'],['w'],['x']])
+
+        #test more complex dependency
+        z = DepVar(t=t, w=w, y=y)
+        g = z.dependency_graph()
+        self.assertCountEqual(g.nodes, [t,u,v,w,x,y,z])
+        self.assertCountEqual(g.edges, [(w,t),(w,u),(w,v),(x,t),(x,v),(y,w),(y,x),
+                                        (z,t),(z,w),(z,y)])
+        self.assertCountEqual([g.edges[a,b]['params'] for (a,b) in g.edges],
```
mphoward, line 221: At this point, you are probably convinced that the params are getting set, so you can just remove this check.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+                              [['t'],['u'],['v'],['t'],['v'],['w'],['x'],['t'],['w'],['y']])
+
+        #test 'multiple' dependency on same object
+        q = DepVar(t=t, u=t, v=t)
+        g = q.dependency_graph()
+        self.assertCountEqual(g.nodes, [t,q])
+        self.assertCountEqual(g.edges, [(q,t)])
+        params = [g.edges[a,b]['params'] for (a,b) in g.edges][0]
+        self.assertCountEqual(params, ['t','u','v'])
```
mphoward, line 230:
```suggestion
        self.assertCountEqual(g.edges[q,t]['params'], ['t','u','v'])
```

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+
+        q = DepVar(t=t, u=t, v=v, w=v)
+        g = q.dependency_graph()
+        self.assertCountEqual(g.nodes, [t,v,q])
+        self.assertCountEqual(g.edges, [(q,t),(q,v)])
+        params0 = sorted([g.edges[a,b]['params'] for (a,b) in g.edges][0])
+        params1 = sorted([g.edges[a,b]['params'] for (a,b) in g.edges][1])
+        self.assertCountEqual([params0,params1], [['t','u'],['v','w']])
```
mphoward, line 238: Similar change as above.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+
+        #test circular dependencies
+        a = DepVar(t=t)
+        a.t = DepVar(t=a)
```
mphoward, line 242: I got a little confused about this test at first, so to make it more explicit why this is circular, I would suggest:
```suggestion
        a = DepVar(t=t)
        b = DepVar(t=a)
        a.t = b
```
Then use `b` below instead of `a.t`.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        g = a.dependency_graph()
+        self.assertCountEqual(g.nodes, [a.t,a])
+        self.assertCountEqual(g.edges, [(a,a.t),(a.t,a)])
+        self.assertCountEqual([g.edges[a,b]['params'] for (a,b) in g.edges],
+                              [['t'],['t']])
```
mphoward, line 247: Split these out to be explicit. Also, that way you won't use `a` and `b` as iterators, which could be confusing since they refer to variables above.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+
+        a = DepVar(t=t)
+        b = DepVar(t=a)
+        c = DepVar(t=b)
+        a.t = DepVar(t=c)
+        g = a.dependency_graph()
+        self.assertCountEqual(g.nodes, [a.t,a,b,c])
+        self.assertCountEqual(g.edges, [(b,a),(c,b),(a.t,c),(a,a.t)])
```
mphoward, line 255: You can probably make this just a 3-ring cycle a, b, c (not a 4 ring cycle) then set `a.t = c`. The checks would then be:
```suggestion
        self.assertCountEqual(g.nodes, [a,b,c])
        self.assertCountEqual(g.edges, [(b,a),(c,b),(a,c)])
```
You can probably remove the `params` check, since it isn't testing anything new.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        self.assertCountEqual([g.edges[a,b]['params'] for (a,b) in g.edges],
+                              [['t'],['t'],['t'],['t']])
+
```
mphoward, line 258: Can we add a test with a cycle in a larger graph that is not trivially a cycle? Like:
```py
a = DepVar(t=t)
b = DepVar(t=a)
a.t = b
c = DepVar(t=b)
```
That should make one cycle and one not cycle: a <-> b <- c

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        u = relentless.DesignVariable(value=2.0)
+        v = relentless.DesignVariable(value=3.0)
+
+        #test derivative with 1 level of dependency
+        w = DepVar(t=t, u=u, v=v)
+        dw = w.derivative(t)
+        self.assertAlmostEqual(dw, -0.5)
+        dw = w.derivative(v)
+        self.assertAlmostEqual(dw, -0.5)
+
```
mphoward, line 271: Can we add a check for the special case `w.derivative(w)`, which should return 1.0?

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        #change value of v
+        v.value = 0.4
+        self.assertEqual(v.value, 0.5)
+        self.assertEqual(w.value, 0.5)
+        self.assertEqual(x.value, 0.5)
+
+    def test_derivative(self):
+        """Test _derivative method."""
+        v = relentless.DesignVariable(value=1.0)
```
mphoward, line 364: Let's use a different value than the derivative, just in case there's a typo.
```suggestion
        v = relentless.DesignVariable(value=2.0)
```
Can you also add one line where you change the value and check the deriv again?

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        self.assertAlmostEqual(y.value, 1.625)
+        self.assertCountEqual(y.params, ('a','b'))
+        self.assertDictEqual({p:v for p,v in y.depends}, {'a':w,'b':x})
+
+        #test same-valued dependencies
+        z = relentless.DesignVariable(value=1.0)
+        w = relentless.ArithmeticMean(u, z)
+        self.assertAlmostEqual(w.value, 1.0)
+        self.assertCountEqual(w.params, ('a','b'))
+        self.assertDictEqual({p:v for p,v in w.depends}, {'a':u,'b':z})
```
mphoward, line 405: You can probably remove this test, it doesn't really check much different than above.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        w = relentless.ArithmeticMean(u, v)
+
+        #test w.r.t. a
+        dw = w._derivative('a')
+        self.assertEqual(dw, 0.5)
+
+        #test w.r.t. b
+        dw = w._derivative('b')
+        self.assertEqual(dw, 0.5)
+
```
mphoward, line 461: Add a quick test where you change the value of u and/or v, and recheck the deriv.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+            w._derivative('c')
+
+class test_GeometricMean(unittest.TestCase):
+    """Unit tests for relentless.GeometricMean"""
+
+    def test_init(self):
+        """Test creation with data."""
+        #test variable dependent on 2 DesignVariables
+        u = relentless.DesignVariable(value=1.0)
+        v = relentless.DesignVariable(value=2.0)
```
mphoward, line 473: Can we make these numbers that are easier to check? I'd suggest setting `v = 16.0` so that one sqrt gives 4.0, and another sqrt gives 2.0.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+        self.assertAlmostEqual(y.value, 1.5422108)
+        self.assertCountEqual(y.params, ('a','b'))
+        self.assertDictEqual({p:v for p,v in y.depends}, {'a':w,'b':x})
+
+        #test same-valued dependencies
+        z = relentless.DesignVariable(value=1.0)
+        w = relentless.GeometricMean(u, z)
+        self.assertAlmostEqual(w.value, 1.0)
+        self.assertCountEqual(w.params, ('a','b'))
+        self.assertDictEqual({p:v for p,v in w.depends}, {'a':u,'b':z})
```
mphoward, line 496: You can probably remove this test.

tests/core/test_variable.py
```diff
@@ -0,0 +1,558 @@
...
+
+        #test same-object dependencies
+        w = relentless.GeometricMean(u, u)
```
mphoward, line 499: Can you do this test with v rather than u, so that it actually multiplies to a nontrivial value?

relentless/core/variable.py
```diff
@@ -0,0 +1,550 @@
...
+    At minimum, a variable must provide the interface to get its value as a property.
+
+    """
+    @property
+    @abc.abstractmethod
+    def value(self):
+        pass
+
+class IndependentVariable(Variable):
+    """Abstract base class for an independent variable.
```
mphoward, line 25:
```suggestion
    """Variable representing one independent quantity.
```

relentless/core/variable.py
```diff
@@ -0,0 +1,550 @@
...
+
+    Parameters
+    ----------
+    value : scalar
+        Value of the variable.
+
+    """
+    def __init__(self, value):
+        self._value = value
```
mphoward, line 34:
```suggestion
        self.value = value
```

relentless/core/variable.py
```diff
@@ -0,0 +1,550 @@
...
+        Raises
+        ------
+        ValueError
+            If the parameter argument is not 'a' or 'b'.
+
+        """
+        if param == 'a':
+            return 0.5*np.sqrt(self.b.value/self.a.value)
+        elif param == 'b':
+            return 0.5*np.sqrt(self.a.value/self.b.value)
```
mphoward, line 548: Should we add a special case to each of these here for when the denominator is zero? It should return `np.inf` in those cases and not do a divide by zero.

tests/core/test_variable.py
```diff
@@ -0,0 +1,588 @@
+"""Unit tests for core.variable module."""
+import unittest
+
+import numpy as np
+
+import relentless
+
```
mphoward, line 7: Can you add a short test for the IndependentVariable class? Just that it can be initialized and have its value set and changed.

tests/core/test_variable.py
```diff
@@ -0,0 +1,588 @@
...
+        self.assertAlmostEqual(w.value, 6.0)
+
+        #test creation with vardicts and kwvars
+        w = DepVar({'t':t, 'u':u}, v=v)
+        self.assertCountEqual(w.params, ('t','u','v'))
+        self.assertDictEqual({p:v for p,v in w.depends}, {'t':t,'u':u,'v':v})
+        self.assertAlmostEqual(w.value, 6.0)
+
+        #test creation with repeated attributes
+        w = DepVar({'t':t, 'u':u}, v=v, x=u)
```
mphoward, line 197: By repeated attributes, I meant that `'u'` should appear twice, like:
```py
w = DepVar({'t':t, 'u':u}, u=v)
```
which should mean that `w.u is v`. Can you modify accordingly?

tests/core/test_variable.py
```diff
@@ -0,0 +1,588 @@
...
+        self.assertCountEqual(g.edges[q,t]['params'], ['t','u','v'])
+
+        q = DepVar(t=t, u=t, v=v, w=v)
+        g = q.dependency_graph()
+        self.assertCountEqual(g.nodes, [t,v,q])
+        self.assertCountEqual(g.edges, [(q,t),(q,v)])
+        self.assertCountEqual(g.edges[q,t]['params'], ['t','u'])
+        self.assertCountEqual(g.edges[q,v]['params'], ['v','w'])
+
+        #test circular dependencies
```
mphoward, line 262: For all of these tests, could you call the method `DependentVariable._assert_acyclic(g)` and confirm that it throws? Maybe you should also call it once above on a graph free from cycles.

relentless/core/variable.py
```diff
@@ -0,0 +1,550 @@
...
+        paths = nx.all_simple_paths(g, source=self, target=var)
+        for path in map(nx.utils.pairwise, paths):
+            deriv += np.prod([g.edges[edge]['deriv'] for edge in path])
+        return deriv
+
+    @abc.abstractmethod
+    def _derivative(self, param):
+        pass
+
+    def _assert_variable(self, v):
```
mphoward, line 372: Let's make this a classmethod, since it isn't specific to an instance:
```suggestion
    @classmethod
    def _assert_variable(cls, v):
```

relentless/core/variable.py
```diff
@@ -0,0 +1,550 @@
...
+        #Checks if the dependent variable depends on another variable.
+        if not isinstance(v, Variable):
+            raise TypeError('Dependent variables can only depend on other variables.')
+        return v
+
+    def _assert_acyclic(self, g):
```
mphoward, line 378:
```suggestion
    @classmethod
    def _assert_acyclic(cls, g):
```

tests/core/test_variable.py
```diff
@@ -0,0 +1,588 @@
...
+        self.assertAlmostEqual(dw, 0.5*np.sqrt(2.0))
+
+        #test w.r.t. b
+        dw = w._derivative('b')
+        self.assertAlmostEqual(dw, 0.5*np.sqrt(0.5))
+
+        #test w.r.t. ~a,~b
+        with self.assertRaises(ValueError):
+            w._derivative('c')
+
```
mphoward, line 586: Add a test of the derivative for the cases where 'a' and/or 'b' are zero.

Issue #8, opened 2019-12-19
Parameter derivatives


mphoward:
These are taken numerically using numdifftools, but this may lead to certain variables taking on unphysical values if the delta is too large. (For example, a negative sigma in the LJ potential.) Some variables may need hard bounds to avoid singularities. We need to find a reliable way to take these derivatives.

mphoward:
There are also currently issues with the way the derivatives are resolved with respect to the cutoff. We need to look into this more carefully and correct it. In particular, what is the mode of truncation for the potentials? Are we doing it on the energy or the force? Need to take integrals and derivatives explicitly depending on this functional form.

mphoward:
I believe we have resolved this!

PR #32, opened 2020-08-10
Expand functionality for PairPotential derivative method

adisr04:
- Modify PairPotential derivative to accept object argument, apply chain rule for DependentVariables
- Change compatibility for taking derivative w.r.t. rmin or rmax, include shift effects
- Add PairParameters dependent_variables method
- Add respective unit test cases

### DIFF
--- a/relentless/potential/potential.py
+++ b/relentless/potential/potential.py
@@ -410,6 +410,9 @@ class PairPotential(abc.ABC):
         ------
         ValueError
             If any value in `r` is negative.
+        TypeError
+            If the parameter with respect to which to take the derivative
+            is not a :py:class:`Variable`.
 
         """
         params = self.coeff.evaluate(pair)
@@ -419,23 +422,41 @@ class PairPotential(abc.ABC):
         if not isinstance(var, Variable):
             raise TypeError('Parameter with respect to which to take the derivative must be a Variable.')
 
-        # only evaluate at points inside [rmin,rmax], if specified
         flags = np.ones(r.shape[0], dtype=bool)
-        if params['rmin'] is not False:
-            flags[r < params['rmin']] = False
-        if params['rmax'] is not False:
-            flags[r > params['rmax']] = False
 
-        # evaluate with respect to parameter
-        for p in self.coeff.params:
-            p_obj = self.coeff[pair][p]
-            if p=='rmin' or p=='rmax' or p=='shift':
-                continue
-            elif isinstance(p_obj, DependentVariable):
-                deriv[flags] += (self._derivative(p, r[flags], **params)
-                                *p_obj.derivative(var))
-            elif isinstance(p_obj, IndependentVariable) and var is p_obj:
-                deriv[flags] += self._derivative(p, r[flags], **params)
+        # evaluate with respect to rmin, if set
+        if params['rmin'] is not False and var is self.coeff[pair]['rmin']:
+            flags[r >= params['rmin']] = False
+            deriv[flags] = self._derivative('rmin', r[flags], **params)
+
+        # evaluate with respect to rmax, if set
+        elif params['rmax'] is not False and var is self.coeff[pair]['rmax']:
+            flags[r <= params['rmax']] = False
+            deriv[flags] = self._derivative('rmax', r[flags], **params)
+            if params['shift']:
+                deriv -= self._derivative('rmax', r, **params)
+
+        # evaluate with respect to other parameter
+        else:
+            if params['rmin'] is not False:
+                below = r < params['rmin']
+                deriv[below] = self._derivative('rmin', r[below], **params)
+                flags[below] = False
+            if params['rmax'] is not False:
+                above = r > params['rmax']
+                deriv[above] = self._derivative('rmax', r[above], **params)
+                flags[above] = False
+            for p in self.coeff.params:
+                p_obj = self.coeff[pair][p]
+                if p=='rmin' or p=='rmax' or p=='shift':
+                    continue
+                if isinstance(p_obj, DependentVariable):
+                    deriv[flags] += (self._derivative(p, r[flags], **params)
+                                    *p_obj.derivative(var))
+                elif isinstance(p_obj, IndependentVariable) and var is p_obj:
+                    deriv[flags] += self._derivative(p, r[flags], **params)
+            if params['shift']:
+                deriv -= self._derivative('rmax', r, **params)
 
         # coerce derivative back into shape of the input
         if scalar_r:
--- a/relentless/potential/potential.py
+++ b/relentless/potential/potential.py
@@ -307,7 +307,7 @@ class PairPotential(abc.ABC):
         ValueError
             If any value in `r` is negative.
         ValueError
-            If the potential is to be shifted without setting `rmax`.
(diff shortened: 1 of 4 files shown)


relentless/potential/potential.py
```diff
@@ -391,14 +416,27 @@ def derivative(self, pair, param, r):
...
         if params['rmin'] is not False:
             flags[r < params['rmin']] = False
         if params['rmax'] is not False:
             flags[r > params['rmax']] = False
-        deriv[flags] = self._derivative(param, r[flags], **params)
+
+        # evaluate with respect to parameter
+        deriv[flags] = 0
+        d_params = [i for i in self.coeff.params if not(i=='rmin' or i=='rmax' or i=='shift')]
+        for p in d_params:
```
mphoward, line 432: We should use this PR to fix issues with the derivatives so that all parameters, including `rmin`, `rmax`, and `shift`, are correctly included (#8). It should be legal to take a derivative with respect to `rmin` and `rmax`, including chain rules, but we will need to use the `_force` method rather than the `_derivative` method. Also, `shift` is never going to be a variable (it is true/false), but we will need to include the effects of shifting too. I have some notes on this in a Mathematica notebook that we can finish working through.

For now, let's correct this loop to just run over `self.coeff.params`, and put in `if` statements that just `continue` the loop if `p` is 'rmin', 'rmax', or 'shift'.

relentless/potential/potential.py
```diff
@@ -391,14 +416,27 @@ def derivative(self, pair, param, r):
...
+            p_obj = self.coeff[pair][p]
+            if isinstance(p_obj, DependentVariable):
+                deriv[flags] += (p_obj.derivative(var)
+                                *self._derivative(p, r[flags], **params))
```
mphoward, line 436: To make the intent clear (as a chain rule), can you reverse the order of these operations?

relentless/potential/potential.py
```diff
@@ -391,14 +416,27 @@ def derivative(self, pair, param, r):
...
+            elif isinstance(p_obj, IndependentVariable) and var is p_obj:
+                deriv[flags] = self._derivative(p, r[flags], **params)
+                break
```
mphoward, line 439: I think you don't want to break here? What happens in a case like:
```py
a = DesignVariable(1.0)
u['A','A'] = {'a': a, 'b': SameAs(a)}
```
If you get to 'a' before 'b', then the loop will terminate without evaluating the chain rule (du/db)(db/da).

Can we add a unit test that covers a case like this?

tests/potential/test_pair.py
```diff
@@ -230,14 +230,16 @@ def test_derivative(self):
         s = relentless.potential.Spline(types=('1',), num_knots=3)
         s.from_array(pair=('1','1'), r=r_arr, u=u_arr)
         d_actual = np.array([1.125,0.625,0])
-        d = s.derivative(pair=('1','1'), param='knot-1', r=[1.5,2.5,3.5])
+        param = [i for i in s.knots(('1','1'))][1][1]
```
mphoward, line 233: You can convert a generator to a sequence via the constructor:
```suggestion
        param = list(s.knots(('1','1')))[1][1]
```

tests/potential/test_potential.py
```diff
@@ -289,6 +289,21 @@ def test_save(self):
...
+        a = relentless.DesignVariable(value=1.0)
+        b = relentless.DesignVariable(value=2.0)
+        c = relentless.SameAs(a=b)
+        d = relentless.ArithmeticMean(a=b, b=c)
+
+        m = relentless.potential.PairParameters(types=('A',), params=('energy','mass','charge'))
+        m['A','A']['energy'] = 1.5
+        m['A','A']['mass'] = a
+        m['A','A']['charge'] = d
+
```
mphoward, line 303: Let's add one more simple test first: if `a` is just used for all 3 parameters (should return a 1 element tuple).

relentless/potential/potential.py
```diff
@@ -391,14 +416,27 @@ def derivative(self, pair, param, r):
...
         # only evaluate at points inside [rmin,rmax], if specified
         flags = np.ones(r.shape[0], dtype=bool)
         if params['rmin'] is not False:
             flags[r < params['rmin']] = False
         if params['rmax'] is not False:
             flags[r > params['rmax']] = False
-        deriv[flags] = self._derivative(param, r[flags], **params)
+
+        # evaluate with respect to parameter
+        d_params = [i for i in self.coeff.params if not(i=='rmin' or i=='rmax' or i=='shift')]
```
mphoward, line 430: Don't need this line anymore.

tests/potential/test_potential.py
```diff
@@ -519,45 +567,92 @@ def test_force(self):
...
+    #def _derivative(self, param, r, **params):
+    #    r,d,s = self._zeros(r)
+    #    if param == 'x':
+    #        d = 2*r
+    #    elif param == 'y':
+    #        d = 3*r
+    #    if s:
+    #        d = d.item()
+    #    return d
```
mphoward, line 642: Remove commented lines.

relentless/potential/potential.py
```diff
@@ -385,20 +410,65 @@ def derivative(self, pair, param, r):
...
+                    below = r < params['rmin']
+                    deriv[below] += -self._force(params['rmin'], **params)
+                above = np.zeros(r.shape[0], dtype=bool)
+                if params['rmax'] is not False:
+                    above = r > params['rmax']
+                    deriv[above] += -self._force(params['rmax'], **params)
+                flags = np.logical_and(~below, ~above)
+                deriv[flags] += self._derivative(p, r[flags], **params)*dp_dvar
+                if params['shift']:
+                    deriv[flags] -= -self._force(params['rmax'], **params)
```
mphoward, line 471: Same thing here. Also, make sure to add an error check (as in the `energy()` method) that it is an error to have `shift` be True and `rmax` not set. It will be a similar if-elif.

tests/potential/test_potential.py
```diff
@@ -519,44 +575,96 @@ def test_force(self):
...
+        """Test derivative method with different param types."""
+        q = LinPot(types=('1',), params=('m',))
+        x = relentless.DesignVariable(value=4.0)
+        y = relentless.DesignVariable(value=64.0)
+        z = relentless.GeometricMean(x, y)
+        q.coeff['1','1']['m'] = z
+
+        #test with respect to dependent variable parameter
+        d = q.derivative(pair=('1','1'), var=z, r=2.0)
+        #self.assertAlmostEqual(d, 2.0)
```
mphoward, line 642: Is this test supposed to be commented out?
```suggestion
        self.assertAlmostEqual(d, 2.0)
```

tests/potential/test_potential.py
```diff
@@ -431,6 +479,14 @@ def test_zeros(self):
         np.testing.assert_allclose(u, u_copy)
         self.assertEqual(s, False)
 
+        #test with 1-element array
+        r = [0.2]
+        r_copy, u, s = p._zeros(r)
+        u_copy = np.zeros(1)
+        np.testing.assert_allclose(r_copy, r)
```
mphoward, line 486: For consistency with next line:
```suggestion
        np.testing.assert_allclose(r, r_copy)
```

PR #33, opened 2020-08-13
Construct Depletion Cutoff as DependentVariable dependent on scalars

adisr04:
- Allow DependentVariable to have scalar attributes, cast as IndependentVariables
- Remove cases for _energy/_force/_derivative calculation below low bound and above high bound
- Construct cutoff for Depletion potential as DependentVariables
- Set the rmax Depletion potential coefficient as the cutoff
- Add unit tests for all new cases

### DIFF
--- a/relentless/core/variable.py
+++ b/relentless/core/variable.py
@@ -377,6 +377,8 @@ class DependentVariable(Variable):
         #Casts a scalar into an independent variable.
         if np.isscalar(v):
             v = IndependentVariable(value=v)
+        if not isinstance(v, Variable):
+            raise TypeError('Dependent variables can only depend on scalars or other variables.')
         return v
 
     @classmethod
--- a/relentless/core/variable.py
+++ b/relentless/core/variable.py
@@ -261,15 +261,13 @@ class DependentVariable(Variable):
             raise AttributeError('No attributes specified for DependentVariable.')
 
         for k,v in attrs.items():
-            if np.isscalar(v):
-                v = IndependentVariable(value=v)
-            super().__setattr__(k,self._assert_variable(v))
+            super().__setattr__(k,self._make_variable(v))
         self._params = tuple(attrs.keys())
 
     def __setattr__(self, name, value):
         #Sets the value of the variable on which the specified DependentVariable depends.
         if name != '_params' and name in self.params:
-            value = self._assert_variable(value)
+            value = self._make_variable(value)
         super().__setattr__(name,value)
 
     @property
@@ -375,10 +373,10 @@ class DependentVariable(Variable):
         pass
 
     @classmethod
-    def _assert_variable(cls, v):
-        #Checks if the dependent variable depends on another variable.
-        if not isinstance(v, Variable):
-            raise TypeError('Dependent variables can only depend on other variables.')
+    def _make_variable(cls, v):
+        #Casts a scalar into an independent variable.
+        if np.isscalar(v):
+            v = IndependentVariable(value=v)
         return v
 
     @classmethod
--- a/relentless/core/variable.py
+++ b/relentless/core/variable.py
@@ -237,6 +237,9 @@ class DependentVariable(Variable):
     value property, it must also supply a derivative method with respect to
     any of its dependent variables.
 
+    The dependent variable attribute can also be set as a scalar value, but during
+    construction the value is casted to a :py:class:`IndepedentVariable`.
+
     The number of dependencies of the variable are locked at construction, but
     their values can be modified.
 
@@ -258,6 +261,8 @@ class DependentVariable(Variable):
             raise AttributeError('No attributes specified for DependentVariable.')
 
         for k,v in attrs.items():
+            if np.isscalar(v):
+                v = IndependentVariable(value=v)
             super().__setattr__(k,self._assert_variable(v))
         self._params = tuple(attrs.keys())
 
--- a/relentless/potential/pair.py
+++ b/relentless/potential/pair.py
@@ -311,14 +311,14 @@ class Spline(PairPotential):
             f = f.item()
         return f
 
-    def derivative(self, pair, param, r):
+    def derivative(self, pair, var, r):
         #Extending PairPotential method to check if r and knot values are DesignVariables.
         for ri,ki in self.knots(pair):
             if not isinstance(ri, core.DesignVariable):
                 raise TypeError('All r values must be DesignVariables')
             if not isinstance(ki, core.DesignVariable):
                 raise TypeError('All knot values must be DesignVariables')
-        return super().derivative(pair, param, r)
+        return super().derivative(pair, var, r)
 
     def _derivative(self, param, r, **params):
         """Evaluates the spline parameter derivative using finite differencing.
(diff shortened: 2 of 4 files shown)


relentless/potential/pair.py
```diff
@@ -589,6 +589,109 @@ class Depletion(PairPotential):
         List of types (A type must be a str).
 
     """
+
+    class LowBound(core.DependentVariable):
```
mphoward, line 593: You could cheat and subclass `ArithmeticMean` for this one
```py
class LowBound(core.ArithmeticMean):
    pass
```
to recycle code. The only tradeoff is that the parameters will be named `a` and `b`.

relentless/potential/pair.py
```diff
@@ -624,14 +727,14 @@ def _energy(self, r, P, sigma_i, sigma_j, sigma_d, **params):
         r,u,s = self._zeros(r)
 
         #clamp lo
-        lo = 0.5*(sigma_i + sigma_j)
+        lo = Depletion.LowBound(sigma_i, sigma_j).value
```
mphoward, line 730: Because `sigma_i` and `sigma_j` could be design parameters, we actually need to set these variable in the coefficient dictionary directly, not in the `_energy`, `_force`, or `_derivative` methods.

The cleanest way to do this is to override the `energy`, `force`, and `derivative` methods. In these, first check if `rmin` and/or `rmax` has been set. If not, then replace the coefficient with the appropriate bound, then call the parent method:
```py
def energy(self, pair, r):
    if self.coeff[pair]['rmin'] is False:
       self.coeff[pair]['rmax'] = self.LowBound(self.coeff[pair]['sigma_i'],self.coeff[pair]['sigma_j'])
    if self.coeff[pair]['rmax'] is False:
        self.coeff[pair]['rmax'] = self.HighBound(self.coeff[pair]['sigma_i'],self.coeff[pair]['sigma_j'],self.coeff[pair]['sigma_d'])
    return super().energy(pair,r)
```

I think the only thing we need to be careful about is that we currently are not actually using a "low bound" in the sense of what we've defined in the rest of the code, since we are extrapolating the force rather than zeroing it. If this is the case, do not set `rmin`, define `lo = 0.5*(sigma_i+sigma_j)`, and we will need to be careful about computing the derivatives appropriately.

relentless/core/variable.py
```diff
@@ -258,6 +261,8 @@ def __init__(self, *vardicts, **kwvars):
             raise AttributeError('No attributes specified for DependentVariable.')
 
         for k,v in attrs.items():
+            if np.isscalar(v):
```
mphoward, line 264: We also need to catch this in the `__setattr__` method below. I think probably what we should do is change `_asert_variable` to a `_make_variable`, and have it do this check rather than raise an error. This would mean you don't need to change the `__init__` or `__setattr__` methods, other than using the new method call.

relentless/core/variable.py
```diff
@@ -370,10 +373,10 @@ def _derivative(self, param):
...
 
     @classmethod
-    def _assert_variable(cls, v):
-        #Checks if the dependent variable depends on another variable.
-        if not isinstance(v, Variable):
-            raise TypeError('Dependent variables can only depend on other variables.')
+    def _make_variable(cls, v):
+        #Casts a scalar into an independent variable.
+        if np.isscalar(v):
+            v = IndependentVariable(value=v)
```
mphoward, line 379: After you do this conversion, you should probably keep the:
```py
if not isinstance(v,Variable):
    raise ...
```
For example, if `v` is a list, it will not get converted to an `IndependentVariable`, but the code will silently proceed.

relentless/potential/pair.py
```diff
@@ -589,9 +589,123 @@ class Depletion(PairPotential):
...
+                raise ValueError('Unknown parameter')
+
     def __init__(self, types, shift=False):
         super().__init__(types=types, params=('P','sigma_i','sigma_j','sigma_d'))
 
+    def energy(self, pair, r):
+        # Override parent method to set rmin and rmax as low and high bounds
+        if self.coeff[pair]['rmin'] is False:
+            self.coeff[pair]['rmin'] = self.LowBound(self.coeff[pair]['sigma_i'],
+                                                     self.coeff[pair]['sigma_j'])
```
mphoward, line 702: Let's not set `rmin` by default. The user can always construct this parameter themselves.

(Also apply to the `force` and `derivative` methods.)

tests/core/test_variable.py
```diff
@@ -264,6 +264,12 @@ def test_init(self):
         self.assertDictEqual({p:v for p,v in w.depends}, {'t':t,'u':u,'v':v})
         self.assertAlmostEqual(w.value, 6.0)
 
+        #test creation with scalar attributes
+        w = DepVar(t=1.0, u=2.0, v=3.0)
+        self.assertCountEqual(w.params, ('t','u','v'))
+        self.assertDictEqual({p:v.value for p,v in w.depends}, {'t':1.0,'u':2.0,'v':3.0})
+        self.assertAlmostEqual(w.value, 6.0)
+
```
mphoward, line 272: Can you try a test where you set one of these attributes (`w.t = 4.0`) as a scalar after construction?

relentless/potential/pair.py
```diff
@@ -589,9 +589,123 @@ class Depletion(PairPotential):
         List of types (A type must be a str).
 
     """
+
+    class LowBound(core.DependentVariable):
```
mphoward, line 593: Let's remove this class if we are not going to enforce it automatically. The user can use `ArithmeticMean` if they want to do this explicitly.

relentless/potential/pair.py
```diff
@@ -589,9 +589,123 @@ class Depletion(PairPotential):
...
+
+            """
+            if param == 'sigma_i':
+                return 0.5
+            elif param == 'sigma_j':
+                return 0.5
+            else:
+                raise ValueError('Unknown parameter')
+
+    class HighBound(core.DependentVariable):
```
mphoward, line 642: Let's rename this `Cutoff` to make it a little more self-explanatory. the user would call it like `Depletion.Cutoff`

tests/potential/test_pair.py
```diff
@@ -334,6 +334,112 @@ def test_derivative(self):
...
+
+        #calculate w.r.t. sigma_j
+        dw = w._derivative('sigma_j')
+        self.assertEqual(dw, 0.5)
+
+        #invalid parameter calculation
+        with self.assertRaises(ValueError):
+            dw = w._derivative('sigma')
+
+class test_HighBound(unittest.TestCase):
```
mphoward, line 387: You can probably just put these tests as functions inside `test_Depletion`, so it's obvious what class they belong to.

tests/potential/test_pair.py
```diff
@@ -334,6 +334,112 @@ def test_derivative(self):
         with self.assertRaises(ValueError):
             u = y._derivative(param='kapppa', r=r_input, epsilon=1.0, kappa=1.0)
 
+class test_LowBound(unittest.TestCase):
```
mphoward, line 337: Remove these tests once you remove the class.

PR #35, opened 2020-09-02
Documentation, features, and unit tests for Ensemble

adisr04:
- Add `RDF` and `Ensemble` documentation
- Implement documented `Volume` abstract base class and concrete subclasses `Parallelepiped`, `TriclinicBox`, `Cuboid`, and `Cube`
- Expand and modify `Ensemble` features/functionalities
- Implement unit tests for `RDF`, `Volume` subclasses, and `Ensemble` in `test_ensemble.py`

### DIFF
--- a/relentless/core/ensemble.py
+++ b/relentless/core/ensemble.py
@@ -5,9 +5,11 @@ import json
 
 import numpy as np
 
-from . import core
+from .collections import FixedKeyDict,PairMatrix
+from .math import Interpolator
+from .volume import *
 
-class RDF(core.Interpolator):
+class RDF(Interpolator):
     r"""Radial distribution function.
 
     Represents the pair distribution function :math:`g(r)` as an ``(N,2)``
@@ -102,13 +104,13 @@ class Ensemble(object):
         self.V = V
 
         # mu-N
-        self._mu = core.FixedKeyDict(keys=types)
-        self._N = core.FixedKeyDict(keys=types)
+        self._mu = FixedKeyDict(keys=types)
+        self._N = FixedKeyDict(keys=types)
         self.mu.update(mu)
         self.N.update(N)
 
         # rdf per-pair
-        self._rdf = core.PairMatrix(types=types)
+        self._rdf = PairMatrix(types=types)
 
         # build the set of constant variables from the constructor
         self._constant = {'T':True,
@@ -139,7 +141,7 @@ class Ensemble(object):
 
     @V.setter
     def V(self, value):
-        if value is not None and not isinstance(value, core.Volume):
+        if value is not None and not isinstance(value, Volume):
             raise TypeError('V can only be set as a Volume object or as None.')
         self._V = value
 
@@ -273,7 +275,7 @@ class Ensemble(object):
         # retrieve thermodynamic parameters
         thermo = {'T':data['T'],
                   'P':data['P'],
-                  'V':getattr(core,data['V']['__name__']).from_json(data['V']['data']),
+                  'V':globals()[data['V']['__name__']].from_json(data['V']['data']),
                   'mu':data['mu'],
                   'N':data['N']
                  }
--- a/relentless/core/ensemble.py
+++ b/relentless/core/ensemble.py
@@ -40,7 +40,7 @@ class Ensemble(object):
     The variables that are "constants" of the ensemble (e.g. *N*, *V*, and *T*
     for the canonical ensemble) are determined from the arguments specified in the
     constructor. The fluctuating (conjugate) variables must be set subsequently.
-    It is an error to set both variables in a conjugate pair (e.g. *P* and *V)
+    It is an error to set both variables in a conjugate pair (e.g. *P* and *V*)
     during construction.
 
     Parameters
@@ -68,6 +68,8 @@ class Ensemble(object):
         If all values of ``N`` are not integers or None.
     ValueError
         If both ``mu`` and ``N`` are set for a type.
+    ValueError
+        If both ``mu`` and ``N`` are None.
 
     """
     def __init__(self, T, P=None, V=None, mu=None, N=None, kB=1.0):
@@ -113,7 +115,7 @@ class Ensemble(object):
         self._rdf = PairMatrix(types=types)
 
         # build the set of constant variables from the constructor
-        self._constant = {'T':True,
+        self._constant = {'T': True,
                           'P': self.P is not None,
                           'V': self.V is not None,
                           'mu': {t: (self.mu[t] is not None) for t in types},
@@ -161,7 +163,7 @@ class Ensemble(object):
 
     @property
     def mu(self):
-        r""":py:class:`FixedKeyDict`: Chemical potential for each type."""
+        r""":py:class:`FixedKeyDict`: Chemical potential of each type."""
         return self._mu
 
     @property
@@ -236,7 +238,7 @@ class Ensemble(object):
                 'V': {'__name__':type(self.V).__name__,
                       'data':self.V.to_json()
                      },
-                'mu':self.mu.todict(),
+                'mu': self.mu.todict(),
                 'N': self.N.todict(),
                 'kB': self.kB,
                 'constant': self.constant,
@@ -273,11 +275,11 @@ class Ensemble(object):
             data = json.load(f)
 
         # retrieve thermodynamic parameters
-        thermo = {'T':data['T'],
-                  'P':data['P'],
-                  'V':globals()[data['V']['__name__']].from_json(data['V']['data']),
-                  'mu':data['mu'],
-                  'N':data['N']
+        thermo = {'T': data['T'],
+                  'P': data['P'],
+                  'V': globals()[data['V']['__name__']].from_json(data['V']['data']),
+                  'mu': data['mu'],
+                  'N': data['N']
                  }
 
         # create Ensemble with constant variables only
--- a/relentless/core/volume.py
+++ b/relentless/core/volume.py
@@ -1,0 +1,399 @@
+__all__ =  ['Volume', 'Cube','Cuboid','Parallelepiped','TriclinicBox']
+
+import abc
+from enum import Enum
+
+import numpy as np
+
+class Volume(abc.ABC):
+    r"""Abstract base class defining a region of space.
+
+    A Volume can be any region of space; typically, it is a simulation "box".
+    Any deriving class must implement the volume method that computes the scalar
+    volume of the region. Additionally, methods to serialize and deserialize a
+    Volume must be specified so that the object can be saved to disk.
+
+    """
+    @property
+    @abc.abstractmethod
+    def volume(self):
+        r"""float: Volume of the region."""
+        pass
+
+    @abc.abstractmethod
+    def to_json(self):
+        r"""Serialize as a dictionary.
+
+        The serialized data can be saved to file as JSON data.
+
+        Returns
+        -------
+        dict
+            The serialized :py:class:`Volume` data.
+
+        """
+        pass
+
+    @classmethod
+    @abc.abstractmethod
+    def from_json(cls, data):
+        r"""Deserialize from a dictionary.
+
+        Returns
+        -------
+        Volume
+            The object reconstructed from the ``data``.
+
+        """
+        pass
+
+class Parallelepiped(Volume):
+    r"""Parallelepiped box defined by three vectors.
+
+    The three vectors **a**, **b**, and **c** must form a right-hand basis
+    so that the box volume *V* is positive:
+
+    .. math::
+
+        V = (\mathbf{a} \cross \mathbf{b}) \cdot \mathbf{c} > 0
+
+    Parameters
+    ----------
+    a : array_like
+        First vector defining the parallelepiped.
+    b : array_like
+        Second vector defining the parallelepiped.
+    c : array_like
+        Third vector defining the parallelepiped.
+
+    Raises
+    ------
+    TypeError
+        If a, b, and c are not all 3-element vectors.
+    ValueError
+        If the volume is not positive.
+
+    """
+    def __init__(self, a, b, c):
+        self.a = np.asarray(a,dtype=np.float64)
+        self.b = np.asarray(b,dtype=np.float64)
+        self.c = np.asarray(c,dtype=np.float64)
+        if not (self.a.shape==(3,) and self.b.shape==(3,) and self.c.shape==(3,)):
+            raise TypeError('a, b, and c must be 3-element vectors.')
+        if self.volume <= 0:
+            raise ValueError('The volume must be positive.')
+
+    @property
+    def volume(self):
+        return np.dot(np.cross(self.a,self.b),self.c)
+
+    def to_json(self):
+        r"""Serialize as a dictionary.
+
+        The dictionary contains the three box vectors **a**, **b**, and **c**
+        as tuples.
+
+        Returns
+        -------
+        dict
+            The serialized Parallelepiped.
+
+        """
+        return {'a': tuple(self.a),
+                'b': tuple(self.b),
+                'c': tuple(self.c)
+               }
+
+    @classmethod
+    def from_json(cls, data):
+        r"""Deserialize from a dictionary.
+
+        Parameters
+        ----------
+        data : dict
+            The serialized equivalent of the Parallelepiped object. The keys
+            of ``data`` should be ``('a','b','c')``, and the data for each
+            is the 3-element box vector.
+
+        Returns
+        -------
+        :py:class:`Parallelepiped`
+            A new Parallelepiped object constructed from the data.
+
+        """
+        return Parallelepiped(**data)
+
+class TriclinicBox(Parallelepiped):
+    r"""Triclinic box.
+
+    A TriclinicBox is a special type of :py:class:`Parallelepiped`. The box is
+    defined by an orthorhombic box oriented along the Cartesian axes and having
+    three vectors of length ``Lx``, ``Ly``, and ``Lz``, respectively. The box is
+    then tilted by factors ``xy``, ``xy``, and ``xz``, which are upper off-diagonal
+    elements of the matrix of box vectors. As a result, the **a** vector is
+    always aligned along the *x*-axis, while the other two vectors may be tilted.
+
+    The tilt factors can be defined using one of two :py:class:`TriclinicBox.Convention`.
+    In the `LAMMPS <https://lammps.sandia.gov/doc/Howto_triclinic.html>`_
+    simulation convention, specified using ``TriclinicBox.Convention.LAMMPS``,
+
+
+    .. math::
+
+        \mathbf{a} = (L_x,0,0)
+        \mathbf{b} = (xy,L_y,0)
+        \mathbf{c} = (xz,yz,L_z)
+
+    In the `HOOMD-blue <https://hoomd-blue.readthedocs.io/en/stable/box.html>`_
+    simulation convention, specified using ``TriclinicBox.Convention.HOOMD``,
+
+    .. math::
+
+        \mathbf{a} = (L_x,0,0)
+        \mathbf{b} = (xy*L_y,L_y,0)
+        \mathbf{c} = (xz*L_z,yz*L_z,L_z)
+
+    Parameters
+    ----------
+    Lx : float
+        Length along the *x* axis.
+    Ly : float
+        Length along the *y* axis.
+    Lz : float
+        Length along the *z* axis.
+    xy : float
+        First tilt factor.
+    xz : float
+        Second tilt factor.
+    yz : float
+        Third tilt factor.
+
+    Raises
+    ------
+    ValueError
+        If *Lx*, *Ly*, *Lz* are not all positive.
+    ValueError
+        If the convention is not `TriclinicBox.Convention.LAMMPS` or
+        `TriclinicBox.Convention.HOOMD`.
+
+    """
+
+    class Convention(Enum):
+        r"""Convention by which the tilt factors are applied to the basis vectors.
+
+        Calculation of the basis vectors by the `LAMMPS convention <https://lammps.sandia.gov/doc/Howto_triclinic.html>`_:
+
+        .. math::
+
+            \mathbf{a} = (L_x,0,0)
+            \mathbf{b} = (xy,L_y,0)
+            \mathbf{c} = (xz,yz,L_z)
+
+        Calculation of the basis vectors by the `HOOMD convention <https://hoomd-blue.readthedocs.io/en/stable/box.html>`_:
+
+        .. math::
+
+            \mathbf{a} = (L_x,0,0)
+            \mathbf{b} = (xy*L_y,L_y,0)
+            \mathbf{c} = (xz*L_z,yz*L_z,L_z)
+
+        Attributes
+        ----------
+        LAMMPS : int
+            LAMMPS convention for applying the tilt factors.
+        HOOMD : int
+            HOOMD convention for applying the tilt factors.
+
+        """
+        LAMMPS = 1
+        HOOMD = 2
+
+    def __init__(self, Lx, Ly, Lz, xy, xz, yz, convention=Convention.LAMMPS):
+        if Lx<=0 or Ly<=0 or Lz<= 0:
+            raise ValueError('All side lengths must be positive.')
+        self._convention = convention
+        if self._convention is TriclinicBox.Convention.LAMMPS:
+            a = (Lx,0,0)
+            b = (xy,Ly,0)
+            c = (xz,yz,Lz)
+        elif self._convention is TriclinicBox.Convention.HOOMD:
+            a = (Lx,0,0)
+            b = (xy*Ly,Ly,0)
+            c = (xz*Lz,yz*Lz,Lz)
+        else:
+            raise ValueError('Triclinic convention must be TriclinicBox.Convention.LAMMPS or TriclinicBox.Convention.HOOMD')
+        super().__init__(a,b,c)
+
+    @property
+    def convention(self):
+        r""":py:class:`TriclinicBox.Convention`: Convention for tilt factors."""
+        return self._convention
+
+    def to_json(self):
+        r"""Serialize as a dictionary.
+
+        The dictionary contains the three box lengths ``Lx``, ``Ly``, and ``Lz``,
+        the three tilt factors ``xy``, ``xz``, and ``yz``, and the ``convention``
+        for the tilt factors.
+
+        Returns
+        -------
+        dict
+            The serialized TriclinicBox.
+
+        """
+        if self._convention is TriclinicBox.Convention.LAMMPS:
+            xy = self.b[0]
+            xz = self.c[0]
+            yz = self.c[1]
+        elif self._convention is TriclinicBox.Convention.HOOMD:
+            xy = self.b[0]/self.b[1]
+            xz = self.c[0]/self.c[2]
+            yz = self.c[1]/self.c[2]
+        return {'Lx':self.a[0],
+                'Ly':self.b[1],
+                'Lz':self.c[2],
+                'xy':xy,
+                'xz':xz,
+                'yz':yz,
+                'convention':self._convention.name
+               }
+
+    @classmethod
+    def from_json(cls, data):
+        r"""Deserialize from a dictionary.
+
+        Parameters
+        ----------
+        data : dict
+            The serialized equivalent of the TriclinicBox object. The keys
+            of ``data`` should be ``('Lx','Ly','Lz','xy','xz','yz','convention')``.
+            The lengths and tilt factors should be floats, and the convention should
+            be a string.
+
+        Returns
+        -------
+        :py:class:`TriclinicBox`
+            A new TriclinicBox object constructed from the data.
+
+        Raises
+        ------
+        ValueError
+            If the convention specified is not LAMMPS or HOOMD.
+
+        """
+        data_ = dict(data)
+        if data['convention']=='LAMMPS':
+            data_['convention'] = TriclinicBox.Convention.LAMMPS
+        elif data['convention']=='HOOMD':
+            data_['convention'] = TriclinicBox.Convention.HOOMD
+        else:
+            return ValueError('Only LAMMPS and HOOMD conventions are supported.')
+        return TriclinicBox(**data_)
+
+class Cuboid(TriclinicBox):
+    r"""Orthorhombic box.
+
+    A Cuboid is a special type of :py:class:`TriclinicBox`. The three box vectors
+    point along the *x*, *y*, and *z* axes, so they are all orthogonal (i.e. the
+    tilt factors ``xy``, ``xz``, and ``yz`` are all 0). Each vector can have a
+    different length, *Lx*, *Ly*, and *Lz*.
+
+    Parameters
+    ----------
+    Lx : float
+        Length along the *x* axis.
+    Ly : float
+        Length along the *y* axis.
+    Lz : float
+        Length along the *z* axis.
+
+    Raises
+    ------
+    ValueError
+        If *Lx*, *Ly*, *Lz* are not all positive.
+
+    """
+    def __init__(self, Lx, Ly, Lz):
+        super().__init__(Lx,Ly,Lz,0,0,0)
+
+    def to_json(self):
+        r"""Serialize as a dictionary.
+
+        The dictionary contains the three box lengths ``Lx``, ``Ly``, and ``Lz``.
+
+        Returns
+        -------
+        dict
+            The serialized Cuboid.
+
+        """
+        return {'Lx':self.a[0],
+                'Ly':self.b[1],
+                'Lz':self.c[2],
+               }
+
+    @classmethod
+    def from_json(cls, data):
+        r"""Deserialize from a dictionary.
+
+        Parameters
+        ----------
+        data : dict
+            The serialized equivalent of the Cuboid object. The keys
+            of ``data`` should be ``('Lx','Ly','Lz')``, and their values
+            should be floats.
+
+        Returns
+        -------
+        :py:class:`Cuboid`
+            A new Cuboid object constructed from the data.
+
+        """
+        return Cuboid(**data)
+
+class Cube(Cuboid):
+    r"""Cubic box.
+
+    A Cube is a special type of :py:class:`Cuboid` where all vectors have the
+    same length *L*.
+
+    Parameters
+    ----------
+    L : float
+        The edge length of the cube.
+
+    """
+    def __init__(self, L):
+        super().__init__(L,L,L)
+
+    def to_json(self):
+        r"""Serialize as a dictionary.
+
+        The dictionary contains the box length ``L``.
+
+        Returns
+        -------
+        dict
+            The serialized Cube.
+
+        """
+        return {'L':self.a[0]}
+
+    @classmethod
+    def from_json(cls, data):
+        r"""Deserialize from a dictionary.
+
+        Parameters
+        ----------
+        data : dict
+            The serialized equivalent of the Cube object. The keys
+            of ``data`` should be ``('L',)``, and its value should be a float.
+
+        Returns
+        -------
+        :py:class:`Cube`
+            A new Cube object constructed from the data.
+
+        """
+        return Cube(**data)
--- a/relentless/core/volume.py
+++ b/relentless/core/volume.py
@@ -1,4 +1,4 @@
-__all__ =  ['Volume', 'Cube','Cuboid','Parallelepiped','TriclinicBox']
+__all__ =  ['Volume','Cube','Cuboid','Parallelepiped','TriclinicBox']
 
 import abc
 from enum import Enum
@@ -8,7 +8,7 @@ import numpy as np
 class Volume(abc.ABC):
     r"""Abstract base class defining a region of space.
 
-    A Volume can be any region of space; typically, it is a simulation "box".
+    A Volume can be any region of space; typically, it is a simulation "box."
     Any deriving class must implement the volume method that computes the scalar
     volume of the region. Additionally, methods to serialize and deserialize a
     Volume must be specified so that the object can be saved to disk.
@@ -137,7 +137,6 @@ class TriclinicBox(Parallelepiped):
     In the `LAMMPS <https://lammps.sandia.gov/doc/Howto_triclinic.html>`_
     simulation convention, specified using ``TriclinicBox.Convention.LAMMPS``,
 
-
     .. math::
 
         \mathbf{a} = (L_x,0,0)
@@ -212,11 +211,11 @@ class TriclinicBox(Parallelepiped):
         if Lx<=0 or Ly<=0 or Lz<= 0:
             raise ValueError('All side lengths must be positive.')
         self._convention = convention
-        if self._convention is TriclinicBox.Convention.LAMMPS:
+        if self.convention is TriclinicBox.Convention.LAMMPS:
             a = (Lx,0,0)
             b = (xy,Ly,0)
             c = (xz,yz,Lz)
-        elif self._convention is TriclinicBox.Convention.HOOMD:
+        elif self.convention is TriclinicBox.Convention.HOOMD:
             a = (Lx,0,0)
             b = (xy*Ly,Ly,0)
             c = (xz*Lz,yz*Lz,Lz)
@@ -250,13 +249,13 @@ class TriclinicBox(Parallelepiped):
             xy = self.b[0]/self.b[1]
             xz = self.c[0]/self.c[2]
             yz = self.c[1]/self.c[2]
-        return {'Lx':self.a[0],
-                'Ly':self.b[1],
-                'Lz':self.c[2],
-                'xy':xy,
-                'xz':xz,
-                'yz':yz,
-                'convention':self._convention.name
+        return {'Lx': self.a[0],
+                'Ly': self.b[1],
+                'Lz': self.c[2],
+                'xy': xy,
+                'xz': xz,
+                'yz': yz,
+                'convention': self._convention.name
                }
 
     @classmethod
@@ -308,11 +307,6 @@ class Cuboid(TriclinicBox):
     Lz : float
         Length along the *z* axis.
 
-    Raises
-    ------
-    ValueError
-        If *Lx*, *Ly*, *Lz* are not all positive.
-
     """
     def __init__(self, Lx, Ly, Lz):
         super().__init__(Lx,Ly,Lz,0,0,0)
@@ -328,9 +322,9 @@ class Cuboid(TriclinicBox):
             The serialized Cuboid.
 
         """
-        return {'Lx':self.a[0],
-                'Ly':self.b[1],
-                'Lz':self.c[2],
+        return {'Lx': self.a[0],
+                'Ly': self.b[1],
+                'Lz': self.c[2],
                }
 
     @classmethod
@@ -378,7 +372,7 @@ class Cube(Cuboid):
             The serialized Cube.
 
         """
-        return {'L':self.a[0]}
+        return {'L': self.a[0]}
 
     @classmethod
     def from_json(cls, data):
--- a/relentless/ensemble.py
+++ b/relentless/ensemble.py
@@ -12,13 +12,15 @@ import numpy as np
 from . import core
 
 class RDF(core.Interpolator):
-    """Constructs a :py:class:`Interpolator` object interpolating through an RDF
-    function :math:`g(r)`, and creates an `nx2` array with columns as `r` and `g`.
+    r"""Radial distribution function.
+
+    Represents the pair distribution function :math:`g(r)` as an `nx2` table that
+    can also be smoothly interpolated.
 
     Parameters
     ----------
     r : array_like
-        1-d array of r values, that must be continually increasing.
+        1-d array of r values (continually increasing).
     g : array_like
         1-d array of g values.
 
@@ -28,7 +30,7 @@ class RDF(core.Interpolator):
         self.table = np.column_stack((r,g))
 
 class Volume(abc.ABC):
-    """Abstract base class for a defined volume/region."""
+    """Abstract base class defining a region of space."""
     @property
     @abc.abstractmethod
     def volume(self):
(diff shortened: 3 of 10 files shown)


tests/test_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertDictEqual(ens.N, {'A':1,'B':2})
+        self.assertCountEqual(ens.conjugates, ('P',))
+
+        #custom conjugates set, allowing P, V, mu, and N
+        ens = relentless.Ensemble(types=('A','B'), T=100, P=1.5, V=3.0, mu={'A':0.1,'B':0.2}, N={'A':1,'B':2}, conjugates=('q',))
+        self.assertEqual(ens.P, 1.5)
+        self.assertAlmostEqual(ens.V, 3.0)
+        self.assertDictEqual(ens.mu, {'A':0.1,'B':0.2})
+        self.assertDictEqual(ens.N, {'A':1,'B':2})
+        self.assertCountEqual(ens.conjugates, ('q',))
```
zeesherman, line 97: So here are all examples where `conjugates` isn't "correct", in that it doesn't specify all the right conjugate variables.  Might need to change behavior depending on what we decided for my earlier comment on this issue.

tests/test_ensemble.py
```diff
@@ -0,0 +1,158 @@
...
+            ens = relentless.Ensemble(types=('A','B'), T=10, P=2.0)
+
+        #conjugates not set, test invalid setting of both P and V
+        with self.assertRaises(ValueError):
+            ens = relentless.Ensemble(types=('A','B'), T=10, P=2.0, V=3.0)
+
+        #conjugates not set, test invalid setting of both mu and N
+        with self.assertRaises(ValueError):
+            ens = relentless.Ensemble(types=('A','B'), T=10, mu={'A':0.1,'B':0.2}, N={'A':1,'B':2})
+
```
zeesherman, line 140: Pending what we decide on the behavior of `conjugates`, do we include a test where both `P` and `V` (or `mu` and `N`) are included in `conjugates`?

zeesherman (changes requested):

relentless/ensemble.py
```diff
@@ -63,10 +108,13 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
 
     @property
     def beta(self):
+        """float: The thermodynamic beta/coldness."""
         return 1./(self.kB*self.T)
 
     @property
     def V(self):
+        """float or int: The volume of the system; raises an AttributeError if
+        volume is not a conjugate variable."""
```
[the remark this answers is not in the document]
adisr04: I think the convention is to document the setter behavior alongside the getter as seen [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)

mphoward:
To chime in on the `conjugates` discussion, I forget why exactly I thought it was useful to let the user specify which of the two variables is the conjugate in the constructor (probably it was something like being able to set both `P` and `V` in the constructor simultaneously), but I don't think we really need this because
```py
ens = Ensemble(P=0.0, V=10.0)
```
can be replaced by
```py
ens = Ensemble(V=10.0)
ens.P = 0.0
```
and this pattern is likely uncommon anyway. Definitely the current behavior is that it is assumed to be correct, regardless of what the user does.

I would support removing the `conjugates` argument from the constructor, and setting the list of conjugates from the parameters that are specified (including the error checking if both are set simultaneously that is already there). I think this makes the specification of the ensemble pretty clear:
```py
nvt = Ensemble(types=('A',) T=1.0, V=10.0, N={'A': 1})
npt = Ensemble(types=('A',) T=1.0, P=0.0, N={'A':2})
```
Once you have made this change, it would also be good to modify the setters so that all the properties can be modified (regardless of whether they are conjugate or not).

Last, I also want to ask whether we should require the user to specify an explicit type list, or if we should deduce it from `N` / `mu` (whichever is specified). It feels like duplication to make them set it both times, and we are now requiring that one of these be set.

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
 
 class Ensemble(object):
-    def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=None):
-        self.types = tuple(types)
-        self.rdf = core.PairMatrix(self.types)
+    """Constructs a thermodynamic ensemble for a set of types with specified parameters.
+
+    Parameters
+    ----------
+    T : float or int
```
mphoward, line 31: Let's suggest to the user that this should be a float (though they can still use an int)
```suggestion
    T : float
```

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+        Temperature of the system.
+    P : float or int
```
mphoward, line 33:
```suggestion
    P : float
```

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+        Pressure of the system (defaults to `None`).
+    V : float or int
```
mphoward, line 35: I am going to make another suggestion to change the volume to be a new class we will implement, but for now this should at minimum be:
```suggestion
    V : float
```

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+        Volume of the system (defaults to `None`).
+    mu : `dict`
+        The chemical potential for each specified type (defaults to empty `dict`).
+    N : `dict`
+        The number of particles for each specified type (defaults to empty `dict`.
```
mphoward, line 40:
```suggestion
        The number of particles for each specified type (defaults to empty `dict`).
```

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+    """
     def __init__(self, r, g):
         super().__init__(r,g)
         self.table = np.column_stack((r,g))
 
 class Ensemble(object):
-    def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=None):
-        self.types = tuple(types)
-        self.rdf = core.PairMatrix(self.types)
+    """Constructs a thermodynamic ensemble for a set of types with specified parameters.
```
mphoward, line 27:
```suggestion
    """Thermodynamic ensemble.
```
Let's also add some details here about how ensembles are defined. This should include that T is always a constant, then either P or V and either mu or N for each type can be specified. Also, that the types are the ones that are in the mu or N dicts.

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+        The chemical potential for each specified type (defaults to empty `dict`).
+    N : `dict`
+        The number of particles for each specified type (defaults to empty `dict`.
+    kB : float
+        Boltzmann constant (defaults to 1.0).
+
+    Raises
+    ------
+    ValueError
+        If either `P` or `V` are not set.
```
mphoward, line 47:
```suggestion
        If neither `P` nor `V` is set.
```

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+    ValueError
+        If either `mu` or `N` are not set for each type.
```
mphoward, line 49: You should probably remove this, right? Since we don't have a type list, this cannot be an error.

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+    TypeError
+        If all values of N are not integers.
+    ValueError
+        If the conjugates are not set, and both `P` and `V` are set.
+    ValueError
+        If the conjugates are not set, and both `mu` and `N` are set for each type.
```
mphoward, line 55:
```suggestion
    ValueError
        If both `P` and `V` are set.
    ValueError
        If both `mu` and `N` are set for a type.
```

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+
+    """
+    def __init__(self, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=None):
```
mphoward, line 58: Remove the conjugates parameter.
```suggestion
    def __init__(self, T, P=None, V=None, mu={}, N={}, kB=1.0):
```

relentless/ensemble.py
```diff
@@ -8,14 +8,59 @@
...
+        types = []
+        types += [t for t in mu.keys()]
+        types += [t for t in N.keys()]
+        self.types = tuple(set(types))
```
mphoward, line 62: I don't think we should save the types explicitly. Let's just make this list here, and then add the `types` property that returns the keys of `self._N`.
```suggestion
        types = list(mu.keys()) + list(N.keys())
        types = tuple(set(types))
```

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
+        self._N = core.FixedKeyDict(keys=self.types)
         for t in self.types:
             if (t not in mu or mu[t] is None) and (t not in N or N[t] is None):
                 raise ValueError('Either mu or N must be set for type {}.'.format(t))
+            if t in N and N[t] is not None and not isinstance(N[t], int):
+                raise TypeError('All values of N must be integers.')
             if t in N:
                 self._N[t] = N[t]
             if t in mu:
                 self._mu[t] = mu[t]
```
mphoward, line 86: This logic is a little hard to read. You could probably just set these using the `update` methods of the dicts, and then just check the things you need. The types are checked below, so I think all you need to add is that N is all ints.

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
-                    conjugates.append('mu_{}'.format(t))
-                if self.N[t] is None:
-                    conjugates.append('N_{}'.format(t))
-            self._conjugates = tuple(conjugates)
+        # conjugates are deduced from the specified parameters
+        if self.P is not None and self.V is not None:
+            raise ValueError('Both P and V cannot be set.')
+        for t in self.types:
+            if self.mu[t] is not None and self.N[t] is not None:
+                raise ValueError('Both mu and N cannot be set for type {}.'.format(t))
```
mphoward, line 93: This part is already error checking the type keys. So, you could skip this check above and just use this one.

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
-        if 'P' in self.conjugates:
-            self._P = value
-        else:
-            raise AttributeError('Pressure is not a conjugate variable.')
+        self._P = value
 
     @property
     def N(self):
+        """dict: The number of particles for each specified type."""
         return self._N.todict()
```
mphoward, line 134: Let's expose `self._N` directly.
```suggestion
        return self._N
```

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
 
     @N.setter
     def N(self, value):
         for t in value:
-            if 'N_{}'.format(t) in self.conjugates:
-                self._N[t] = value[t]
-            else:
-                raise AttributeError('Number is not a conjugate variable.')
+            self._N[t] = value[t]
```
mphoward, line 139: Let's remove the N setter. If the user wants to set the number of a type, they can do this by key using the getter like
```py
ens.N['A'] = 4
```

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
 
     @property
     def mu(self):
+        """dict: The chemical potential for each specified type."""
         return self._mu.todict()
```
mphoward, line 144: Same here
```suggestion
        return self._mu
```

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
 
     @mu.setter
     def mu(self, value):
         for t in value:
-            if 'mu_{}'.format(t) in self.conjugates:
-                self._mu[t] = value[t]
-            else:
-                raise AttributeError('Chemical potential is not a conjugate variable.')
+            self._mu[t] = value[t]
```
mphoward, line 149: Remove this.

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
 
     @property
     def conjugates(self):
+        """array_like: A list of the fluctuating variables conjugate to the
+        constant parameters of the ensemble."""
         return self._conjugates
 
     def reset(self):
```
mphoward, line 157: To make the intent of this method more obvious, let's rename it `clear` so that it **clears** the values of the conjugate variables of the ensemble.

relentless/ensemble.py
```diff
@@ -28,94 +73,96 @@ def __init__(self, types, T, P=None, V=None, mu={}, N={}, kB=1.0, conjugates=Non
...
-            self._conjugates = tuple(conjugates)
+        # conjugates are deduced from the specified parameters
+        if self.P is not None and self.V is not None:
+            raise ValueError('Both P and V cannot be set.')
+        for t in self.types:
+            if self.mu[t] is not None and self.N[t] is not None:
+                raise ValueError('Both mu and N cannot be set for type {}.'.format(t))
+
+        # build the set of conjugate variables from the constructor
+        conjugates = []
```
mphoward, line 96: I did this in a clunky way. Let's switch this to be a list of **constants** rather than conjugates, and make it a dict of bools, so that you can check what is True / False.
```py
self._constant = {
    'P': self.P is not None,
    'V': self.V is not None,
    'mu': {self.mu[t] is not None for t in types},
    'N': {self.N[t] is not None for t in types}
}
```
Then, expose this dict via a readonly `constant` property so that it can be used like:
```py
if ens.constant['P']:
    # const P
if ens.constant['mu']['A']:
    # GC-A
```
The user can technically still modify the contents of this dict, so we should warn them not to do that in the docs.

relentless/ensemble.py
```diff
@@ -125,15 +172,32 @@ def reset(self):
...
+        """Copies the ensemble and returns the copy with reset parameters.
+
+        Returns
+        -------
+        :py:class:`Ensemble`
+            A copy of the ensemble object with reset parameters.
+
+        """
         ens = copy.deepcopy(self)
         return ens.reset()
```
mphoward, line 189: I don't know why I wanted to reset this ensemble on copy, but it does not make sense to do that. The user can always do:
```py
ens2 = ens.copy().clear()
```
Let's remove the call to `reset` here.

relentless/ensemble.py
```diff
@@ -125,15 +172,32 @@ def reset(self):
...
 
     def save(self, basename='ensemble'):
```
mphoward, line 191: We should revisit if there is a better way to save this other than making multiple files. Just making a note for now.

mphoward (changes requested):
Great first pass! Here are some suggestions for enhancements. I will defer checking the unit tests until you have worked on some of these, as they will change the interface a bit.

mphoward:
I would like to propose that we change the `V` argument from being a true volume (i.e., a float) to a new class that defines a "volume" in the sense of a region of space. This can be used to setup different types of simulation boxes. Currently, we need to assume the box is a cube, but for certain design targets it might be important to choose a good box shape.

I would make an abstract base class `Volume` (or a `Region`), which defines the interface for computing its total volume:
```py
class Volume:
    @property
    @abc.abstractmethod
    def volume(self):
        """Volume of the region."""
```
Then we should implement the three common boxes that are used in LAMMPS / HOOMD:
```py
class Parallelepiped(Volume):
    """General triclinic box."""
    def __init__(self, a, b, c):
        self.a = np.asarray(a)
        self.b = np.asarray(b)
        self.c = np.asarray(c)
        self.matrix = np.column_stack((self.a,self.b,self.c))
        # check that self.matrix.shape == (3,3), or something went wrong

    @property
    def volume(self):
        """float: Volume computed using scalar triple product."""
        return np.linalg.norm(np.dot(np.cross(self.a,self.b),self.c))

class Cuboid(Parallelepiped):
    """Orthorhombic box."""
    def __init__(self, Lx, Ly, Lz):
        super().__init__([Lx,0,0],[0,Ly,0],[0,0,Lz])

class Cube(Cuboid):
    """Cubic box."""
    def __init__(self, L):
        super().__init__(L,L,L)
```

Usage would be like
```py
>>> V = Cuboid(1,2,3)
>>> print(V.volume)
6.0
```
Simulation engines can then check if they support simulation of ensemble against the list of volumes they work with:
```py
if not isinstance(ens.V, Parallelepiped):
    raise TypeError('HOOMD only simulates parallelpiped (triclinic) boxes')
```

relentless/ensemble.py
```diff
@@ -8,132 +12,273 @@
...
             self._P = None
         for t in self.types:
-            if 'N_{}'.format(t) in self.conjugates:
+            if not self.constant['N'][t]:
                 self._N[t] = None
-            if 'mu_{}'.format(t) in self.conjugates:
+            if not self.constant['mu'][t]:
                 self._mu[t] = None
-        self.rdf = core.PairMatrix(self.types)
+        self.rdf = core.PairMatrix(types=self.types)
```
mphoward, line 257: Just in case someone is holding a reference to the RDF pair matrix, let's just reset these. I think this should work:
```suggestion
        for pair in self.rdf:
            self.rdf[pair] = None
```

relentless/ensemble.py
```diff
@@ -8,132 +12,273 @@
...
 
         return self
 
     def copy(self):
-        ens = copy.deepcopy(self)
-        return ens.reset()
+        """Returns a copy of the specified ensemble.
```
mphoward, line 262:
```suggestion
        """Make a copy of the ensemble.
```

relentless/ensemble.py
mphoward: Noting here: as we discussed, we will dump all the data into one JSON file.

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+    def test_init(self):
+        """Test creation from data"""
+        #test valid construction
+        r = [1,2,3]
+        g = [2,9,5]
+        rdf = relentless.RDF(r=r, g=g)
+        np.testing.assert_allclose(rdf.table, np.array([[1,2],
+                                                        [2,9],
+                                                        [3,5]]))
```
mphoward, line 20: Can you add a quick test that if you call the RDF (as an Interpolator) you get the right value?

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        np.testing.assert_allclose(p.b, np.array([3,4,5]))
+        np.testing.assert_allclose(p.c, np.array([9,9,0]))
+        np.testing.assert_allclose(p.matrix, np.array([[1,3,9],
+                                                       [2,4,9],
+                                                       [1,5,0]]))
+        self.assertAlmostEqual(p.volume, 36)
+
+        #test invalid construction
+        with self.assertRaises(TypeError):
+            p = relentless.Parallelepiped(a=(1,2,1),b=(3,4,5),c=(9,9))
```
mphoward, line 49: Add a check that construction fails if the vectors are not righthanded, i.e., the volume would be negative.

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+    """Unit tests for relentless.Parallelepiped"""
+
+    def test_init(self):
+        """Test creation from data."""
+        #test valid construction
+        p = relentless.Parallelepiped(a=(1,2,1),b=(3,4,5),c=(9,9,0))
+        np.testing.assert_allclose(p.a, np.array([1,2,1]))
+        np.testing.assert_allclose(p.b, np.array([3,4,5]))
+        np.testing.assert_allclose(p.c, np.array([9,9,0]))
+        np.testing.assert_allclose(p.matrix, np.array([[1,3,9],
```
mphoward, line 42: Per comment above, you can remove `matrix` from all of these tests.

relentless/ensemble.py
```diff
@@ -8,132 +12,273 @@
...
+        containing elements as (`x`,`y`,`z`).
+
+    Raises
+    ------
+    TypeError
+        If a, b, and c are not all `3x1` arrays.
+
+    """
+    def __init__(self, a, b, c):
+        self.a = np.asarray(a)
```
mphoward, line 60: I think we should add `dtype=np.float64` to these so that we get double precision. It isn't strictly necessary, but lower precision could cause issues with the cross / dot products in really skewed boxes.

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        np.testing.assert_allclose(c.c, np.array([0,0,3]))
+        np.testing.assert_allclose(c.matrix, np.array([[3,0,0],
+                                                       [0,3,0],
+                                                       [0,0,3]]))
+        np.testing.assert_allclose(c.volume, 27)
+
+        #test invalid construction
+        with self.assertRaises(TypeError):
+            c = relentless.Cube(L=(3,4,5))
+
```
mphoward, line 88: Add a test that an error is raised if `L <= 0`. (The Cuboid will do the error handling, so you don't need to modify `Cube`.)

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        np.testing.assert_allclose(c.c, np.array([0,0,5]))
+        np.testing.assert_allclose(c.matrix, np.array([[3,0,0],
+                                                       [0,4,0],
+                                                       [0,0,5]]))
+        np.testing.assert_allclose(c.volume, 60)
+
+        #test invalid construction
+        with self.assertRaises(TypeError):
+            c = relentless.Cuboid(Lx=(3,4),Ly=4,Lz=5)
+
```
mphoward, line 69: Add a test / error check to the `Cuboid` that `Lx > 0`, `Ly >0`, and `Lz > 0`. This might get caught by the parallelepiped, but we can be more specific that "all lengths must be positive".

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        v_obj = relentless.Cube(L=3.0)
+
+        #V and N set
+        ens = relentless.Ensemble(T=20, V=v_obj, N={'A':1,'B':2})
+        self.assertCountEqual(ens.types, ('A','B'))
+        self.assertCountEqual(ens.rdf.pairs, (('A','B'),('A','A'),('B','B')))
+        self.assertAlmostEqual(ens.kB, 1.0)
+        self.assertAlmostEqual(ens.T, 20)
+        self.assertEqual(ens.P, None)
+        self.assertAlmostEqual(ens.V.volume, 27.0)
```
mphoward, line 119: Can you also check that `ens.V is v_obj`?

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        self.assertAlmostEqual(ens.beta, 0.05)
+
+        #mu and N used for one type each
+        ens = relentless.Ensemble(T=100, V=v_obj, mu={'A':0.1}, N={'B':2})
+        self.assertCountEqual(ens.types, ('A','B'))
+        self.assertCountEqual(ens.rdf.pairs, (('A','B'),('A','A'),('B','B')))
+        self.assertAlmostEqual(ens.kB, 1.0)
+        self.assertAlmostEqual(ens.T, 100)
+        self.assertEqual(ens.P, None)
+        self.assertAlmostEqual(ens.V.volume, 27.0)
```
mphoward, line 135: Check `ens.V` here too.

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        self.assertAlmostEqual(ens.beta, 0.01)
+
+        #test creation with single type
+        ens = relentless.Ensemble(T=100, V=v_obj, mu={'A':0.1})
+        self.assertCountEqual(ens.types, ('A',))
+        self.assertCountEqual(ens.rdf.pairs, (('A','A'),))
+        self.assertAlmostEqual(ens.kB, 1.0)
+        self.assertAlmostEqual(ens.T, 100)
+        self.assertEqual(ens.P, None)
+        self.assertAlmostEqual(ens.V.volume, 27.0)
```
mphoward, line 151: `ens.V`

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        self.assertDictEqual(ens.mu.todict(), {'A':0.1})
+        self.assertDictEqual(ens.N.todict(), {'A':None})
+        self.assertDictEqual(ens.constant, {'P':False,
+                                            'V':True,
+                                            'mu':{'A':True},
+                                            'N':{'A':False}})
+        self.assertAlmostEqual(ens.beta, 0.01)
```
mphoward, line 158: In one of these tests, could set Boltzmann's constant to something other than 1.0 to make sure that beta is using it properly? For example, its physical value is 1.38e-23 J/K but you can just use some other number like 2.0.

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+
+        #test setting rdf
+        ens = relentless.Ensemble(T=100, V=v_obj, mu={'A':0.1}, N={'B':2})
+        r = [1,2,3]
+        g_ab = [2,9,5]
+        g_aa = [3,7,4]
+        g_bb = [1,9,3]
```
mphoward, line 165: To compact the code, just put these in the spot where you are constructing these objects, since you don't use them again.

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        with self.assertRaises(ValueError):
+            ens = relentless.Ensemble(T=10, P=2.0)
+
+        #test invalid setting of both P and V
+        with self.assertRaises(ValueError):
+            ens = relentless.Ensemble(T=10, P=2.0, V=v_obj)
+
+        #test invalid setting of both mu and N
+        with self.assertRaises(ValueError):
+            ens = relentless.Ensemble(T=10, mu={'A':0.1,'B':0.2}, N={'A':1,'B':2})
```
mphoward, line 193: Can you add another test like this but where you do one of the pairs correctly and the other pair incorrectly?

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+
+        #test setting N as a float
+        with self.assertRaises(TypeError):
+            ens = relentless.Ensemble(T=10, P=1.0, N={'A':1.0})
+
+    def test_set_conjugates(self):
```
mphoward, line 199: Should these tests be renamed and condensed now that you're allowed to set anything? I would probably just make an NVT ensemble, and then try to set everything in turn.

Also, you might want to try to set `ens.N` and `ens.mu` directly here, and make sure you get the python error for when a property has no setter. (I forget what gets raised.)

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        ens.V = v_obj1
+        self.assertAlmostEqual(ens.V.volume, 64.0)
+        ens.mu['A'] = 0.2
+        ens.mu['B'] = 0.3
+        self.assertDictEqual(ens.mu.todict(), {'A':0.2,'B':0.3})
+        ens.N['A'] = 2
+        ens.N['B'] = 3
+        self.assertDictEqual(ens.N.todict(), {'A':2,'B':3})
+
+    def test_clear(self):
```
mphoward, line 258: This test, though, is good to use all these different combos since they will get cleared differently.

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        g_ab = [2,9,5]
+        g_aa = [3,7,4]
+        g_bb = [1,9,3]
+        ens.rdf['A','B'] = relentless.RDF(r=r, g=g_ab)
+        ens.rdf['A','A'] = relentless.RDF(r=r, g=g_aa)
+        ens.rdf['B','B'] = relentless.RDF(r=r, g=g_bb)
+        ens.clear()
+        self.assertDictEqual(ens.rdf['A','B'], {})
+        self.assertDictEqual(ens.rdf['A','A'], {})
+        self.assertDictEqual(ens.rdf['B','B'], {})
```
mphoward, line 307: Shouldn't these check if they are `None` rather than empty dict?

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        """Test copy method"""
+        v_obj = relentless.Cube(L=1.0)
+
+        #P and mu set
+        ens = relentless.Ensemble(T=10, P=2.0, mu={'A':0.1,'B':0.2})
+        ens.V = v_obj
+        ens.N['A'] = 1
+        ens.N['B'] = 2
+        ens_ = ens.copy()
+        assert ens_ is not ens
```
mphoward, line 319: You should use `self.assertIsNot` rather than raw `assert`

tests/test_ensemble.py
```diff
@@ -0,0 +1,381 @@
...
+        ens = relentless.Ensemble(T=100, V=v_obj, mu={'A':0.1}, N={'B':2})
+        r = [1,2,3]
+        g_ab = [2,9,5]
+        g_aa = [3,7,4]
+        g_bb = [1,9,3]
+        ens.rdf['A','B'] = relentless.RDF(r=r, g=g_ab)
+        ens.rdf['A','A'] = relentless.RDF(r=r, g=g_aa)
+        ens.rdf['B','B'] = relentless.RDF(r=r, g=g_bb)
+        ens_ = ens.copy()
+        assert ens_ is not ens
```
mphoward, line 371: You should additionally check that `ens.rdf is not ens_.rdf` to make sure the deep copy worked. (A shallow copy should fail this check.)

mphoward (changes requested):
Really excellent work! I have some suggestions / changes. The number is large because I tried to manually edit as much as possible to make it easier for you.

I will make one more general comment on the PR about an additional `Volume` I would like to add.

mphoward:
For convenience for users of LAMMPS & HOOMD, could we add one more type of `Volume` called `Triclinic`? A triclinic box is defined by 6 parameters: Lx, Ly, Lz, xy, xz, and yz. Lx, Ly, and Lz are the lengths of an orthorhombic box, and then xy, xz, and yz are "tilt factors" of that box. In LAMMPS, they are literally the amount (distance) that you skew the vectors. In HOOMD, they are fractional amounts relative to other lengths, but they basically do the same thing.

Here's one way you might implement it:

```py
class Triclinic(Parallelpiped):
    LAMMPS = 1
    HOOMD = 2

    def __init__(self, Lx, Ly, Lz, xy, xz, yz, convention=Triclinic.LAMMPS):
        if convention is Triclinic.LAMMPS:
            a = (Lx,0,0)
            b = (xy,Ly,0)
            c = (xz,yz,Lz)
        elif convention is Triclinic.HOOMD:
            a = (Lx,0,0)
            b = (xy*Ly,Ly,0)
            c = (xz*Lz,yz*Lz,Lz)
        else:
             raise ValueError('Triclinic convention must be Triclinic.LAMMPS or Triclinic.HOOMD')
        super().__init__(a,b,c)
```

You could also make an `Enum` for the conventions, up to you what you think looks best for documenting. For further reading, here are the pages that describe what a triclinic box is:

https://lammps.sandia.gov/doc/Howto_triclinic.html

https://hoomd-blue.readthedocs.io/en/stable/box.html

relentless/ensemble.py
```diff
@@ -1,176 +1,437 @@
...
+        filename : `str`
+            The name of the file to save data in.
+
+        """
+        data = []
+
+        # append thermo data
+        thermo = {'T': self.T,
+                  'P': self.P,
+                  'V': self.V,
```
mphoward, line 386: Ah, now we do have a problem (like we had with the Variables) because Volume is a class so we will need to add a method to serialize / deserialize it. Let's add abstract `to_json` instance method and `from_json` class method to Volume. The `to_json` method should return a dict containing whatever arguments are required to reconstruct it. The `from_json` method should take a dict, parse it and construct a new instance of the object, and return the object.

Then, when you are dumping the volume, I would save it as a two part dictionary:
```py
data['V'] = {'__name__' : self.V.__name__,
             'data': self.V.to_json()}
```
so you can use the name in the load. I think something like this would work for that:
```py
self.V = globals()[data['V']['__name__']].from_json(data['V']['data'])
```
You will need to implement these methods for each of the volume classes.

relentless/ensemble.py
```diff
@@ -1,176 +1,437 @@
...
+                  'mu':self.mu,
+                  'N': self.N,
+                  'kB': self.kB,
+                  'constant': self.constant} #need constant?
```
mphoward, line 390: Yes, definitely save `constant`.

relentless/ensemble.py
```diff
@@ -1,176 +1,437 @@
...
+        data.append(thermo)
+
+        # append rdf data
+        rdf = {pair:self.rdf[pair].table for pair in self.rdf}
```
mphoward, line 394: You will want to do this in a for loop:
```py
for pair in self.rdf:
    data['rdf'][pair] = self.rdf[pair].table
```
You should also add a check if `self.rdf[pair] is not None`, or else this access will fail. In that case, you can either record "None" or skip the pair. I think either is sensible.

relentless/ensemble.py
```diff
@@ -1,176 +1,437 @@
...
+        data.append(rdf)
+
+        # dump data to json file
+        with open('{}.json'.format(filename),'w') as f:
+            json.dump(data, f, indent=4)
 
     @classmethod
-    def load(self, basename='ensemble'):
-        with open('{}.json'.format(basename)) as f:
+    def load(self, filename):
```
mphoward, line 402: Let's rename this `from_file` so that it's clear that it is actually making a new class.

relentless/ensemble.py
```diff
@@ -1,176 +1,437 @@
...
-                       N=data['N'],
-                       mu=data['mu'],
-                       kB=data['kB'],
-                       conjugates=data['conjugates'])
+        thermo = data[0]
+        rdf = data[1]
+
+        #reset fluctuating variables?
+
+        ens = Ensemble(T=thermo['T'],
```
mphoward, line 425: This probably won't work because you might have both constant and fluctuating variables saved in the file. I think you will need to go through `constants` to figure out what the constructor arguments are and use those to make the ensemble, then fill in the fluctuating ones.

relentless/ensemble.py
```diff
@@ -1,176 +1,437 @@
...
+                       P=thermo['P'],
+                       V=thermo['V'],
+                       mu=thermo['mu'],
+                       N=thermo['N'],
+                       kB=thermo['kB'])
 
         for pair in ens.rdf:
```
mphoward, line 432: Again, will need to be careful about cases where some (or all) of the RDFs are not set. Also, you will likely need to unpack the RDF arrays yourself (they won't have a `table` attribute).

relentless/core/ensemble.py
```diff
@@ -0,0 +1,312 @@
...
+        - The temperature (``T``) is always a constant.
+        - Either the pressure (``P``) or the volume (``V``) can be specified.
+        - Either the chemical potential (``mu``) or the particle number (``N``)
+          can be specified for each type. The particle types in the Ensemble are
+          determined from the keys in the ``mu`` and ``N`` dictionaries.
+
+    The variables that are "constants" of the ensemble (e.g. *N*, *V*, and *T*
+    for the canonical ensemble) are determined from the arguments specified in the
+    constructor. The fluctuating (conjugate) variables must be set subsequently.
+    It is an error to set both variables in a conjugate pair (e.g. *P* and *V)
```
mphoward, line 43:
```suggestion
    It is an error to set both variables in a conjugate pair (e.g. *P* and *V*)
```

relentless/core/ensemble.py
```diff
@@ -0,0 +1,312 @@
...
+    Raises
+    ------
+    ValueError
+        If neither ``P`` nor ``V`` is set.
+    ValueError
+        If both ``P`` and ``V`` are set.
+    TypeError
+        If all values of ``N`` are not integers or None.
+    ValueError
+        If both ``mu`` and ``N`` are set for a type.
```
mphoward, line 70:
```suggestion
        If both ``mu`` and ``N`` are set for a type.
    ValueError
        If both ``mu`` and ``N`` are None.
```

relentless/core/ensemble.py
```diff
@@ -0,0 +1,312 @@
...
+        self._mu = FixedKeyDict(keys=types)
+        self._N = FixedKeyDict(keys=types)
+        self.mu.update(mu)
+        self.N.update(N)
+
+        # rdf per-pair
+        self._rdf = PairMatrix(types=types)
+
+        # build the set of constant variables from the constructor
+        self._constant = {'T':True,
```
mphoward, line 116:
```suggestion
        self._constant = {'T': True,
```

relentless/core/ensemble.py
```diff
@@ -0,0 +1,312 @@
...
+        self._P = value
+
+    @property
+    def N(self):
+        r""":py:class:`FixedKeyDict`: Number of particles of each type."""
+        return self._N
+
+    @property
+    def mu(self):
+        r""":py:class:`FixedKeyDict`: Chemical potential for each type."""
```
mphoward, line 164:
```suggestion
        r""":py:class:`FixedKeyDict`: Chemical potential of each type."""
```

relentless/core/ensemble.py
```diff
@@ -0,0 +1,312 @@
...
+        filename : str
+            The name of the file to save data in.
+
+        """
+        data = {'T': self.T,
+                'P': self.P,
+                'V': {'__name__':type(self.V).__name__,
+                      'data':self.V.to_json()
+                     },
+                'mu':self.mu.todict(),
```
mphoward, line 239:
```suggestion
                'mu': self.mu.todict(),
```

relentless/core/ensemble.py
```diff
@@ -0,0 +1,312 @@
...
+        with open(filename) as f:
+            data = json.load(f)
+
+        # retrieve thermodynamic parameters
+        thermo = {'T':data['T'],
+                  'P':data['P'],
+                  'V':globals()[data['V']['__name__']].from_json(data['V']['data']),
+                  'mu':data['mu'],
+                  'N':data['N']
+                 }
```
mphoward, line 281:
```suggestion
        thermo = {'T': data['T'],
                  'P': data['P'],
                  'V': globals()[data['V']['__name__']].from_json(data['V']['data']),
                  'mu': data['mu'],
                  'N': data['N']
                 }
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
+__all__ =  ['Volume', 'Cube','Cuboid','Parallelepiped','TriclinicBox']
```
mphoward, line 1:
```suggestion
__all__ =  ['Volume','Cube','Cuboid','Parallelepiped','TriclinicBox']
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+
+import abc
+from enum import Enum
+
+import numpy as np
+
+class Volume(abc.ABC):
+    r"""Abstract base class defining a region of space.
+
+    A Volume can be any region of space; typically, it is a simulation "box".
```
mphoward, line 11:
```suggestion
    A Volume can be any region of space; typically, it is a simulation "box."
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+    defined by an orthorhombic box oriented along the Cartesian axes and having
+    three vectors of length ``Lx``, ``Ly``, and ``Lz``, respectively. The box is
+    then tilted by factors ``xy``, ``xy``, and ``xz``, which are upper off-diagonal
+    elements of the matrix of box vectors. As a result, the **a** vector is
+    always aligned along the *x*-axis, while the other two vectors may be tilted.
+
+    The tilt factors can be defined using one of two :py:class:`TriclinicBox.Convention`.
+    In the `LAMMPS <https://lammps.sandia.gov/doc/Howto_triclinic.html>`_
+    simulation convention, specified using ``TriclinicBox.Convention.LAMMPS``,
+
```
mphoward, line 139:
```suggestion
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+        """
+        LAMMPS = 1
+        HOOMD = 2
+
+    def __init__(self, Lx, Ly, Lz, xy, xz, yz, convention=Convention.LAMMPS):
+        if Lx<=0 or Ly<=0 or Lz<= 0:
+            raise ValueError('All side lengths must be positive.')
+        self._convention = convention
+        if self._convention is TriclinicBox.Convention.LAMMPS:
```
mphoward, line 215:
```suggestion
        if self.convention is TriclinicBox.Convention.LAMMPS:
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+            a = (Lx,0,0)
+            b = (xy,Ly,0)
+            c = (xz,yz,Lz)
+        elif self._convention is TriclinicBox.Convention.HOOMD:
```
mphoward, line 219:
```suggestion
        elif self.convention is TriclinicBox.Convention.HOOMD:
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+            xy = self.b[0]/self.b[1]
+            xz = self.c[0]/self.c[2]
+            yz = self.c[1]/self.c[2]
+        return {'Lx':self.a[0],
+                'Ly':self.b[1],
+                'Lz':self.c[2],
+                'xy':xy,
+                'xz':xz,
+                'yz':yz,
+                'convention':self._convention.name
```
mphoward, line 259:
```suggestion
        return {'Lx': self.a[0],
                'Ly': self.b[1],
                'Lz': self.c[2],
                'xy': xy,
                'xz': xz,
                'yz': yz,
                'convention': self._convention.name
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+        Length along the *x* axis.
+    Ly : float
+        Length along the *y* axis.
+    Lz : float
+        Length along the *z* axis.
+
+    Raises
+    ------
+    ValueError
+        If *Lx*, *Ly*, *Lz* are not all positive.
```
mphoward, line 314:
```suggestion
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+
+        Returns
+        -------
+        dict
+            The serialized Cuboid.
+
+        """
+        return {'Lx':self.a[0],
+                'Ly':self.b[1],
+                'Lz':self.c[2],
```
mphoward, line 333:
```suggestion
        return {'Lx': self.a[0],
                'Ly': self.b[1],
                'Lz': self.c[2],
```

relentless/core/volume.py
```diff
@@ -0,0 +1,399 @@
...
+
+        The dictionary contains the box length ``L``.
+
+        Returns
+        -------
+        dict
+            The serialized Cube.
+
+        """
+        return {'L':self.a[0]}
```
mphoward, line 381:
```suggestion
        return {'L': self.a[0]}
```

tests/test_ensemble.py
```diff
@@ -0,0 +1,374 @@
...
+        self.assertDictEqual(ens.constant, {'T': True,
+                                            'P':True,
+                                            'V':False,
+                                            'mu':{'A':True,'B':True},
+                                            'N':{'A':False,'B':False}})
+        self.assertAlmostEqual(ens.beta, 0.1)
+
+        v_obj = relentless.Cube(L=3.0)
+
+        #V and N set, non-default value of kB
```
mphoward, line 59:
```suggestion
        #V and N set, non-default value of kB
        v_obj = relentless.Cube(L=3.0)
```

tests/test_ensemble.py
```diff
@@ -0,0 +1,374 @@
...
+        ens = relentless.Ensemble(T=20, V=v_obj, N={'A':1,'B':2}, kB=2.0)
+        self.assertCountEqual(ens.types, ('A','B'))
+        self.assertCountEqual(ens.rdf.pairs, (('A','B'),('A','A'),('B','B')))
+        self.assertAlmostEqual(ens.kB, 2.0)
+        self.assertAlmostEqual(ens.T, 20)
+        self.assertEqual(ens.P, None)
+        assert ens.V is v_obj
```
mphoward, line 66:
```suggestion
        self.assertIs(ens.V,v_obj)
```

tests/test_ensemble.py
```diff
@@ -0,0 +1,374 @@
...
+        self.assertAlmostEqual(ens.beta, 0.025)
+
+        #mu and N used for one type each
+        ens = relentless.Ensemble(T=100, V=v_obj, mu={'A':0.1}, N={'B':2})
+        self.assertCountEqual(ens.types, ('A','B'))
+        self.assertCountEqual(ens.rdf.pairs, (('A','B'),('A','A'),('B','B')))
+        self.assertAlmostEqual(ens.kB, 1.0)
+        self.assertAlmostEqual(ens.T, 100)
+        self.assertEqual(ens.P, None)
+        assert ens.V is v_obj
```
mphoward, line 84:
```suggestion
        self.assertIs(ens.V,v_obj)
```

tests/test_ensemble.py
```diff
@@ -0,0 +1,374 @@
...
+        self.assertAlmostEqual(ens.beta, 0.01)
+
+        #test creation with single type
+        ens = relentless.Ensemble(T=100, V=v_obj, mu={'A':0.1})
+        self.assertCountEqual(ens.types, ('A',))
+        self.assertCountEqual(ens.rdf.pairs, (('A','A'),))
+        self.assertAlmostEqual(ens.kB, 1.0)
+        self.assertAlmostEqual(ens.T, 100)
+        self.assertEqual(ens.P, None)
+        assert ens.V is v_obj
```
mphoward, line 102:
```suggestion
        self.assertIs(ens.V,v_obj)
```

tests/test_ensemble.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def test_set_params(self):
+        """Test setting constant and fluctuating parameter values."""
+        v_obj = relentless.Cube(L=3.0)
+        v_obj1 = relentless.Cube(L=4.0)
+
+        #NVT ensemble
+        ens = relentless.Ensemble(T=10, V=v_obj, N={'A':1,'B':2})
+        self.assertEqual(ens.P, None)
+        assert ens.V is v_obj
```
mphoward, line 159:
```suggestion
        self.assertIs(ens.V,v_obj)
```

tests/test_ensemble.py
```diff
@@ -0,0 +1,374 @@
...
+        self.assertAlmostEqual(ens.V.volume, 27.0)
+        self.assertDictEqual(ens.mu.todict(), {'A':None,'B':None})
+        self.assertDictEqual(ens.N.todict(), {'A':1,'B':2})
+
+        #set constant values
+        ens.V = v_obj1
+        assert ens.V is v_obj1
```
mphoward, line 166:
```suggestion
        self.assertIs(ens.V,v_obj1)
```

tests/test_ensemble.py
```diff
@@ -0,0 +1,374 @@
...
+        """Test copy method"""
+        v_obj = relentless.Cube(L=1.0)
+
+        #P and mu set
+        ens = relentless.Ensemble(T=10, P=2.0, mu={'A':0.1,'B':0.2})
+        ens.V = v_obj
+        ens.N['A'] = 1
+        ens.N['B'] = 2
+        ens_ = ens.copy()
+        self.assertIsNot(ens_, ens)
```
mphoward, line 243:
```suggestion
        self.assertIsNot(ens, ens_)
```

mphoward (changes requested):
Looks good to go! I am going to apply these cosmetic changes and merge.

mphoward (approved):

PR #38, opened 2020-09-18
Documentation, features, and unit tests for core.data

adisr04:
- Fix errors in `Directory` and `Project` classes
- Add unit tests in `test_data.py`


tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+
+        cwd = os.getcwd()
+
+        #test creation with existing path
+        d = relentless.data.Directory(f.name)
+        self.assertEqual(d.path, f.name)
+        self.assertEqual(d._start, None)
+        #enter and exit
+        with d:
+            self.assertEqual(d._start, cwd)
```
mphoward, line 23: When you are inside the context, you should also check that the current directory is right:
```py
self.assertEqual(os.getcwd(), f.name)
```

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertEqual(d._start, None)
+
+        #test creation with non-existent path (absolute)
+        foo = f.name + '/foo'
```
mphoward, line 27: For crossplatform compatibility, use join
```py
foo = os.path.join(f.name, 'foo')
```

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+        d1 = relentless.data.Directory(foo)
+        self.assertEqual(d1.path, foo)
+        self.assertEqual(d1._start, None)
+        #enter and exit
+        with d1:
+            self.assertEqual(d1._start, cwd)
```
mphoward, line 33: Again, check that `os.getcwd() == foo`.

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertEqual(d1._start, None)
+
+        #test creation with non-existent path (recursive)
+        foobar = foo + '/bar/foobar'
```
mphoward, line 37: Same here:
```suggestion
        foobar = os.path.join(foo,'bar','foobar')
```

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+        d2 = relentless.data.Directory(foobar)
+        self.assertEqual(d2.path, foobar)
+        self.assertEqual(d2._start, None)
+        #enter and exit
+        with d2:
+            self.assertEqual(d2._start, cwd)
```
mphoward, line 43: Check `os.getcwd()`

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+
+        #create nested directory structure
+        d = relentless.data.Directory(f.name)
+        d1 = d.directory('foo')
+        d2 = d.directory('bar')
+        d3 = d1.directory('bar/foobar')
+        self.assertEqual(d.path, f.name)
+        self.assertEqual(d1.path, f.name+'/foo')
+        self.assertEqual(d2.path, f.name+'/bar')
+        self.assertEqual(d3.path, f.name+'/foo/bar/foobar')
```
mphoward, line 89: Use `os.path.join` for all of these.

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def test_context(self):
+        """Test context methods for Directory."""
+        f = tempfile.TemporaryDirectory()
+
+        #create nested directory structure
+        d = relentless.data.Directory(f.name)
+        d1 = d.directory('foo')
+        d2 = d.directory('bar')
+        d3 = d1.directory('bar/foobar')
```
mphoward, line 85: To avoid needing to `os.path.join` these together, maybe create the directory like this instead:
```suggestion
        d3 = d2.directory('foobar')
```

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertEqual(d2.path, f.name+'/bar')
+        self.assertEqual(d3.path, f.name+'/foo/bar/foobar')
+
+        #test creating files
+        x = d.file('spam.txt')
+        x1 = d1.file('ham.txt')
+        x2 = d1.file('eggs.txt')
+        x3 = d3.file('baz.txt')
+        o = open(x, 'w')
+        o.close()
```
mphoward, line 97: You can touch the files in one line like this:
```suggestion
        open(x, 'w').close()
```

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+        o1 = open(x1, 'w')
+        o1.close()
+        o2 = open(x2, 'w')
+        o2.close()
+        o3 = open(x3, 'w')
+        o3.close()
+        self.assertEqual(os.path.abspath(x), f.name+'/spam.txt')
+        self.assertEqual(os.path.abspath(x1), f.name+'/foo/ham.txt')
+        self.assertEqual(os.path.abspath(x2), f.name+'/foo/eggs.txt')
+        self.assertEqual(os.path.abspath(x3), f.name+'/foo/bar/foobar/baz.txt')
```
mphoward, line 107: Use join here.

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+
+        #test clearing directory structure
+        #delete sub-directory
```
mphoward, line 110: These are good tests, just make sure to use join on the paths, or you can use the `path` property of the directories.

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+
+class test_Project(unittest.TestCase):
+    """Unit tests for core.data.Project."""
+    def test_init(self):
+        """Test basic creation of Project."""
+        f = tempfile.TemporaryDirectory()
+        os.chdir(f.name)
+
+        #unspecified workspace and scratch
+        p = relentless.data.Project()
```
mphoward, line 131: Let's also access `Project` however we do for the `Directory`.

Can you add a check on the type of `workspace` and `scratch`? e.g.
```py
self.assertIsInstance(p.workspace, relentless.Directory)
```

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+class test_Directory(unittest.TestCase):
+    """Unit tests for core.data.Directory."""
+    def test_init(self):
+        """Test basic creation and methods of Directory."""
+        f = tempfile.TemporaryDirectory()
+
+        cwd = os.getcwd()
+
+        #test creation with existing path
+        d = relentless.data.Directory(f.name)
```
mphoward, line 18: Is the import structure of `core` allowing this? I'm not sure if we really want this... as in our other tests, let's design this to test the way we want it to be used:
```py
relentless.Directory()
```
or
```py
relentless.core.Directory()
```
I will open an issue to look into how we do the `core` imports.

tests/core/test_data.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertCountEqual(os.listdir(f.name+'/foo'), ['bar','ham.txt','eggs.txt'])
+        self.assertCountEqual(os.listdir(f.name+'/foo/bar/foobar'), [])
+        #delete parent directory
+        self.assertCountEqual(os.listdir(f.name), ['foo','bar','spam.txt'])
+        d.clear()
+        self.assertCountEqual(os.listdir(f.name), [])
+
+        f.cleanup()
+
+class test_Project(unittest.TestCase):
```
mphoward, line 123: Remember to `join` all these paths in the tests.

mphoward (changes requested):
This looks really good! Most of my comments are related to making sure paths are referred to in a cross-platform way. There are a couple other small things, but I think this will be good to go after you make another pass through.

relentless/core/data.py
```diff
@@ -0,0 +1,213 @@
...
+        """Exit the directory context.
+
+        If possible, the working directory is reset to the path before entering
+        the context. The change is silently ignored if the original directory
+        no longer exists.
+
+        """
+        try:
+            if self._start[-1] is not os.getcwd():
+                os.chdir(self._start.pop())
```
mphoward, line 83: This will not pop the directory if you don't need to change, which means that the stack will get off by one.

I also ran a quick test on my system, and it costs 5x more to call `os.getcwd()` than it does to call `os.chdir()` to the cwd. So, I think there is no benefit to avoid calling `os.chdir`.

This should fix both:
```suggestion
            os.chdir(self._start.pop())
```

mphoward (approved):
Thank you! I just have one small item, which I will correct and merge.

PR #43, opened 2020-10-08
Add adapter class that can support multiple simulation backends

mphoward:
This adds an adapter class `SimulationOperationAdapter` that will translate a "generic" operation into a specific implementation for a `Simulation` backend. The translation is done at runtime by mapping the type of the simulation backend (stored in the `SimulationInstance`) and the name of the class into an equivalent operation. The translation is only done once, provided that the simulation backend does not change between calls.

The generic operations are imported into the `simulate` namespace for convenient use like: 

```
relentless.simulate.InitializeRandomly()
```


relentless/simulate/generic.py
```diff
@@ -0,0 +1,301 @@
...
+    """Initializes a simulation box and pair potentials from a file.
+
+    Parameters
+    ----------
+    filename : str
+        The file from which to read the system data.
+    options : kwargs
+        Options for file reading.
+    r_buff : float
+        Buffer width (defaults to 0.4).
```
mphoward, line 61: Should `r_buff` be caught in `**options`? lammps and hoomd both use it, but not sure if this is a "general" option. Let's discuss.

relentless/simulate/generic.py
```diff
@@ -0,0 +1,301 @@
...
+    Parameters
+    ----------
+    energy_tolerance : float
+        Energy convergence criterion.
+    force_tolerance : float
+        Force convergence criterion.
+    max_iterations : int
+        Maximum number of iterations to run the minimization.
+    dt : float
+        Maximum step size.
```
mphoward, line 94: This is likely going to be a `hoomd` specific option, probably should be caught in `**options`.

relentless/simulate/generic.py
```diff
@@ -0,0 +1,301 @@
...
+    """:py:class:`SimulationOperation` that removes a specified integration operation.
+
+    Parameters
+    ----------
+    add_op : :py:class:`SimulationOperation`
+        The addition/integration operation to be removed.
+
+    """
+    def __init__(self, add_op):
+        super().__init__(add_op)
```
mphoward, line 122: These are base classes (they user will not initialize directly), so they don't need to be included here.

tests/simulate/test_generic.py
```diff
@@ -0,0 +1,52 @@
...
+
+        #Invalid backend
+        with self.assertRaises(TypeError):
+            sim = relentless.simulate.Simulation(ops)
+            sim.run(self.ensemble, self.potentials, self.directory)
+
+        #Invalid operation (in valid backend)
+        with self.assertRaises(TypeError):
+            ops = relentless.simulate.hoomd.Initialize()
+            dilute = relentless.simulate.Dilute([ops])
```
mphoward, line 45: Let's use a "generic" op that is not implemented for the backend. In this case, you could use the `AddNPTIntegrator` because `Dilute` does not support it now. But, make sure you leave a very clear comment about it now so that if someone does implement it later, they will not be confused about why this test fails.

relentless/simulate/generic.py
```diff
@@ -0,0 +1,301 @@
...
+            try:
+                BackendOp = getattr(backend,op_name)
+            except AttributeError:
+                raise TypeError('{}.{}.{} operation not found.'.format(backend.__module__,backend.__name__,op_name))
+
+            self._op = BackendOp(*self.args,**self.kwargs)
+            self._backend = sim.backend
+
+        return self._op(sim)
+
```
mphoward, line 35: Making a note to myself: we should find a way to set (and forward) attributes... this is tricky because the backend `_op` is not constructed until `__call__`. If `__call__` has been executed once, no problem, but the more common case is that it has not. Probably we need to track what properties get set on the adapter class, then loop over them to forward them to the backend op in call when it gets created.

mphoward (commented):
Excellent work! Here are few small suggestions.

relentless/simulate/generic.py
```diff
@@ -0,0 +1,296 @@
+import importlib
+import inspect
+
+import numpy as np
+
+from . import simulate
+
+class SimulationOperationAdapter(simulate.SimulationOperation):
```
mphoward, line 8: Sorry to request this cosmetic change, but could you rename this class to `GenericOperation`? I think that will make its intent clearer.

relentless/simulate/generic.py
```diff
@@ -0,0 +1,296 @@
...
+    @classmethod
+    def add_backend(cls, backend, module=None):
+        """Adds backend attribute to :py:class:`SimulationOperationAdapter`.
+
+        Parameters
+        ----------
+        backend : :py:class:`Simulation`
+            Class to add as a backend.
+        module : ??
+            Name of the module in which the backend is defined (defaults to `None`).
```
mphoward, line 56:
```suggestion
        module : module or str or None
            Module in which the backend is defined. If `None` (default),
            try to deduce the module from ``backend.__module__``.
            ``module`` will be imported if it has not already been.
```

relentless/simulate/generic.py
```diff
@@ -0,0 +1,296 @@
...
+    """Initializes a simulation box and pair potentials from a file.
+
+    Parameters
+    ----------
+    filename : str
+        The file from which to read the system data.
+    options : kwargs
+        Options for file reading.
+    neighbor_buffer : float
+        Buffer width for neighbor list (defaults to 0.4).
```
mphoward, line 80: Please reorder these. Also, let's remove the default value of `neighbor_buffer`. The user should be explicit about this choice. You should also remove this default in the `hoomd` module.

relentless/simulate/generic.py
```diff
@@ -0,0 +1,296 @@
...
+
+class InitializeRandomly(SimulationOperationAdapter):
+    """Initializes a randomly generated simulation box and pair potentials.
+
+    Parameters
+    ----------
+    seed : int
+        The seed to randomly initialize the particle locations (defaults to `None`).
+    neighbor_buffer : float
+        Buffer width for neighbor list (defaults to 0.4).
```
mphoward, line 94: Same here, remove default.

relentless/simulate/generic.py
```diff
@@ -0,0 +1,296 @@
...
+
+    """
+    def __init__(self, seed=None, neighbor_buffer=0.4):
+        super().__init__(seed, neighbor_buffer)
+
+## integrators
+class MinimizeEnergy(SimulationOperationAdapter):
+    """:py:class:`SimulationOperation` that runs a FIRE energy minimzation until converged.
```
mphoward, line 102:
```suggestion
    """:py:class:`SimulationOperation` that runs an energy minimization until converged.
```

relentless/simulate/generic.py
```diff
@@ -0,0 +1,296 @@
...
+
+    Parameters
+    ----------
+    seed : int
+        The seed to randomly initialize the particle locations (defaults to `None`).
+    neighbor_buffer : float
+        Buffer width for neighbor list (defaults to 0.4).
+
+    """
+    def __init__(self, seed=None, neighbor_buffer=0.4):
```
mphoward, line 97: Should we add an `**options` argument here to catch any extras?

relentless/simulate/simulate.py
```diff
@@ -74,13 +74,19 @@ def operations(self, ops):
...
+                                  potentials,
+                                  directory,
+                                  **self.options)
 
 class SimulationInstance:
     """Specific instance of a simulation and its data.
 
     Parameters
     ----------
+    backend : :py:class:`Simulation`
```
mphoward, line 88:
```suggestion
    backend : type
```

mphoward (commented):
Looks great. I just have a few additional comments. Once you address these, I'll run the tests again and merge.

adisr04 (approved):

mphoward:
Thank you! We can handle any additional cleanup (if I missed anything) in a subsequent PR.

PR #44, opened 2020-10-11
Refactor potentials

mphoward:
This is a proposed refactoring of the potentials with two goals:
1. Simplify the `Simulation` to accept a `ForceField` rather than a pretabulated set of potentials.
2. Open the possibility for other types of potentials later down the road (e.g., bond potentials).

Changes:
* Make `PairPotential` and `PairParameters` special instances of a base `Potential` and `Parameters`.
* Refactor `Tabulator` as `PotentialTabulator`. Simplify to constant width tables. `PairPotentialTabulator` now **always** shifts the potential, and it optionally truncates the force (but does modify the potential).
* Create `ForceField` class to wrap multiple `PotentialTabulator`s for use in `Simulation`.

The unit tests need to be refactored to match the class structure. We should try to test the base classes (if possible), and reorg the tests so that the pair potentials / parameters are tested in `test_pair.py`.

Note: I switched `Parameters` so that it does not derive from `PairMatrix`. I'm not sure if there was a reason we did this before or if it was a vestigial thing, but if there are any methods that are defined in `PairMatrix` that we are missing, we can redefine them on this base class.


relentless/potential/potential.py
```diff
@@ -1,170 +1,45 @@
...
-        >>> m.shared = {'energy': 0.5}
-        >>> print(m['B','B'])
-        {'energy': None, 'mass': 0.1}
-        >>> print(m.shared)
-        {'energy': 0.5, 'mass': None}
-        >>> print(m.evaluate(('B','B'))
-        {'energy':0.5, 'mass':0.1}
-
-    """
+class Parameters:
```
mphoward, line 8: This class needs documentation and possibly some testing.

relentless/potential/potential.py
```diff
@@ -208,510 +83,91 @@ def design_variables(self):
...
         """:py:class:`FixedKeyDict`: The shared parameters."""
         return self._shared
 
-class PairPotential(abc.ABC):
-    """Generic pair potential evaluator.
-
-    A PairPotential object is created with coefficients as a PairParameters object.
-    This abstract base class can be extended in order to evaluate custom force,
-    energy, and derivative/gradient functions.
+class Potential(abc.ABC):
```
mphoward, line 116: This class needs documentation.

relentless/simulate/dilute.py
```diff
@@ -86,13 +86,10 @@ def __call__(self, sim):
...
 
         # pair distribution function
-        for pair in sim.potentials:
-            r = sim.potentials[pair].get('r')
-            u = sim.potentials[pair].get('u')
-            if r is None or u is None:
-                raise ValueError('r and u must be set in the potentials matrix.')
-            gr = np.exp(-sim.ensemble.beta*u)
-            ens.rdf[pair] = RDF(r,gr)
+        for pair in ens.rdf:
```
mphoward, line 89: I tweaked the `Ensemble` object to give it a `pairs` property as well:
```suggestion
        for pair in ens.pairs:
```

relentless/simulate/simulate.py
```diff
@@ -109,3 +112,184 @@ class Data:
     @abc.abstractmethod
     def __call__(self, sim):
         pass
+
+class Potentials:
+    """Combination of multiple potentials.
+    """
+    def __init__(self):
+        self._pair = PairPotentialTabulator(None,None)
```
mphoward, line 120: It might be good to give a convenience argument to the constructor `pair_potentials`:
```suggestion
    def __init__(self, pair_potentials=None):
        self._pair = PairPotentialTabulator(None,None,potentials=pair_potentials)
```

relentless/simulate/simulate.py
```diff
@@ -109,3 +112,184 @@ class Data:
...
+
+    @property
+    def pair(self):
+        return self._pair
+
+class PotentialTabulator:
```
mphoward, line 126: This needs docs. Testing is a little tricky because we don't have potentials yet that work with this.

relentless/simulate/simulate.py
```diff
@@ -109,3 +112,184 @@ class Data:
...
+        """
+        d = np.zeros_like(self.x)
+        for pot in self.potentials:
+            try:
+                d += pot.derivative(key,var,self.x)
+            except KeyError:
+                pass
+        return d
+
+class PairPotentialTabulator(PotentialTabulator):
```
mphoward, line 251: This needs docs and testing.

mphoward (commented):
@adisr04 I've marked the spots that are most in need of attention.

relentless/potential/potential.py
```diff
@@ -1,170 +1,88 @@
...
+    per-type, or shared between all parameters. The per-type parameters take precedence
+    over the shared parameters.
 
     Parameters
     ----------
     types : array_like
         List of types (A type must be a `str`).
     params : array_like
-        List of parameters (A parameter must be a `str`).
+        List of types (A parameter must be a `str`).
```
mphoward, line 20: I think this change is accidental and should be reverted?

relentless/potential/potential.py
```diff
@@ -208,510 +126,105 @@ def design_variables(self):
...
 
     @property
     def shared(self):
         """:py:class:`FixedKeyDict`: The shared parameters."""
         return self._shared
 
-class PairPotential(abc.ABC):
-    """Generic pair potential evaluator.
+class Potential(abc.ABC):
+    """Abstract base class for potential coefficients.
```
mphoward, line 160:
```suggestion
    """Abstract base class for interaction potential.
```

relentless/simulate/simulate.py
```diff
@@ -34,8 +37,8 @@ def run(self, ensemble, potentials, directory):
         ensemble : :py:class:`Ensemble`
             Simulation ensemble. Must include values for *N* and *V* even if
             these variables fluctuate.
-        potentials : :py:class:`PairMatrix`
-            Matrix of tabulated potentials for each pair.
+        potentials : :py:class:`Potentials`
+            The tabulated potentials.
```
mphoward, line 41:
```suggestion
            The interaction potentials.
```
(we don't "tabulate" explicitly anymore)

relentless/simulate/simulate.py
```diff
@@ -81,8 +84,8 @@ class SimulationInstance:
     ensemble : :py:class:`Ensemble`
         Simulation ensemble. Must include values for *N* and *V* even if
         these variables fluctuate.
-    potentials : :py:class:`PairMatrix`
-        Matrix of tabulated potentials for each pair.
+    potentials : :py:class:`Potentials`
+        The tabulated potentials.
```
mphoward, line 88:
```suggestion
        The interaction potentials.
```

relentless/simulate/simulate.py
```diff
@@ -109,3 +112,288 @@ class Data:
     @abc.abstractmethod
     def __call__(self, sim):
         pass
+
+class Potentials:
+    """Combination of multiple potentials.
+
+    Iniitializes a :py:class:`PairPotentialTabulator` object that can store multiple potentials.
```
mphoward, line 119: Can you also tell the user that they will need to set `rmax` and `num` (I think that's the right names) for `pair` before the `Potentials` object can be used (if `pair is not None`)?

relentless/simulate/simulate.py
```diff
@@ -109,3 +112,288 @@ class Data:
...
+
+        """
+        u = super().energy(pair)
+        u -= u[-1]
+        return u
+
+    def force(self, pair):
+        """Evaluates and accumulates force for all potentials.
+
+        All forces greater than or equal to fmax are set to 0.
```
mphoward, line 361:
```suggestion
        If set, all forces are truncated to be less than or equal to ``|fmax|``.
```

mphoward (commented):
This is really great work. I have only a few small requests, and then I would like to merge this.

adisr04 (approved):

PR #45, opened 2020-10-14
HOOMD simulation framework

adisr04:
- Create framework for running simulations in HOOMD using existing `Simulation` infrastructure
- Add simulation operations: initialization, energy minimization, integrators, run, and callbacks
- Document and add unit tests for `hoomd.py`


relentless/simulate/hoomd.py
```diff
@@ -0,0 +1,784 @@
...
+            Simulation object.
+
+        Raises
+        ------
+        ValueError
+            If the length of the *r*, *u*, and *f* arrays are not all equal.
+
+        """
+        # first write all potentials to disk
+        os.chdir(sim.directory.path)
```
mphoward, line 164: There is no need to call this explicitly. You can either do:
```py
with sim.directory:
   # in the dir
```
or use `Directory.file()` will return the _absolute_ path so that you don't need to change directory at all.

In this case, we don't need this line at all, so remove.

relentless/simulate/hoomd.py
```diff
@@ -0,0 +1,784 @@
...
+        table_size = None
+        files = PairMatrix(sim.ensemble.types)
+        for i,j in sim.potentials:
+            r = sim.potentials[i,j].get('r')
+            u = sim.potentials[i,j].get('u')
+            f = sim.potentials[i,j].get('f')
+            if f is None:
+                ur = Interpolator(r,u)
+                f  = -ur.derivative(r,1)
```
mphoward, line 173: Let's require that the forces be set (remove this stuff). With the revised `Potentials` structure, the force will always be available, so don't worry about error checking this carefully.

relentless/simulate/hoomd.py
```diff
@@ -0,0 +1,784 @@
...
+            if len(r) != table_size or len(u) != table_size or len(f) != table_size:
+                raise ValueError('HOOMD requires equal sized tables.')
+
+            files[i,j] = sim.directory.file('table_{i}_{j}.dat'.format(i=i,j=j))
+            header = '# Tabulated pair for ({i},{j})\n'.format(i=i,j=j)
+            header += '# r u f'
+            np.savetxt(files[i,j],
+                       np.column_stack((r,u,f)),
+                       header=header,
+                       comments='')
```
mphoward, line 187: We are trying to avoid files completely, so don't write this data to file.

relentless/simulate/hoomd.py
```diff
@@ -0,0 +1,784 @@
...
+            for line in n.readlines():
+                line = line.strip()
+                if line[0] == '#':
+                    continue
+                cols = line.split()
+                values = [float(i) for i in cols]
+                r.append(values[0])
+                u.append(values[1])
+                f.append(values[2])
+        return (r, u, f)
```
mphoward, line 216: remove this and use the arrays you already have in memory

relentless/simulate/hoomd.py
```diff
@@ -0,0 +1,784 @@
...
+    filename : str
+        The file from which to read the system data.
+    options : kwargs
+        Options for file reading (as used in :py:func:`hoomd.init.read_gsd()`).
+
+    """
+    def __init__(self, filename, **options):
+        self.filename = os.path.realpath(filename)
+        self.options = options
+        super().__init__()
```
mphoward, line 240: Call this super method *first*, and also add the `r_buff` parameter I forgot.

relentless/simulate/hoomd.py
```diff
@@ -0,0 +1,784 @@
...
+
+    Parameters
+    ----------
+    seed : int
+        The seed to randomly initialize the particle locations (defaults to `None`).
+
+    """
+    def __init__(self, seed=None):
+        self.seed = seed
+        super().__init__()
```
mphoward, line 272: Same here as above.

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,109 @@
...
+        os.chdir(self.directory.path)
+        f = gsd.hoomd.open(name='test.gsd', mode='wb')
+        s = gsd.hoomd.Snapshot()
+        s.particles.N = 5
+        s.particles.types = ['A','B']
+        s.particles.typeid = [0,0,1,1,1]
+        s.particles.position = np.random.random(size=(5,3))
+        s.configuration.box = [2,2,2,0,0,0]
+        f.append(s)
+        f.close()
```
mphoward, line 30: Don't need this if you open `f` as a context

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,109 @@
...
+    """Unit tests for relentless.HOOMD"""
+
+    def setUp(self):
+        self._tmp = tempfile.TemporaryDirectory()
+        self.directory = relentless.Directory(self._tmp.name)
+
+    #mock gsd file for testing
+    def create_gsd(self):
+        os.chdir(self.directory.path)
+        f = gsd.hoomd.open(name='test.gsd', mode='wb')
```
mphoward, line 22:
```suggestion
        with gsd.hoomd.open(name=self.directory.file('test.gsd'), mode='wb') as f:
```

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,109 @@
...
+            pot[pair]['r'] = np.array([1.,2.,3.])
+            pot[pair]['u'] = np.array([2.,4.,6.])
+
+        #InitializeFromFile
+        f = self.create_gsd()
+        op = relentless.simulate.hoomd.InitializeFromFile(filename=f.file.name)
+        h = relentless.simulate.hoomd.HOOMD(operations=op)
+        sim = h._new_instance(ens, pot, self.directory)
+        op(sim)
+        self.assertIsInstance(sim._opdata[op].neighbor_list, hoomd.md.nlist.tree)
```
mphoward, line 47: These are supposed to accessible like:
```suggestion
        self.assertIsInstance(sim[op].neighbor_list, hoomd.md.nlist.tree)
```

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,109 @@
...
+        pot = relentless.PairMatrix(types=ens.types)
+        for pair in pot:
+            pot[pair]['r'] = np.array([1.,2.,3.])
+            pot[pair]['u'] = np.array([2.,4.,6.])
+
+        #InitializeFromFile
+        f = self.create_gsd()
+        op = relentless.simulate.hoomd.InitializeFromFile(filename=f.file.name)
+        h = relentless.simulate.hoomd.HOOMD(operations=op)
+        sim = h._new_instance(ens, pot, self.directory)
```
mphoward, line 45: It is probably better to just use `run`, rather than use a protected method. `sim` will be returned by `run`.

mphoward (changes requested):
Great start! Here's some suggestions on two of the more major parts. Let's work to get this filled in as much as possible so that we can merge before it gets too big.

tests/simulate/test_dilute.py
```diff
@@ -0,0 +1,152 @@
...
+        ens_ = analyzer.extract_ensemble(sim)
+        self.assertAlmostEqual(ens_.T, 3.0)
+        self.assertAlmostEqual(ens_.V.volume, 64.0)
+        self.assertAlmostEqual(ens_.P, 0.0302839)
+        self.assertDictEqual(ens_.N.todict(), {'A':3,'B':4})
+        for pair in ens_.rdf:
+            np.testing.assert_allclose(ens_.rdf[pair].table,
+                                       np.array([[2.,np.exp(-6./3.)],
+                                                 [4.,np.exp(-12./3.)],
+                                                 [6.,np.exp(-18./3.)]]))
```
mphoward, line 115: Tests like this can be removed if you move the ops above to `test_simulate.py`. Just check that the output is correct for one or two values of the ensemble.

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,183 @@
...
+class test_HOOMD(unittest.TestCase):
+    """Unit tests for relentless.HOOMD"""
+
+    def setUp(self):
+        self._tmp = tempfile.TemporaryDirectory()
+        self.directory = relentless.Directory(self._tmp.name)
+
+    #mock (NVT) ensemble and potential for testing
+    def ens_pot(self):
+        ens = relentless.Ensemble(T=100.0, V=relentless.Cube(L=10.0), N={'A':2,'B':3})
```
mphoward, line 20: If you are going to call `run` anywhere, I would suggest making T order 1 or else things will likely blow up.
```suggestion
        ens = relentless.Ensemble(T=2.0, V=relentless.Cube(L=10.0), N={'A':2,'B':3})
```

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,183 @@
...
+        return (ens,pot)
+
+    #mock gsd file for testing
+    def create_gsd(self):
+        with gsd.hoomd.open(name=self.directory.file('test.gsd'), mode='wb') as f:
+            s = gsd.hoomd.Snapshot()
+            s.particles.N = 5
+            s.particles.types = ['A','B']
+            s.particles.typeid = [0,0,1,1,1]
+            s.particles.position = np.random.random(size=(5,3))
```
mphoward, line 35: You should probably draw these uniformly in the box. Default random numbers run [0,1), so you have a high chance of particles overlapping in the initial config. Your potential is attractive, so that's probably OK, but just something to watch out for.
```suggestion
            s.particles.position = np.random.uniform(low=-5.0,high=5.0,size=(5,3))
```

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def test_minimization(self):
+        """Test running energy minimization simulation operation."""
+        #MinimizeEnergy
+        ens,pot = self.ens_pot()
+        op = [relentless.simulate.hoomd.InitializeRandomly(seed=1),
+              relentless.simulate.hoomd.MinimizeEnergy(energy_tolerance=1e-7,
+                                                       force_tolerance=1e-7,
+                                                       max_iterations=1000,
+                                                       dt=10)
```
mphoward, line 67: I'm not sure if this matters, but usually this is a smaller number like:
```suggestion
                                                       dt=0.01)
```

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,183 @@
...
+        init = relentless.simulate.hoomd.InitializeRandomly(seed=1)
+
+        #BrownianIntegrator
+        ens,pot = self.ens_pot()
+        brn = relentless.simulate.hoomd.AddBrownianIntegrator(dt=0.5,
+                                                              friction=1.0,
+                                                              seed=2)
+        brn_r = relentless.simulate.hoomd.RemoveBrownianIntegrator(add_op=brn)
+        op = [init,brn]
+        h = relentless.simulate.hoomd.HOOMD(operations=op)
```
mphoward, line 83: You can just create one HOOMD simulation instance (at the start of this method), and then you can set the operations as a list:
```py
h.operations = [init,brn]
```
Same goes for the other test functions

tests/simulate/test_hoomd.py
```diff
@@ -0,0 +1,183 @@
...
+
+        #extract ensemble
+        ens_ = analyzer.extract_ensemble(sim)
+        self.assertIsNotNone(ens_.T)
+        self.assertNotEqual(ens_.T, 0)
+        self.assertIsNotNone(ens_.P)
+        self.assertNotEqual(ens_.P, 0)
+        self.assertIsNotNone(ens_.V)
+        self.assertNotEqual(ens_.V.volume, 0)
+        self.assertEqual(thermo.num_samples, 100)
```
mphoward, line 170: Can you check on the RDF that is has some values in it?

tests/simulate/test_dilute.py
```diff
@@ -0,0 +1,152 @@
...
+
+        #no operations, with options
+        d = relentless.simulate.Dilute(**options)
+        self.assertCountEqual(d.operations, [])
+        self.assertDictEqual(d.options, options)
+
+        #with operations, with options
+        d = relentless.simulate.Dilute(operations, **options)
+        self.assertCountEqual(d.operations, operations)
+        self.assertDictEqual(d.options, options)
```
mphoward, line 65: Most of these tests can just go in `test_simulate.py`. Maybe just do one to make sure you can setup a Dilute simulation with one or two of the `dilute` operations.

tests/simulate/test_simulate.py
```diff
@@ -1,13 +1,83 @@
...
 
 import numpy as np
 
 import relentless
 
+class test_Simulation(unittest.TestCase):
+    """Unit tests for relentless.Simulation"""
+
+    #mock function for use as operation
+    class OperationA(relentless.simulate.SimulationOperation):
```
mphoward, line 13: I would replace these with two ops, like you have in `test_dilute.py`, that use the simple attribute setting I suggested there.

mphoward (changes requested):
This looks really good! I have a couple small suggestions, and then I think we are good to go.

mphoward (approved):
Excellent, thank you!

PR #48, opened 2020-11-10
LAMMPS simulation framework

adisr04:
- Create framework for running simulations in LAMMPS using existing `Simulation` infrastructure
- Add simulation operations: initialization, energy minimization, integrators, run, and callbacks
- Document and add unit tests for `lammps.py`


relentless/simulate/lammps.py
```diff
@@ -0,0 +1,120 @@
+import abc
+import os
+from packaging import version
```
mphoward, line 3: I think this import is unused

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _new_instance(self, ensemble, potentials, directory):
+        sim = super()._new_instance(ensemble,potentials,directory)
+
+        # create lammps instance with all output disabled
+        #quiet_launch = ['-echo','none',
+        #                '-log','none',
+        #                '-screen','none',
+        #                '-nocite']
+        sim.lammps = lammps.lammps()#cmdargs=quiet_launch)
```
mphoward, line 31: We will want to uncomment these for production, but your testing issues point out that it may not always be good to run LAMMPS in "quiet" mode. Maybe we should consider adding an argument to the constructor that when true (default) uses these arguments, and maybe when false allows only output to console? The argument could be called `quiet` if it works like this or `verbose` if you negate the behavior?

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,120 @@
...
+        lo = -0.5*np.array([Lx,Ly,Lz])
+        hi = lo + V.a + V.b + V.c
+        xlo = lo[0]
+        xhi = hi[0]
+        ylo = lo[1]
+        yhi = hi[1]
+        zlo = lo[2]
+        zhi = hi[2]
+
+        return np.array([xlo,xhi,ylo,yhi,zlo,zhi,xy,xz,yz])
```
mphoward, line 79:
```suggestion
        return np.array([lo[0],hi[0],lo[1],hi[1],lo[2],hi[2],xy,xz,yz])
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,120 @@
...
+            cmds += ["region box prism {} {} {} {} {} {} {} {} {}".format(*box)]
+        else:
+            cmds += ["region box block {} {} {} {} {} {}".format(*box[:-3])]
+        cmds += ["create_box {} box".format(len(sim.ensemble.types))]
+
+        # use lammps random initialization routines
+        for i in sim.ensemble.types:
+            cmds += ["create_atoms {typeid} random {N} {seed} box".format(typeid=sim.type_map[i],
+                                                                          N=sim.ensemble.N[i],
+                                                                          seed=self.seed+sim.type_map[i])]
```
mphoward, line 114: This is a pedantic change, but it makes sure that `seed` is actually used for the first type (rather than `seed+1`).
```suggestion
                                                                          seed=self.seed+sim.type_map[i]-1)]
```

mphoward (changes requested):
This is a good start!

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+        super().__init__(neighbor_buffer)
+        self.filename = filename
+        self.units = units
+        self.atom_style = atom_style
+
+    def to_commands(self, sim):
+        cmds = ['units {style}'.format(style=self.units),
+                'boundary p p p',
+                'atom_style {style}'.format(style=self.atom_style),
+                'read_data {filename}'.format(filename=filename), #TODO: keyword args?
```
mphoward, line 146: I don't think we need any of the extra args here.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+        cmds += ['pair_style table linear {N}'.format(N=Nr)]
+        for i,j in sim.ensemble.pairs:
+            # get lammps type indexes, lowest type first
+            id_i,id_j = pair_map(sim,(i,j))
+            cmds += ['pair_coeff {id_i} {id_j} {filename} TABLE_{id_i}_{id_j}'.format(id_i=id_i,id_j=id_j,filename=file_)]
+
+        return cmds
+
+class InitializeFromFile(Initialize):
+    def __init__(self, filename, neighbor_buffer, units='lj', atom_style='atomic'):
```
mphoward, line 136: Since `units` and `atom_style` are used by both `InitializeFromFile` and `InitializeRandomly`, you should move them into the base `Initialize` to reduce duplication.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+        self.filename = filename
+        self.units = units
+        self.atom_style = atom_style
+
+    def to_commands(self, sim):
+        cmds = ['units {style}'.format(style=self.units),
+                'boundary p p p',
+                'atom_style {style}'.format(style=self.atom_style),
+                'read_data {filename}'.format(filename=filename), #TODO: keyword args?
+                self.attach_potentials(sim)]
```
mphoward, line 147: This is accidentally making a list of lists, it should be like:
```suggestion
                ]
        cmds += self.attach_potentials(sim)
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+    def __init__(self, filename, neighbor_buffer, units='lj', atom_style='atomic'):
+        super().__init__(neighbor_buffer)
+        self.filename = filename
+        self.units = units
+        self.atom_style = atom_style
+
+    def to_commands(self, sim):
+        cmds = ['units {style}'.format(style=self.units),
+                'boundary p p p',
+                'atom_style {style}'.format(style=self.atom_style),
```
mphoward, line 145: Similarly, you could have these first three commands returned by the `to_commands` method of `Initialize`, then use it here and in the random initialization method:
```py
cmds = super().to_commands(sim)
cmds += ['read_data ...']
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+
+        return cmds
+
+class MinimizeEnergy(simulate.SimulationOperation):
+    def __init__(self, energy_tolerance, force_tolerance, max_iterations):
+        self.energy_tolerance = energy_tolerance
+        self.force_tolerance = force_tolerance
+        self.max_iterations = max_iterations
+
+    def to_commands(self, sim): #TODO: min_style, min_modfy?
```
mphoward, line 190: hoomd uses the FIRE energy minimization method, but lammps uses conjugate gradient by default. We could force these to methods to work in a similar way, and give the analogous `dt` parameter... not sure about this though.

For now, let's leave it be, and I will open an issue to come back and consider this later.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+        cmds += ['minimize {etol} {ftol} {maxiter}'.format(etol=self.energy_tolerance,
+                                                           ftol=self.force_tolerance,
+                                                           maxiter=self.max_iterations)]
+
+        return cmds
+
+# Brownian dynamics not supported by LAMMPS
+
+class AddLangevinIntegrator(simulate.SimulationOperation):
```
mphoward, line 199: This should derive from `LAMMPSOperation`.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+    def __init__(self, idx, group_idx, t_start, t_stop, damp, seed): #TODO: condense keywords, reconcile with generic?
```
mphoward, line 200: Yes, this (and other options) needs to work with a similar interface to generic & hoomd. A few things that need adapting:

* `idx` is the fix index. We should add a method to `LAMMPSOperation` for getting a new fix index automatically, since the user doesn't need to know anything about it. For example, you can make a class variable `fix_id` that is initially assigned a value of 1. Then, add a class method or instance method `add_fix` that returns the current value of `fix_id` and increments it by 1. Then, you can use & save this ID for the fix in the operation.
* `group_idx` should be `all` (see `hoomd.group.all()`).
* `t_start` and `t_stop` should both be set to `ens.T`.
* `damp` should be renamed `friction`, and it should be given the same meaning as in `hoomd` (which is the inverse of the definition in LAMMPS). Let's discuss this at our next meeting.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+        self.idx = idx
+        self.group_idx = group_idx
+        self.t_start = t_start
+        self.t_stop = t_stop
+        self.damp = damp
+        self.seed = seed
+
+    def to_commands(self, sim):
+        cmds += ['fix {idx} {group_idx} langevin {t_start} {t_stop} {damp} {seed}'.format(idx=self.idx,
```
mphoward, line 209: This needs another fix command that creates an NVE integrator for `all`, as on its own fix langevin is not an integrator. Also, if we want to support different friction coefficients for each type (as in hoomd), we would need to make a separate `fix langevin` for each type, and make it work on each type group individually.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+                                                                                          group_idx=self.group_idx,
+                                                                                          t_start=self.t_start,
+                                                                                          t_stop=self.t_stop,
+                                                                                          damp=self.damp,
+                                                                                          seed=self.seed)]
+
+        return cmds
+
+class RemoveLangevinIntegrator(simulate.SimulationOperation):
```
mphoward, line 218: Here and elsewhere, derive from `LAMMPSOperation`.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+    def __init__(self, add_op):
+        if not isinstance(add_op, AddLangevinIntegrator):
+            raise TypeError('Addition operation is not AddLangevinIntegrator.')
+        self.add_op = add_op
+
+    def to_commands(self, sim):
+        cmds += ['unfix {idx}'.format(idx=self.add_op.idx)]
```
mphoward, line 225: Unfix based on the fix ID you automatically assign. Also, you will need to unfix both the NVE integrator and the langevin fixes.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+
+        return cmds
+
+class AddNPTIntegrator(simulate.SimulationOperation):
+    def __init__(self, idx, group_idx): #TODO: reconcile with generic
```
mphoward, line 230: See comments above: `idx` can be assigned automatically and `group_idx` is `all`.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+        self.idx = idx
+        self.group_idx = group_idx
+
+    def to_commands(self, sim):
+        cmds += ['fix {idx} {group_idx} npt'.format(idx=self.idx,group_idx=self.group_idx)]
```
mphoward, line 235: This needs the additional arguments setting T and P from the ensemble, and the time constants that should be added to the constructor.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+        return cmds
+
+class RemoveNPTIntegrator(simulate.SimulationOperation):
+    def __init__(self, add_op):
+        if not isinstance(add_op, AddNPTIntegrator):
+            raise TypeError('Addition operation is not AddNPTIntegrator.')
+        self.add_op = add_op
+
+    def to_commands(self, sim):
+        cmds += ['unfix {idx}'.format(idx=self.add_op.idx)]
```
mphoward, line 246: Same comments on unfixing here and after.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,290 @@
...
+
+        return cmds
+
+class AddNVTIntegrator(simulate.SimulationOperation):
```
mphoward, line 250: Same changes are needed here as for the NPT integrator.

mphoward (changes requested):
Great continued progress. Here are some comments in response to your questions. I also pushed a commit addressing some of the issues you were having with I/O. I think it was mostly an issue of confusion about writing the table, and then I fixed some spurious I/O things. Let's discuss the open points at our next meeting.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def to_commands(self, sim):
+        cmds = super().to_commands(sim)
+        cmds += ['read_data {filename}'.format(filename=self.filename)]
+        cmds += self.attach_potentials(sim)
+
+        return cmds
+
+class InitializeRandomly(Initialize):
+    def __init__(self, neighbor_buffer, seed, units='lj', atom_style='atomic'):
```
mphoward, line 164: `seed` is an optional argument in `generic` and `hoomd`, but probably it makes more sense to make it required like here for reproducibility. We should also then use the same ordering as in `InitializeFromFile`, which would be: `__init__(self, seed, neighbor_buffer, ...)`. Can you make those changes across all these modules?

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+        cmds += ['mass * 1.0',
+                 'velocity all create {temp} {seed}'.format(temp=sim.ensemble.T,
+                                                            seed=self.seed)]
+
+        cmds += self.attach_potentials(sim)
+
+        return cmds
+
+class MinimizeEnergy(LAMMPSOperation):
+    def __init__(self, energy_tolerance, force_tolerance, max_iterations, dt):
```
mphoward, line 193: The `dt` parameter is not used here. If we switch to FIRE, we would need it though. Deferring this for now.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+        self.energy_tolerance = energy_tolerance
+        self.force_tolerance = force_tolerance
+        self.max_iterations = max_iterations
+
+    def to_commands(self, sim):
+        cmds = ['minimize {etol} {ftol} {maxiter} {maxeval}'.format(etol=self.energy_tolerance,
+                                                                    ftol=self.force_tolerance,
+                                                                    maxiter=self.max_iterations,
+                                                                    maxeval=self.max_iterations)]
```
mphoward, line 202: This will allow up to 100 evaluations of the energy per iteration, which is similar (but not exactly the same) as what the HOOMD minimizer does.
```suggestion
                                                                    maxeval=100*self.max_iterations)]
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+
+# Brownian dynamics not supported by LAMMPS
+
+class AddLangevinIntegrator(LAMMPSOperation):
+    def __init__(self, dt, friction, seed):
+        self.friction = friction
+        self.seed = seed
+
+        self.lgv_idx = super().add_fix()
+        self.nve_idx = super().add_fix()
```
mphoward, line 214: Class methods can be called on an index. Also, these should not be public attributes, so let's call them a little different:
```suggestion
        self._fix_langevin = self.add_fix()
        self._fix_nve = self.add_fix()
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def to_commands(self, sim):
```
mphoward, line 216: The friction factor needs to be treated a little differently here than in HOOMD. The fix only supports one value for `damp`, so we need to use the `scale` parameter multiple times to adjust it for each type.

The damp parameter is related to the friction factor (used in HOOMD) by:
```
damp=mass/friction
```

So, we need to:
1. Get the mass of each type using `extract_atom`.
2. Compute the `damp` factor per-type using the mass and the specified `friction` (which may be a float or an array per-type).
3. Choose one of the `damp` values as the reference, then `scale` each type relative to this value.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+
+class RemoveLangevinIntegrator(LAMMPSOperation):
+    def __init__(self, add_op):
+        if not isinstance(add_op, AddLangevinIntegrator):
+            raise TypeError('Addition operation is not AddLangevinIntegrator.')
+        self.add_op = add_op
+
+    def to_commands(self, sim):
+        cmds = ['unfix {idx}'.format(idx=self.add_op.lgv_idx),
+                'unfix {idx}'.format(idx=self.add_op.nve_idx)]
```
mphoward, line 236: Use the renamed protected variables here.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+
+        return cmds
+
+class AddNPTIntegrator(LAMMPSOperation):
+    def __init__(self, dt, tau_T, tau_P):
+        self.tau_T = tau_T
+        self.tau_P = tau_P
+
+        self.idx = super().add_fix()
```
mphoward, line 245:
```suggestion
        self._fix = super().add_fix()
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,315 @@
...
+    def to_commands(self, sim):
+        cmds = ['unfix {idx}'.format(idx=self.add_op.idx)]
+
+        return cmds
+
+class AddNVTIntegrator(LAMMPSOperation):
+    def __init__(self, dt, tau_T):
+        self.tau_T = tau_T
+
+        self.idx = super().add_fix()
```
mphoward, line 274: Same change as for NPT here, carry forward to other ops.

tests/simulate/test_lammps.py
```diff
@@ -0,0 +1,146 @@
...
+        """Test running initialization simulation operations."""
+        #InitializeFromFile
+        ens,pot = self.ens_pot()
+        file_ = self.create_file()
+        op = relentless.simulate.lammps.InitializeFromFile(filename=file_, neighbor_buffer=0.4)
+        l = relentless.simulate.lammps.LAMMPS(operations=op, quiet=False)
+        sim = l.run(ensemble=ens, potentials=pot, directory=self.directory)
+
+        #InitializeRandomly
+        ens,pot = self.ens_pot()
```
mphoward, line 70: You should be able to reuse the `ens` and `pot` you created above.

tests/simulate/test_lammps.py
```diff
@@ -0,0 +1,146 @@
...
+        sim = l.run(ensemble=ens, potentials=pot, directory=self.directory)
+
+    def test_minimization(self):
+        """Test running energy minimization simulation operation."""
+        #MinimizeEnergy
+        ens,pot = self.ens_pot()
+        file_ = self.create_file()
+        op = [relentless.simulate.lammps.InitializeFromFile(filename=file_, neighbor_buffer=0.4),
+              relentless.simulate.lammps.AddNVTIntegrator(dt=0.5,tau_T=1.0),
+              relentless.simulate.lammps.MinimizeEnergy(energy_tolerance=1e-7,
```
mphoward, line 82: You don't need (or want) the NVT integrator here because MinimizeEnergy executes its own run. You can AddNVTIntegrator and Run after the minimization, or in a separate test.

tests/simulate/test_lammps.py
```diff
@@ -0,0 +1,146 @@
...
+                                                        force_tolerance=1e-7,
+                                                        max_iterations=1000,
+                                                        dt=0.01)
+             ]
+        l = relentless.simulate.lammps.LAMMPS(operations=op, quiet=False)
+        sim = l.run(ensemble=ens, potentials=pot, directory=self.directory)
+
+    def test_integrators(self): #TODO: check if fixing/unfixing actually occurs
```
mphoward, line 90: You can probably do this using the PyLammps wrapper around the `sim.lammps` instance, and checking the `fixes` attribute.

mphoward (changes requested):

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+
+class LAMMPSOperation(simulate.SimulationOperation):
+    fix_id = 1
+
+    def __call__(self, sim):
+        cmds = self.to_commands(sim)
+        sim.lammps.commands_list(cmds)
+
+    @classmethod
+    def add_fix(cls):
```
mphoward, line 55: Let's rename this to make it clear what the method does:
```suggestion
    def new_fix_id(cls):
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+        sim.type_map = {}
+        sim.typeid_map = {}
+        for i,t in enumerate(sim.ensemble.types):
+            sim.type_map[t] = i+1
+            sim.typeid_map[sim.type_map[t]] = t
+
+        return sim
+
+class LAMMPSOperation(simulate.SimulationOperation):
+    fix_id = 1
```
mphoward, line 48: We don't want people accessing this variable publicly, so let's make it clear it's protected and also rename to make it clear that it's just *counting* the fixes and not actually the id of the current operation:
```suggestion
    _fix_counter = 1
```

relentless/simulate/hoomd.py
```diff
@@ -230,15 +230,15 @@ class InitializeRandomly(Initialize):
...
     neighbor_buffer : float
         Buffer width.
-    seed : int
-        The seed to randomly initialize the particle locations (defaults to `None`).
     options : kwargs
         Options for random initialization.
 
     """
-    def __init__(self, neighbor_buffer, seed=None, **options):
+    def __init__(self, seed, neighbor_buffer, **options):
```
mphoward, line 241: We don't use `**options` anywhere here. Is there a reason we needed this, or can it be removed?

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+        cmds = ['units {style}'.format(style=self.units),
+                'boundary p p p',
+                'atom_style {style}'.format(style=self.atom_style)]
+
+        return cmds
+
+    def extract_box_params(self, sim):
+        for t in sim.ensemble.types:
+            if sim.ensemble.N[t] is None:
+                raise ValueError('Number of particles for type {} must be set.'.format(t))
```
mphoward, line 80: I don't think this belongs here since `extract_box_params` does not care about the number of particles. You should put it in `to_commands` of `InitializeRandomly` because `InitializeFromFile` also does not need this check.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+        # make box from ensemble
+        box = self.extract_box_params(sim)
+        if not np.all(np.isclose(box[-3:],0)):
+            cmds += ['region box prism {} {} {} {} {} {} {} {} {}'.format(*box)]
+        else:
+            cmds += ['region box block {} {} {} {} {} {}'.format(*box[:-3])]
+        cmds += ['create_box {N} box'.format(N=len(sim.ensemble.types))]
+
+        # use lammps random initialization routines
+        for i in sim.ensemble.types:
```
mphoward, line 185: You could actually check that N is set for each type here, as you are looping.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+        self._fix_langevin = self.add_fix()
+        self._fix_nve = self.add_fix()
+
+    def to_commands(self, sim):
+        # obtain per-type mass
+        mass_per_type = {}
+
+        mass_ptr = sim.lammps.extract_atom('mass',2)
+        Ntypes = len(sim.ensemble.types)
+        mass = np.ctypeslib.as_array(mass_ptr,(Ntypes+1,))[1:]
```
mphoward, line 225: You should try the `numpy_wrapper` that should do these conversions automatically:
```suggestion
        mass = sim.lammps.numpy.extract_atom('mass')
        if mass is None or len(mass) == Ntypes+1:
            # raise some error because the mass was not obtained
```
I think you should also try to use the masses 1-indexed (as they come from lammps) because we already have the `type_map` to convert a name into the lammps index. You could then get the mass like:
```py
        mass[sim.type_map[t]]
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+
+        types_ptr = sim.lammps.extract_atom('type',0)
+        Natoms = sum(sim.ensemble.N.todict().values())
+        types = np.unique(np.ctypeslib.as_array(types_ptr,(Natoms,)))
+        types = [str(t) for t in types]
+
+        for t,m in zip(types,mass):
+            mass_per_type[t] = m
```
mphoward, line 233: This will give you lots of trouble in MPI because each MPI rank will own some slice of particles, and not all types are present on all ranks. My suggestion above avoids this issue, and this code can be removed.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # obtain per-type friction factor
+        if np.isscalar(self.friction):
+            friction_per_type = {}
+            for t in types:
+                friction_per_type[t] = self.friction
+        else:
+            if sorted(self.friction.keys()) != sorted(types):
+                return ValueError('The friction factor types must match the atom types.')
+            friction_per_type = self.friction
```
mphoward, line 243: You might want to do this in a lazy way like I did in HOOMD so that you don't need as much code:
```py
friction = np.zeros_like(mass)
for t in sim.ensemble.types:
    try:
        friction[sim.type_map[t]] = self.friction[t]
    except KeyError:
        friction[sim.type_map[t]] = self.friction
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # compute per-type damping parameter
+        damp_per_type = {}
+        for t in friction_per_type:
+            damp_per_type[t] = mass_per_type[t]/friction_per_type[t]
```
mphoward, line 248: You can use safe-divide to do this:
```py
damp = np.divide(mass,friction,where=(friction > 0))
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+
+        damp_ref = list(damp_per_type.values())[0]
+        scale_str = ''
+        for t in damp_per_type:
+            if damp_per_type[t] == damp_ref:
+                continue
+            scale_str += (str(t) + ' ' + str(damp_per_type[t]/damp_ref) + ' ')
+        scale_str.rstrip()
```
mphoward, line 256: I think this will achieve what you're trying to do more compactly. I also repeated the "scale" keyword, since I think you need it for each type? Note the `i+1` is important to index the types correctly with enumerate.
```suggestion
        damp_ref = damp[1]
        scale = damp / damp_ref
        scale_str = ' '.join(['scale {} {}'.format(i+1,s) for i,s in enumerate(scale[1:])])
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,361 @@
...
+
+        cmds = ['fix {idx} {group_idx} langevin {t_start} {t_stop} {damp} {seed} scale {scaling}'.format(idx=self._fix_langevin,
```
mphoward, line 258:
```suggestion
        cmds = ['fix {idx} {group_idx} langevin {t_start} {t_stop} {damp} {seed} {scaling}'.format(idx=self._fix_langevin,
```

mphoward (changes requested):
Great work! Here are some suggestions, and I will look into the thermo analysis soon.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,347 @@
...
+    _lammps_found = True
+except ImportError:
+    _lammps_found = False
+
+class LAMMPS(simulate.Simulation):
+    def __init__(self, operations, quiet=True, **options):
+        if not _lammps_found:
+            raise ImportError('LAMMPS not found.')
+        elif lammps.lammps().version() < 20201029:
+            raise ImportError('Only LAMMPS 29 Oct 2020 or newer is supported.')
```
mphoward, line 20: Ah, I forgot that `version()` is only a property of a _running_ lammps instance. We don't want to make one of these before we actually need it just to check the version, so move this check down into `_new_instance` (after you create `sim.lammps`).

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,347 @@
...
+            except TypeError:
+                friction[sim.type_map[t]-1] = self.friction
+            except KeyError:
+                return KeyError('The friction factor types must match the atom types.')
+
+        # compute per-type damping parameter
+        damp = np.divide(mass, friction, where=(friction>0))
+        damp_ref = damp[0]
+        scale = damp/damp_ref
+        scale_str = ' '.join(['scale {} {}'.format(i+1,s) for i,s in enumerate(scale[1:])])
```
mphoward, line 242: This is changing the lammps 1-indexing back to 0-indexing. Since this is only used internally, I think it would be more readable to stick with the LAMMPS indexing and make a note about it.
```suggestion
        # obtain per-type mass
        # all arrays will be 1-indexed using the lammps convention for typeid
        Ntypes = len(sim.ensemble.types)
        mass = sim.lammps.numpy.extract_atom('mass')
        if mass is None or mass.shape != (Ntypes+1,1):
            raise ValueError('Per-type masses not set.')

        # obtain per-type friction factor
        friction = np.zeros_like(mass)
        for t in sim.ensemble.types:
            try:
                friction[sim.type_map[t]] = self.friction[t]
            except TypeError:
                friction[sim.type_map[t]] = self.friction

        # compute per-type damping parameter
        damp = np.divide(mass, friction, where=(friction>0))
        damp_ref = damp[1]

        # rescale damp parameters if there are multiple types
        if Ntypes > 1:
            scale = damp/damp_ref
            scale_str = ' '.join(['scale {} {}'.format(i+1,s) for i,s in enumerate(scale[2:])])
        else:
            scale_str = ''
```

I also made the following changes:
1. Remove the `KeyError`. I'm not sure I see the case where this would ever trigger? Do you have an example?
2. Only compute scale if there are at least 2 types, as this would produce an error with a 1-type system.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,347 @@
...
+        self.seed = seed
+
+        self._fix_langevin = self.new_fix_id()
+        self._fix_nve = self.new_fix_id()
+
+    def to_commands(self, sim):
+        # obtain per-type mass
+        Ntypes = len(sim.ensemble.types)
+        _mass = sim.lammps.numpy.extract_atom('mass')
+        if _mass is None or _mass.shape != (Ntypes+1,1):
```
mphoward, line 224: Is `_mass` really being returned as a 2d array? If so, we should coerce it into a 1d array so that its shape is `(Ntypes+1,)`. Maybe try specifying `dim=1` in `extract_atom`?

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,347 @@
...
+        self.tau_P = tau_P
+
+        self._fix = super().new_fix_id()
+
+    def to_commands(self, sim):
+        if not sim.ensemble.aka('NPT'):
+            raise ValueError('Simulation ensemble is not NPT.')
+        cmds = ['fix {idx} {group_idx} npt temp {Tstart} {Tstop} {Tdamp} iso {Pstart} {Pstop} {Pdamp}'.format(idx=self._fix,
+                                                                                                              group_idx='all',
+                                                                                                              Tstart = sim.ensemble.T,
```
mphoward, line 280: Small style point: remove whitespace around `=` signs in function keyword args.

tests/requirements.txt
```diff
@@ -1,3 +1,4 @@
 hoomd=2.9.3
 freud=2.2.0
 gsd=2.2.0
+lammps=2020.10.29
```
mphoward, line 4: I'm glad this tag is working, it wasn't for me before!

tests/simulate/test_lammps.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def test_integrators(self):
+        """Test adding and removing integrator operations."""
+        default_fixes = [{'name':''}]
+
+        file_ = self.create_file()
+        init = relentless.simulate.lammps.InitializeFromFile(filename=file_, neighbor_buffer=0.4)
+        l = relentless.simulate.lammps.LAMMPS(operations=init, quiet=False)
+
+        #LangevinIntegrator
```
mphoward, line 100: Can you run a test of the Langevin integrator with only one type? That would catch bugs like the one I think I spotted.

mphoward (changes requested):

relentless/simulate/lammps.py
mphoward: There's also `np.squeeze` if you want to just smoosh the array down.

https://numpy.org/doc/stable/reference/generated/numpy.squeeze.html

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,870 @@
...
+            sim.type_map[t] = i+1
+            sim.typeid_map[sim.type_map[t]] = t
+
+        return sim
+
+class LAMMPSOperation(simulate.SimulationOperation):
+    _fix_counter = 1
+
+    def __call__(self, sim):
+        """Evaluates the LAMMPS simulation operation.
```
mphoward, line 62: Add a sentence describing how this works. Each deriving class must implement a `to_commands` method that returns a list or tuple of LAMMPS commands that can be executed by the LAMMPS `commands_list` method.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,870 @@
...
+
+    @classmethod
+    def new_fix_id(cls):
+        # sets fix_id for LAMMPS fixes
+        idx = int(cls._fix_counter)
+        cls._fix_counter += 1
+        return idx
+
+    @abc.abstractmethod
+    def to_commands(self, sim):
```
mphoward, line 81: Document that this is the method that deriving classes should override, and it should return a list of valid LAMMPS commands.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,870 @@
...
+        The LAMMPS style of atoms used in a simulation (defaults to `atomic`).
+
+    """
+    def __init__(self, neighbor_buffer, units='lj', atom_style='atomic'):
+        self.neighbor_buffer = neighbor_buffer
+        self.units = units
+        self.atom_style = atom_style
+
+    def to_commands(self, sim):
+        """Calls the appropriate LAMMPS commands for the initialization simulation operation.
```
mphoward, line 103: Don't bother documenting this method in any of the deriving objects. It would be better to put useful implementation details in the class docstring.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,870 @@
...
+            If the simulation ensemble does not have constant V.
+        ValueError
+            If the per-type particle masses are not set.
+        KeyError
+            If the friction factor for all types are not set (if per-type
+            friction factors are given).
+
+        """
+        if not sim.ensemble.constant['V']:
+            raise ValueError('Simulation ensemble does not have constant V.')
```
mphoward, line 427: You should instead check if this is NVT ensemble.

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,870 @@
...
+
+        # obtain per-type mass (arrays 1-indexed using lammps convention)
+        Ntypes = len(sim.ensemble.types)
+        mass = np.squeeze(sim.lammps.numpy.extract_atom('mass'))
+        if mass is None or mass.shape != (Ntypes+1,):
+            raise ValueError('Per-type masses not set.')
```
mphoward, line 433: `np.squeeze` will fail if `extract_atom` returns `None`.
```suggestion
        mass = sim.lammps.numpy.extract_atom('mass')
        if mass is None or mass.shape != (Ntypes+1,1):
            raise ValueError('Per-type masses not set.')
        mass = np.squeeze(mass)
```

relentless/simulate/lammps.py
```diff
@@ -0,0 +1,870 @@
...
+        # compute per-type damping parameter and rescale if multiple types
+        damp = np.divide(mass, friction, where=(friction>0))
+        damp_ref = damp[1]
+        if Ntypes>1:
+            scale = damp/damp_ref
+            scale_str = ' '.join(['scale {} {}'.format(i+1,s) for i,s in enumerate(scale[2:])])
+        else:
+            scale_str = ''
+
+        cmds = ['fix {idx} {group_idx} langevin {t_start} {t_stop} {damp} {seed} {scaling}'.format(idx=self._fix_langevin,
```
mphoward, line 454: This is a pedantic point, and I'm not sure if it even matters, but can you put the NVE fix before the langevin fix to be safe?

tests/simulate/test_lammps.py
```diff
@@ -0,0 +1,220 @@
...
+        file_ = self.directory.file('test.data')
+        with open(file_,'w') as f:
+            f.write(('LAMMPS test data\n'
+                     '\n'
+                     '5 atoms\n'
+                     '2 atom types\n'
+                     '\n'
+                     '0.0 10.0 xlo xhi\n'
+                     '0.0 10.0 ylo yhi\n'
+                     '0.0 10.0 zlo zhi\n'
```
mphoward, line 44: LAMMPS supports general box origins like this, but we made the choice in `TriclinicBox` to use boxes centered around (0,0,0), so could you shift all your coordinates by -5.0? Like,
```suggestion
                     '-5.0 5.0 xlo xhi\n'
                     '-5.0 5.0 ylo yhi\n'
                     '-5.0 5.0 zlo zhi\n'
```
And similarly in the Atoms section.

mphoward (changes requested):
Just a couple small comments left, and then I think we can merge. We'll clean up any remaining small items in subsequent PRs.

mphoward (approved):
Thank you, this is impressive!!

PR #52, opened 2021-02-08
Implement optimize module

adisr04:
- Add `ObjectiveFunction` abstract base class, with `ObjectiveFunctionResult` helper class
- Add `Optimizer` abstract base class, with `SteepestDescent` concrete implementation
- Document and add unit tests for `objective.py` (using `QuadraticObjective` concrete  implementation)
- Document and add unit tests for `method.py`


relentless/optimize/__init__.py
```diff
@@ -1,3 +1,2 @@
-from .core import *
-from .algorithm import *
 from .objective import *
+from .optimizer import *
```
mphoward, line 2: When I refactored code, I have moved away from using `import *`, as it is better to explicit than implicit. A few things that we want to tweak related to this:
1. Our organization has been to implement a file `optimize.py` (`from . import optimize`) that contains the abstract base classes, then other modules with specific instances. You can explicitly import these into `__init__.py` using a tuple (go check one of the other modules).
2. Let's rename `optimizer` to `method`, I think this is cleaner.
3. Remove `from . import *` and replace by `from . import module_name`.
4. Remove `__all__` from the modules, as we are not doing any `import *` anywhere, so we don't really need these.

relentless/optimize/objective.py
```diff
@@ -1,84 +1,34 @@
...
-                diff += scipy.integrate.trapz(x=r, y=4.*np.pi*r**2*(sim_factor*gsim[i,j](r)-tgt_factor*gtgt[i,j](r))**2)
-
-        return diff
+__all__ = ['ObjectiveFunction','Result','RelativeEntropy']
+
+import abc
+
+class ObjectiveFunction(abc.ABC):
+    @abc.abstractmethod
+    def run(self):
```
mphoward, line 7: I propose to rename this method `compute` to make it clear that it is doing a calculation.

relentless/optimize/objective.py
```diff
@@ -1,84 +1,34 @@
...
+        pass
+
+    @abc.abstractmethod
+    def design_variables(self):
+        pass
+
+class Result:
+    def __init__(self, value, gradient):
+        self.value = value
+        self.gradient = gradient
```
mphoward, line 17: This class is actually not very meaningful as a standalone object, so you could nest it inside `ObjectiveFunction`, so that its namespace becomes `ObjectiveFunction.Result`.

mphoward, line 17: Should we add a method that safely checks the gradient?
```
def gradient(self, var):
    try:
        return self._gradient[var]
    except KeyError:
        return 0.0
```
I could imagine this being a common pattern?

relentless/optimize/objective.py
```diff
@@ -1,84 +1,34 @@
...
+
+class RelativeEntropy(ObjectiveFunction):
```
mphoward, line 19: This is a shell for now, so I am not reviewing it yet. For cleanness on the PR, could you remove this and we'll focus on the abstract base classes / interface? We can then specifically implement RelativeEntropy later.

relentless/optimize/optimizer.py
```diff
@@ -0,0 +1,23 @@
...
+import abc
+
+from .objective import ObjectiveFunction
+
+class Optimizer(abc.ABC):
+    def __init__(self, obj):
+        self.objective = obj
+
+    @abc.abstractmethod
+    def run(self):
```
mphoward, line 12: Maybe `optimize` or `compute` instead?

I think we should change this interface slightly too. The Optimizer doesn't necessarily need to know the objective function when it is constructed, since it only needs it at the time it is called. (Or equivalently, you could use an Optimizer to minimize multiple objective functions using the same hyperparameters.)

So, I would move the objective function to be an argument of `run`. The constructor should be used for setting hyperparameters (like alpha below).

relentless/optimize/optimizer.py
```diff
@@ -0,0 +1,23 @@
...
+        pass
+
+class SteepestDescent(Optimizer):
+    def __init__(self, obj, alpha):
+        super().__init__(obj)
+        self.alpha = alpha
+
+    def run(self):
+        res = self.objective.run()
+        for x in res.gradient:
```
mphoward, line 22: I think we want to get the design variables first, and iterate over those. It should be the case that the design variables are the same as what's in the result, but I guess this is not necessarily guaranteed. With the new `gradient` method I suggested:
```
vars = self.objective.design_variables()
for x in vars:
    grad = res.gradient(x)
```

relentless/optimize/optimizer.py
```diff
@@ -0,0 +1,23 @@
...
+            x.value += self.alpha*res.gradient[x]
```
mphoward, line 23: This will get fleshed out more later, since it's missing constraints and stopping criteria right now. It also needs to iterate to convergence. I think also the sign is wrong and should be `-=`?

tests/optimize/test_objective.py
```diff
@@ -0,0 +1,30 @@
...
+class QuadraticObjective(relentless.optimize.ObjectiveFunction):
+    """Mock objective function used to test relentless.optimize.ObjectiveFunction"""
+
+    def __init__(self, x):
+        self.x = x
+
+    def run(self):
+        val = self.x.value**2
+        grad = {self.x:2*self.x.value}
+        return relentless.optimize.Result(val, grad)
```
mphoward, line 15: If you nest `Result`, this can be shortened as:
```
return self.Result(val, grad)
```

tests/optimize/test_objective.py
```diff
@@ -0,0 +1,30 @@
...
+    """Unit tests for relentless.optimize.ObjectiveFunction"""
+
+    def test_run(self):
+        """Test run method"""
+        x = relentless.variable.DesignVariable(value=3.0)
+        q = QuadraticObjective(x=x)
+
+        res = q.run()
+        self.assertAlmostEqual(res.value, 9.0)
+        self.assertAlmostEqual(res.gradient[x], 6.0)
```
mphoward, line 30: Update if you change the gradient method.

tests/optimize/test_objective.py
```diff
@@ -0,0 +1,30 @@
...
+        val = self.x.value**2
+        grad = {self.x:2*self.x.value}
+        return relentless.optimize.Result(val, grad)
+
+    def design_variables(self):
+        return (self.x,)
+
+class test_ObjectiveFunction(unittest.TestCase):
+    """Unit tests for relentless.optimize.ObjectiveFunction"""
+
```
mphoward, line 22: Add a test for the `design_variables`?

tests/optimize/test_optimizer.py
```diff
@@ -0,0 +1,21 @@
+"""Unit tests for optimizer module."""
+import unittest
+
+import relentless
+
+from .test_objective import QuadraticObjective
+
+class test_Optimizer(unittest.TestCase):
```
mphoward, line 8: This is really a test of `SteepestDescent`, so I would rename accordingly.

tests/optimize/test_objective.py
```diff
@@ -0,0 +1,30 @@
...
+class QuadraticObjective(relentless.optimize.ObjectiveFunction):
+    """Mock objective function used to test relentless.optimize.ObjectiveFunction"""
+
+    def __init__(self, x):
+        self.x = x
+
+    def run(self):
+        val = self.x.value**2
+        grad = {self.x:2*self.x.value}
```
mphoward, line 14: To make these a little more interesting, how about you shift to make the function `(x-1)^2`? The minimum of this function is `x = 1`.

tests/optimize/test_optimizer.py
```diff
@@ -0,0 +1,21 @@
...
+        """Test run method."""
+        x = relentless.variable.DesignVariable(value=3.0)
+        q = QuadraticObjective(x=x)
+        o = relentless.optimize.SteepestDescent(obj=q, alpha=0.25)
+
+        o.run()
+        self.assertAlmostEqual(x.value, 4.5)
+
+        o.run()
+        self.assertAlmostEqual(x.value, 6.75)
```
mphoward, line 21: These tests are not correct. You should check that `x.value` gets (close) to the analytically known minimum of the objective. It's OK for this test to fail right now, as you haven't finished implementing the optimizer.

mphoward (approved):
Good first pass! Here is some feedback and suggestions.

mphoward:
[Sorry, I pushed the approve button by mistake. :-)]

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+from .objective import ObjectiveFunction
+
+class Optimizer(abc.ABC):
+
+    @abc.abstractmethod
+    def optimize(self, objective):
+        pass
+
+class SteepestDescent(Optimizer):
+    def __init__(self, alpha, max_iter, abs_tol, mode):
```
mphoward, line 12: It doesn't do anything now, but just in case, you should call the parent constructor
```
super().__init__()
```

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def optimize(self, objective):
+        dvars = objective.design_variables()
+        iter_num = 1
+        update = {}
+
+        while True:
+            res = objective.compute()
+            for x in dvars:
+                update[x] = self.alpha*res.gradient(x)
```
mphoward, line 26: You can do this in one line with list comprehension:
```
update = {self.alpha*res.gradient(x) for x in dvars}
```

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+
+            if iter_num > self.max_iter:
```
mphoward, line 28: I don't think it should be an error to not converge... it might be better if `optimize` returned a bool saying whether the result converged or not, as the user could then decide to continue the optimization.

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+        self.max_iter = max_iter
+        self.abs_tol = abs_tol
+        self.mode = mode
+
+    def optimize(self, objective):
+        dvars = objective.design_variables()
+        iter_num = 1
+        update = {}
+
+        while True:
```
mphoward, line 23: `while True` is scary! I would rather see this standard idiom:
```
iter_num = 0
converged = False
while not converged and iter_num < self.max_iter:
    ...
    converged = res._has_converged(update)
    iter_num += 1
```

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+            res = objective.compute()
+            for x in dvars:
+                update[x] = self.alpha*res.gradient(x)
+
+            if iter_num > self.max_iter:
+                raise RuntimeError('Could not find solution within ' + str(self.max_iter) + ' iterations.')
+            elif self._has_converged(update):
```
mphoward, line 30: I think I would still always apply the update, even if the method has converged, because it shouldn't hurt to advance that remaining small bit.

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+                print('Solution found in ' + str(iter_num) + ' iterations.')
+                break
+            #TODO: add this branch?
```
mphoward, line 33: Not sure what this does, so I would remove it.

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+            #elif alpha*update/value
+                # raise ValueError('Converged to wrong minimum.')
+            else:
+                for x in dvars:
+                    x.value -= update[x]
+
+            iter_num += 1
+
+    def _has_converged(self, update):
```
mphoward, line 42: Should this be a method of the `Optimizer`? Should it be public? Also, shouldn't this accept the result directly? I think we talked about checking the value of the components of the gradient rather than the step that's applied.

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+        if self.mode=='grad_diff':
+            for x in update:
+                if abs(update[x]) > (self.abs_tol[x] if isinstance(self.abs_tol, dict) else self.abs_tol):
```
mphoward, line 45: Maybe `try-except` or preprocess the dict? This is pretty hard to read.

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+            #elif alpha*update/value
+                # raise ValueError('Converged to wrong minimum.')
+            else:
+                for x in dvars:
+                    x.value -= update[x]
+
+            iter_num += 1
+
+    def _has_converged(self, update):
+        if self.mode=='grad_diff':
```
mphoward, line 43: I find this mode name pretty unclear. For now, let's omit a mode (and just use the one test) until we come up with a list of concrete other options.

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @abc.abstractmethod
+    def optimize(self, objective):
+        pass
+
+class SteepestDescent(Optimizer):
+    def __init__(self, alpha, max_iter, abs_tol, mode):
+        self.alpha = alpha
+        self.max_iter = max_iter
+        self.abs_tol = abs_tol
```
mphoward, line 15: Do these need property getters / setters to check on values?

relentless/optimize/method.py
```diff
@@ -0,0 +1,49 @@
...
+        self.mode = mode
+
+    def optimize(self, objective):
+        dvars = objective.design_variables()
```
mphoward, line 19: Exit early if there are no design variables?

tests/optimize/test_objective.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_compute(self):
+        """Test compute method"""
+        x = relentless.variable.DesignVariable(value=4.0)
+        q = QuadraticObjective(x=x)
+
+        res = q.compute()
+        self.assertAlmostEqual(res.value, 9.0)
+        self.assertAlmostEqual(res.gradient(x), 6.0)
+        self.assertEqual(res.gradient('x'), 0.0)
```
mphoward, line 31: This test confused me at first. Maybe make it another variable (rather than a string with the same name) to make the intent clear?

mphoward (changes requested):
Great work! Here are some additional suggestions.

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, alpha, max_iter, abs_tol):
+        super().__init__()
+        self._alpha = alpha
+        self._max_iter = max_iter
+        self._abs_tol = abs_tol
+
+    def optimize(self, objective):
+        dvars = objective.design_variables()
+        if len(dvars) == 0:
+            return False
```
mphoward, line 21: Maybe `return None` as nothing was done? (It is neither converged nor unconverged.)

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+
+        iter_num = 0
+        converged = False
+        while not converged and iter_num < self.max_iter:
+            res = objective.compute()
+            converged = self._has_converged(dvars, res)
+            for x in dvars:
+                x.value -= self.alpha*res.gradient(x)
```
mphoward, line 29: I was thinking about this again, and I'm not sure whether or not you should apply the update if the algorithm is determined to have converged. The distinction probably doesn't matter too much, but we should make sure we're in agreement. Let's discuss at our meeting.

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _has_converged(self, dvars, res):
+        for x in dvars:
+            grad = res.gradient(x)
+            try:
+                if abs(grad) > self.abs_tol:
+                    return False
+            except TypeError:
+                if abs(grad) > self.abs_tol[x]:
+                    return False
```
mphoward, line 42: Hmmm, this is definitely better but still reads a little funny. What happens in the except block if the key for `x` does not exist? I guess it will raise a `KeyError` with some ugly warning about an exception being raised while handling the other exception? Maybe try something like this to get more sensible output? It also reduces code duplication.
```py
try:
    tol = self.abs_tol[x]
except TypeError:
    tol = self.abs_tol
except KeyError:
    raise KeyError('Tolerance not set for one of the design variables')
if abs(grad) > tol:
    return False
```

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @property
+    def max_iter(self):
+        return self._max_iter
+
+    @max_iter.setter
+    def max_iter(self, value):
+        if not isinstance(value, int):
+            raise TypeError('The maximum number of iterations must be an integer.')
+        if value <= 0:
```
mphoward, line 63: You've required it to be an integer, so just check if it's < 1.
```suggestion
        if value < 1:
```

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+            raise ValueError('The maximum number of iterations must be positive.')
+        self._max_iter = value
+
+    @property
+    def abs_tol(self):
+        return self._abs_tol
+
+    @abs_tol.setter
+    def abs_tol(self, value):
+        try:
```
mphoward, line 73: I find this part confusing to read. When would a `TypeError` be raised inside the try block? Maybe you want something more like this:
```py
try:
    self._abs_tol = dict(value)
    err = any([tol < 0 for tol in self._abs_tol.values()])
except TypeError:
    self._abs_tol = value
    err = self._abs_tol < 0
if err:
    raise ValueError('Tolerances must be non-negative')
```

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+    @abc.abstractmethod
+    def optimize(self, objective):
+        pass
+
+class SteepestDescent(Optimizer):
+    def __init__(self, alpha, max_iter, abs_tol):
+        super().__init__()
+        self._alpha = alpha
+        self._max_iter = max_iter
+        self._abs_tol = abs_tol
```
mphoward, line 16: Assign these using the setters on these so that you get error checking?

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+        while not converged and iter_num < self.max_iter:
+            res = objective.compute()
+            converged = self._has_converged(dvars, res)
+            for x in dvars:
+                x.value -= self.alpha*res.gradient(x)
+            iter_num += 1
+
+        return converged
+
+    def _has_converged(self, dvars, res):
```
mphoward, line 34: Let's discuss whether this method should be moved to `Optimizer`, potentially along with the `abs_tol` property.

relentless/optimize/method.py
```diff
@@ -0,0 +1,83 @@
...
+            try:
+                if abs(grad) > self.abs_tol:
+                    return False
+            except TypeError:
+                if abs(grad) > self.abs_tol[x]:
+                    return False
+        return True
+
+    @property
+    def alpha(self):
```
mphoward, line 46: I didn't notice this before, but you do want to call this something more descriptive like `step` or `step_size`? `alpha` is very unclear without seeing a mathematical formula.

mphoward (changes requested):

relentless/optimize/method.py
```diff
@@ -0,0 +1,84 @@
...
+    @abs_tol.setter
+    def abs_tol(self, value):
+        try:
+            self._abs_tol = dict(value)
+            err = any([tol < 0 for tol in self._abs_tol.values()])
+        except TypeError:
+            self._abs_tol = value
+            err = self._abs_tol < 0
+        if err:
+            raise ValueError('Absolute tolerances must be non-negative.')
```
mphoward, line 40: This is a minor point: you don't actually want to set `self._abs_tol` if there is an error, as it leaves the object in a state where the value is invalid after the exception is raised.

I would suggest replacing `self._abs_tol = value` etc. by `abs_tol = value` on lines 34-38, then after you raise the exception, set `self._abs_tol = abs_tol`. That way, you only perform the setting if there is no error.

relentless/optimize/method.py
```diff
@@ -0,0 +1,84 @@
...
+
+class SteepestDescent(Optimizer):
+    def __init__(self, step_size, max_iter, abs_tol):
```
mphoward, line 43: I would prefer to keep parent arguments listed first, so move `abs_tol` to the beginning of this list.

relentless/optimize/objective.py
```diff
@@ -1,84 +1,32 @@
...
-        # try loading the simulation
-        try:
-            with env.data(step):
-                thermo = Ensemble.load('ensemble')
-        except FileNotFoundError:
-            thermo = None
+            # gradient must be keyed only on objective design variables
+            self.__gradient = _collections.FixedKeyDict(keys=objective.design_variables())
+            self.__gradient.update(gradient)
+            self._gradient = self.__gradient.todict()
```
mphoward, line 14: You don't actually want to save a separate `__gradient` variable (be careful with double underscores, it does weird name mangling). I think this is equivalent?
```suggestion
            self._gradient = _collections.FixedKeyDict(keys=objective.design_variables())
            self._gradient.update(gradient)
```

relentless/optimize/objective.py
```diff
@@ -1,84 +1,32 @@
...
+                return self._gradient[var]
+            except KeyError:
+                return 0.0
 
-            self.engine.run(env, step)
-            thermo = self.engine.process(env, step)
-            with env.data(step):
-                thermo.save('ensemble')
+        @property
+        def variables(self):
```
mphoward, line 23: For consistency, should this be:
```suggestion
        def design_variables(self):
```
?

mphoward (changes requested):
Excellent work! I have a few small points for you to address, and then this coding is done. Still to do:

- Please document all new modules and classes with docstrings.
- Add the necessary sphinx files to build the new documentation, and confirm there are no errors. (Look at other modules for examples, ask me if you have questions.)
- Please convert PR from draft status.

relentless/optimize/method.py
```diff
@@ -0,0 +1,193 @@
...
+        Parameters
+        ----------
+        objective : :py:class:`ObjectiveFunction`
+            The objective function to be optimized.
+
+        """
+        pass
+
+    def has_converged(self, result):
+        """Checks if the convergence criteria is satisfied.
```
mphoward, line 72: Again, here:
```suggestion
        """Check if the convergence criteria is satisfied.
```
and so on for the rest.

relentless/optimize/method.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Parameters
+        ----------
+        result : :py:class:`ObjectiveFunctionResult`
+            The computed value of the objective function.
+
+        Returns
+        -------
+        bool
+            ``True`` if the criteria is satisfied, ``False`` otherwise.
```
mphoward, line 85: Boolean logic does not need both cases defined. :-)
```suggestion
            ``True`` if the criteria is satisfied.
```

relentless/optimize/method.py
```diff
@@ -0,0 +1,193 @@
...
+            abs_tol = value
+            err = value < 0
+        if err:
+            raise ValueError('Absolute tolerances must be non-negative.')
+        else:
+            self._abs_tol = abs_tol
+
+class SteepestDescent(Optimizer):
+    r"""Steepest descent algorithm.
+
```
mphoward, line 125: This needs a long description showing the algorithm as an equation with the `.. math::` directive.

relentless/optimize/objective.py
```diff
@@ -1,84 +1,133 @@
-__all__ = ['RelativeEntropy']
+"""
+Objective Functions
+===================
 
-import numpy as np
-import scipy.integrate
+An :class:`ObjectiveFunction` defines an objective function to be optimized.
```
mphoward, line 5: Same here, try to write something a little plainer for people to read.

relentless/optimize/objective.py
```diff
@@ -1,84 +1,133 @@
...
+
+    """
+    def __init__(self, value, gradient, objective):
+        self.value = value
+
+        self.design_variables = objective.design_variables()
+        self._gradient = _collections.FixedKeyDict(keys=self.design_variables)
+        self._gradient.update(gradient)
+
+        self.variable_values = {x : x.value for x in self.design_variables}
```
mphoward, line 112: Is this the best name for this attribute? The code on Slack was just a snippet.

Maybe we should combine this with the `design_variables` here to be a dictionary with the objects as keys and the values as values (basically, what is currently `variable_values`). Then, you could still do:
```
for var in result.design_variables:
    pass
```
to iterate the objects, and you could get their values at the time like:
```
result.design_variables[var]
```
Regardless, I would store this as a private variable (`_variable_values`) with a property getter (no setter), and make the storage type be a `FixedKeyDict` for consistency with the gradient.

relentless/simulate/hoomd.py
```diff
@@ -198,7 +198,7 @@ class InitializeFromFile(Initialize):
     neighbor_buffer : float
         Buffer width.
     options : kwargs
-        Options for file reading (as used in :py:func:`hoomd.init.read_gsd()`).
+        Options for file reading (as used in :py:meth:`hoomd.init.read_gsd()`).
```
mphoward, line 201: Did this cause a compile error? `read_gsd` is a function not a class method in HOOMD 2.9.x. If it is causing issues, we might need to change the intersphinx url so it finds the right version on RTD.

Actually, maybe we haven't linked HOOMD into the intersphinx docs yet? Let's make a note to do that.

relentless/simulate/hoomd.py
```diff
@@ -433,7 +433,7 @@ class AddBrownianIntegrator(AddMDIntegrator):
     seed : int
         Seed used to randomly generate a uniform force.
     options : kwargs
-        Options used in :py:func:`hoomd.md.integrate.brownian()`.
+        Options used in :py:meth:`hoomd.md.integrate.brownian()`.
```
mphoward, line 436: Similarly, this (& all the integration methods) should be a `:class:`

relentless/simulate/lammps.py
```diff
@@ -65,8 +66,8 @@ def __call__(self, sim):
         """Evaluates the LAMMPS simulation operation.
 
         Each deriving class of :py:class:`LAMMPSOperation` must implement a
-        :py:func:`to_commands()` method that returns a list or tuple of LAMMPS
-        commands that can be executed by :py:func:`lammps.commands_list()`.
+        :py:meth:`to_commands()` method that returns a list or tuple of LAMMPS
+        commands that can be executed by :py:meth:`lammps.commands_list()`.
```
mphoward, line 70: Not sure if we can get LAMMPS into intersphinx, will look into it.

tests/optimize/test_objective.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_compute(self):
+        """Test compute method"""
+        x = relentless.variable.DesignVariable(value=4.0)
+        q = QuadraticObjective(x=x)
+
+        res = q.compute()
+        self.assertAlmostEqual(res.value, 9.0)
+        self.assertAlmostEqual(res.gradient(x), 6.0)
+        self.assertCountEqual(res.design_variables, q.design_variables())
+        self.assertDictEqual(res.variable_values, {x : 4.0})
```
mphoward, line 34: This will need updating if you make the change I suggested.

mphoward (changes requested):
Looking really good! Here are some final comments, and then I think we can merge this.

adisr04:
So far, the only Sphinx documentation is for the `potential` module, `variable.py`, and the `optimize` module (albeit without examples). I haven't actually set up the Sphinx for the `simulate` module and others. I figured that we would do that in another PR. What do you think?

mphoward:
Yes, no need to update that now. It's best to keep changes localized to the files you're working on. I thought that you were making changes in `simulate` because of sphinx errors, but we will deal with that later. Also, thanks for the corrections to the other files, much appreciated.

mphoward (approved):

PR #53, opened 2021-03-10
Implement line search algorithm for hyperparameter optimization

adisr04:
- Add `LineSearch` class with documentation in `method.py`
- Add `LineSearch` as an optional parameter for `SteepestDescent` objects and modify `SteepestDescent.optimize()` to use the line search if designated
- Expand `SteepestDescent` unit tests and add `LineSearch` unit tests in `test_method.py`
- Add `KeyedArray` class with documentation in `_collections.py`
- Add `KeyedArray` unit tests in `test_collections.py`

### DIFF
--- a/relentless/_collections.py
+++ b/relentless/_collections.py
@@ -272,8 +272,9 @@ class PairMatrix:
 class KeyedArray(FixedKeyDict):
     """Numerical array with fixed keys.
 
-    Can be used to perform arithmetic operations element-wise (per key) and
-    vector algebraic operations (norm, dot product).
+    Can be used to perform arithmetic operations between two arrays (element-wise)
+    or between an array and a scalar, as well as vector algebraic operations
+    (norm, dot product).
 
     Parameters
     ----------
@@ -294,12 +295,23 @@ class KeyedArray(FixedKeyDict):
         k1.update({'A':2.0, 'B':3.0})
         k2.update({'A':3.0, 'B':4.0})
 
-    Perform vector arithmetic operations::
+    Perform array-array arithmetic operations::
 
         >>> print(k1 + k2)
         {'A':5.0, 'B':7.0}
         >>> print(k1 - k2)
         {'A':-1.0, 'B':-1.0}
+        >>> print(k1*k2)
+        {'A':6.0, 'B':12.0}
+        >>> print(k1/k2)
+        {'A':0.6666666666666666, 'B':0.75}
+
+    Perform array-scalar arithmetic operations::
+
+        >>> print(k1 + 3)
+        {'A':5.0, 'B':6.0}
+        >>> print(3 - k1)
+        {'A':1.0, 'B':0.0}
         >>> print(3*k1)
         {'A':6.0, 'B':9.0}
         >>> print(k1/10)
@@ -321,71 +333,142 @@ class KeyedArray(FixedKeyDict):
 
     def _assert_same_keys(self, val):
         if (self.keys != val.keys):
-            raise KeyError('''Both KeyedArrays must have identical keys
-                              to perform mathematical operations.''')
-
-    def _assert_scalar(self, val):
-        if not np.isscalar(val):
-            raise TypeError('''A scalar is required for multiplication or
-                               division with a KeyedArray.''')
+            raise KeyError('Both KeyedArrays must have identical keys to perform mathematical operations.')
 
     def __add__(self, val):
-        """Element-wise addition of two arrays."""
-        self._assert_same_keys(val)
+        """Element-wise addition of two arrays, or of an array and a scalar."""
         k = KeyedArray(keys=self.keys)
-        k.update({x: self[x] + val[x] for x in self})
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            k.update({x: self[x] + val[x] for x in self})
+        elif np.isscalar(val):
+            k.update({x: self[x] + val for x in self})
+        else:
+            raise TypeError('A KeyedArray can only add a scalar or a KeyedArray.')
+        return k
+
+    def __radd__(self, val):
+        """Element-wise addition of a scalar and an array."""
+        k = KeyedArray(keys=self.keys)
+        if np.isscalar(val):
+            k.update({x: val + self[x] for x in self})
+        else:
+            raise TypeError('A KeyedArray can only add a scalar or a KeyedArray.')
         return k
 
     def __iadd__(self, val):
-        """Iterative element-wise addition of two arrays."""
-        self._assert_same_keys(val)
-        self.update({x: self[x] + val[x] for x in self})
+        """In-place element-wise addition of two arrays, or of an array or scalar."""
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            for x in self:
+                self[x] += val[x]
+        elif np.isscalar(val):
+            for x in self:
+                self[x] += val
+        else:
+            raise TypeError('A KeyedArray can only add a scalar or a KeyedArray.')
         return self
 
     def __sub__(self, val):
-        """Element-wise subtraction of two arrays."""
-        self._assert_same_keys(val)
+        """Element-wise subtraction of two arrays, or of an array and a scalar."""
         k = KeyedArray(keys=self.keys)
-        k.update({x: self[x] - val[x] for x in self})
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            k.update({x: self[x] - val[x] for x in self})
+        elif np.isscalar(val):
+            k.update({x: self[x] - val for x in self})
+        else:
+            raise TypeError('A KeyedArray can only subtract a scalar or a KeyedArray.')
+        return k
+
+    def __rsub__(self, val):
+        """Element-wise subtraction of a scalar and an array."""
+        k = KeyedArray(keys=self.keys)
+        if np.isscalar(val):
+            k.update({x: val - self[x] for x in self})
+        else:
+            raise TypeError('A KeyedArray can only subtract a scalar or a KeyedArray.')
         return k
 
     def __isub__(self, val):
-        """Iterative element-wise subtraction of two arrays."""
-        self._assert_same_keys(val)
-        self.update({x: self[x] - val[x] for x in self})
+        """In-place element-wise subtraction of two arrays."""
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            for x in self:
+                self[x] -= val[x]
+        elif np.isscalar(val):
+            for x in self:
+                self[x] -= val
+        else:
+            raise TypeError('A KeyedArray can only subtract a scalar or a KeyedArray.')
         return self
 
     def __mul__(self, val):
-        """Element-wise multiplication of an array by a scalar."""
-        self._assert_scalar(val)
+        """Element-wise multiplication of two arrays, or of an array and a scalar."""
         k = KeyedArray(keys=self.keys)
-        k.update({x: self[x]*val for x in self})
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            k.update({x: self[x]*val[x] for x in self})
+        elif np.isscalar(val):
+            k.update({x: self[x]*val for x in self})
+        else:
+            raise TypeError('A KeyedArray can only multiply a scalar or a KeyedArray.')
         return k
 
     def __rmul__(self, val):
         """Element-wise multiplication of a scalar by an array."""
-        self._assert_scalar(val)
         k = KeyedArray(keys=self.keys)
-        k.update({x: val*self[x] for x in self})
+        if np.isscalar(val):
+            k.update({x: val*self[x] for x in self})
+        else:
+            raise TypeError('A KeyedArray can only multiply a scalar or a KeyedArray.')
         return k
 
     def __imul__(self, val):
-        """Iterative element-wise multiplication of an array by a scalar."""
-        self._assert_scalar(val)
-        self.update({x: self[x]*val for x in self})
+        """In-place element-wise multiplication of two arrays, or of an array by a scalar."""
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            for x in self:
+                self[x] *= val[x]
+        elif np.isscalar(val):
+            for x in self:
+                self[x] *= val
+        else:
+            raise TypeError('A KeyedArray can only multiply a scalar or a KeyedArray.')
         return self
 
     def __truediv__(self, val):
-        """Element-wise division of an array by a scalar."""
-        self._assert_scalar(val)
+        """Element-wise division of two arrays, or of an array by a scalar."""
         k = KeyedArray(keys=self.keys)
-        k.update({x: self[x]/val for x in self})
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            k.update({x: self[x]/val[x] for x in self})
+        elif np.isscalar(val):
+            k.update({x: self[x]/val for x in self})
+        else:
+            raise TypeError('A KeyedArray can only divide a scalar or a KeyedArray.')
+        return k
+
+    def __rtruediv__(self, val):
+        """Element-wise division of a scalar by an array."""
+        k = KeyedArray(keys=self.keys)
+        if np.isscalar(val):
+            k.update({x: val/self[x] for x in self})
+        else:
+            raise TypeError('A KeyedArray can only divide a scalar or a KeyedArray.')
         return k
 
     def __itruediv__(self, val):
-        """Iterative element-wise division of an array by a scalar."""
-        self._assert_scalar(val)
-        self.update({x: self[x]/val for x in self})
+        """In-place element-wise division of two arrays, or of an array by a scalar."""
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            for x in self:
+                self[x] /= val[x]
+        elif np.isscalar(val):
+            for x in self:
+                self[x] /= val
+        else:
+            raise TypeError('A KeyedArray can only divide a scalar or a KeyedArray.')
         return self
 
     def __neg__(self):
--- a/relentless/_collections.py
+++ b/relentless/_collections.py
@@ -391,7 +391,7 @@ class KeyedArray(FixedKeyDict):
         return k
 
     def __isub__(self, val):
-        """In-place element-wise subtraction of two arrays."""
+        """In-place element-wise subtraction of two arrays, or of an array and a scalar."""
         if isinstance(val, KeyedArray):
             self._assert_same_keys(val)
             for x in self:
@@ -478,14 +478,14 @@ class KeyedArray(FixedKeyDict):
         return k
 
     def norm(self):
-        r"""Vector norm.
+        r"""Vector :math:`ell^2`-norm.
 
         For a vector :math:`\mathbf{x}=\left[x_1,\ldots,x_n\right]`, the
         Euclidean 2-norm :math:`\lVert\mathbf{x}\rVert` is computed as:
 
-            .. math::
+        .. math::
 
-                \lVert\mathbf{x}\rVert = \sqrt{\sum_{k=1}^{n} {x_k}^2}
+            \lVert\mathbf{x}\rVert = \sqrt{\sum_{k=1}^{n} {x_k}^2}
 
         Returns
         -------
@@ -502,9 +502,9 @@ class KeyedArray(FixedKeyDict):
         :math:`\mathbf{y}=\left[y_1,\ldots,y_n\right]`, the vector dot product
         :math:`\mathbf{x}\cdot\mathbf{y}` is computed as:
 
-            .. math::
+        .. math::
 
-                \mathbf{x}\cdot\mathbf{y} = \sum_{k=1}^{n} {x_k y_k}
+            \mathbf{x}\cdot\mathbf{y} = \sum_{k=1}^{n} {x_k y_k}
 
         Parameters
         ----------
--- a/relentless/_collections.py
+++ b/relentless/_collections.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 class FixedKeyDict:
     """Dictionary with fixed keys.
 
@@ -5,8 +7,8 @@ class FixedKeyDict:
     ----------
     keys : array_like
         List of keys to be fixed.
-    default
-        Initial values to fill in the dictionary, defaults to `None`.
+    default : scalar
+        Initial value to fill in the dictionary, defaults to `None`.
 
     Examples
     --------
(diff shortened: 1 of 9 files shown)


relentless/optimize/method.py
```diff
@@ -204,3 +204,56 @@ def max_iter(self, value):
...
+        super().__init__(abs_tol, max_step_size, max_iter)
+
+    def optimize(self, objective):
+        sd = super(abs_tol=self.abs_tol, step_size=self.max_step_size, max_iter=1)
+
+        iter_num = 0
+        converged = False
+        while not converged and iter_num < self.max_iter:
+            converged = sd.optimize(objective)
+            sd.step_size = self._update_step(objective)
```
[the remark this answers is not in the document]
mphoward: I agree that this is a bit of a strange pattern. I would prefer to have a separate `LineSearch` object, which can then be consumed by multiple classes. For example, `SteepestDescent` could take accept `LineSearch` as its `step_size` arg. Then in that optimize method, do something like this:
```py
while not converged and iter_num < self.max_iter:
    res = objective.compute()
    try:
        step = self.step_size.find(objective,res.value,res.gradient)
    except AttributeError:
        step = self.step_size
    for x in dvars:
        x.value -= step*res.gradient(x)
    converged = self.has_converged(res)
    iter_num += 1
```
By doing this, you isolate the logic of the `LineSearch` from the optimizaton method, and you can reuse it in multiple classes.

Zach can provide the best feedback on what arguments `find` would need.

One other small thing: I'm noticing that I missed in previous review that we agreed that you should only update the variable values if the algorithm has not converged yet. Right now, this always updates then checks convergence. We should make a note to fix this in this PR, or open a separate issue to remind ourselves.

relentless/optimize/method.py
```diff
@@ -204,3 +204,56 @@ def max_iter(self, value):
...
+        iter_num = 0
+        step_size = self.max_step_size
+        step_interval = np.array([0, self.max_step_size])
+        target_interval = np.array([target, target])
+        while np.abs(target) > self.ls_tol and iter_num < self.ls_max_iter:
+            step_size = ((step_interval[0]*target_interval[1] - step_interval[1]*target_interval[0])
+                        /(target_interval[1] - target_interval[0]))
+
+            if step_size < step_interval[0] or step_size > step_interval[1]:
+                step_size = 0.5*(step_interval[0] + step_interval[1])
```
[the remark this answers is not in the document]
mphoward: You had a good reason, but I don't remember know either!

relentless/optimize/method.py
```diff
@@ -179,17 +187,34 @@ def optimize(self, objective):
...
             if not converged:
-                if isinstance(self.step_size, LineSearch):
-                    step = self.step_size.find(objective=objective, result_0=res)
-                else:
-                    step = self.step_size
+                if self.line_search is not None:
+                    for x in dvars:
+                        x.value = res.design_variables[x] - self.step_size*res.gradient(x)
+                    res_0 = objective.compute()
+                    res = self.line_search.find(objective=objective, res_start=res_0, res_end=res)
```
zeesherman, line 196: So the default behavior for steepest descent + line search will have a search interval that changes magnitude throughout the optimization (the interval will be large for early iterations and small for late iterations). That is absolutely fine with me, but I think the more common use case of steepest descent + line search will be to have the search intervals all have the same magnitude. This'll let you take larger steps in the late iterations without being limited by numerical stability at early iterations (which I think will win out performance-wise).

I know we've talked about this a ton and I don't want to belabor the point, just want to make sure we all agree that this is the default behavior we want!

relentless/optimize/method.py
```diff
@@ -206,81 +231,81 @@ def step_size(self, value):
...
+        res = objective.compute()
 
         # compute initial and post-adjustment target values
         target_0 = 0
         target = 0
-        for x in dvars:
-            target_0 += (d[x]*-result_0.gradient(x))
-            target += (d[x]*-result.gradient(x))
+        for x in ovars:
+            target_0 += (d[x]*-res_0.gradient(x))
```
zeesherman, line 270: FYI, if we want to use a relative tolerance, we would base it on this quantity, i.e. `np.abs(target) < rel_tol*np.abs(target_0)`

relentless/optimize/method.py
```diff
@@ -206,81 +231,81 @@ def step_size(self, value):
...
+                x.value = ovars[x]
+            return res
 
         # iterate to minimize target
         else:
-            rstep = np.array([0, self.max_step_size])
+            rstep = np.array([0, max_step])
             rtarget = np.array([target_0, target])
             iter_num = 0
             while np.abs(target) > self.abs_tol and iter_num < self.max_iter:
```
zeesherman: Just want to confirm if we want to count the first step at the maximum step size as an iteration or not. Right now, if `self.max_iter = 1`, we would take the initial maximum step, enter the while loop, and then take an additional step somewhere in the interval. To me, that kind of seems like 2 iterations.

Currently, pure steepest descent would be equivalent to steepest descent + line search (0 max iterations). If we would like pure steepest descent to be equivalent to steepest descent + line search (1 max iteration), we should probably initialize `iter_num = 1`.

relentless/optimize/method.py
```diff
@@ -174,27 +186,145 @@ def optimize(self, objective):
...
+            iter_num = 0
+            while np.abs(target) > self.abs_tol and iter_num < self.max_iter:
+                # linear interpolation for step size
+                step = (rstep[0]*rtarget[1] - rstep[1]*rtarget[0])/(rtarget[1] - rtarget[0])
+
+                # adjust variables based on new step size, compute target
+                for x in ovars:
+                    x.value = ovars[x] + step*d[x]
+                res = objective.compute()
+                target = 0
```
mphoward, line 292: Call this like `new_target` or `cur_target` or something to distinguish it from all the other ones above.

relentless/optimize/method.py
```diff
@@ -174,27 +186,145 @@ def optimize(self, objective):
...
+        if target > 0 or np.abs(target) < self.abs_tol: #combine conditions as (target > -self.abs_tol) ?
+            for x in ovars:
+                x.value = ovars[x]
+            return res
+
+        # iterate to minimize target
+        else:
+            rstep = np.array([0, max_step])
+            rtarget = np.array([target_0, target])
+            iter_num = 0
```
mphoward, line 283: I think you also want to keep the value of the `result` at either end of this interval. That way, you can return the one that is your converged result.

relentless/optimize/method.py
```diff
@@ -174,27 +186,145 @@ def optimize(self, objective):
...
+                # update search intervals
+                if target > 0:
+                    rstep[0] = step
+                    rtarget[0] = target
+                else:
+                    rstep[1] = step
+                    rtarget[1] = target
+
+                iter_num += 1
+
```
mphoward, line 305: Set `result = results[1]` (if you stored the results in `results`)

relentless/optimize/method.py
```diff
@@ -174,27 +186,145 @@ def optimize(self, objective):
...
+            for x in ovars:
+                x.value = ovars[x]
+            return res
```
mphoward, line 308: Move these out one indent

mphoward (changes requested):

relentless/optimize/method.py
```diff
@@ -174,27 +186,145 @@ def optimize(self, objective):
...
+
+        # compute initial and post-adjustment target values
+        target_0 = 0
+        target = 0
+        for x in ovars:
+            target_0 += (d[x]*-res_0.gradient(x))
+            target += (d[x]*-res.gradient(x))
+
+        # check if max step size is acceptable
+        if target > 0 or np.abs(target) < self.abs_tol: #combine conditions as (target > -self.abs_tol) ?
```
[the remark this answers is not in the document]
adisr04: At least when we defined the search direction based on the gradient, `target_0` was always positive. What do you think about this @zeesherman?

relentless/_collections.py
```diff
@@ -266,3 +268,171 @@ def __str__(self):
...
+        """Element-wise addition of two arrays."""
+        self._assert_same_keys(val)
+        k = KeyedArray(keys=self.keys)
+        k.update({x: self[x] + val[x] for x in self})
+        return k
+
+    def __iadd__(self, val):
+        """Iterative element-wise addition of two arrays."""
+        self._assert_same_keys(val)
+        self.update({x: self[x] + val[x] for x in self})
```
mphoward, line 342: This works, but the intention of this method might be clearer as:
```py
for x in self:
    self[x] += val[x]
```

relentless/_collections.py
```diff
@@ -266,3 +268,171 @@ def __str__(self):
...
+
+    def __add__(self, val):
+        """Element-wise addition of two arrays."""
+        self._assert_same_keys(val)
+        k = KeyedArray(keys=self.keys)
+        k.update({x: self[x] + val[x] for x in self})
+        return k
+
+    def __iadd__(self, val):
+        """Iterative element-wise addition of two arrays."""
```
mphoward, line 340:
```suggestion
        """In-place element-wise addition of two arrays."""
```

relentless/_collections.py
```diff
@@ -266,3 +268,171 @@ def __str__(self):
...
+
+    def __sub__(self, val):
+        """Element-wise subtraction of two arrays."""
+        self._assert_same_keys(val)
+        k = KeyedArray(keys=self.keys)
+        k.update({x: self[x] - val[x] for x in self})
+        return k
+
+    def __isub__(self, val):
+        """Iterative element-wise subtraction of two arrays."""
```
mphoward, line 353: Same comments as for `__iadd__`.

relentless/_collections.py
```diff
@@ -266,3 +268,171 @@ def __str__(self):
...
+    def _assert_scalar(self, val):
+        if not np.isscalar(val):
+            raise TypeError('''A scalar is required for multiplication or
+                               division with a KeyedArray.''')
+
+    def __add__(self, val):
+        """Element-wise addition of two arrays."""
+        self._assert_same_keys(val)
+        k = KeyedArray(keys=self.keys)
+        k.update({x: self[x] + val[x] for x in self})
```
mphoward, line 336: For all of these methods, you might check if `val` is a `KeyedArray`, and if so, do element-wise addition. Otherwise, check if it is a scalar and just add that value. Otherwise, it is an error:
```py
if isinstance(val, KeyedArray):
    # element-wise
elif np.isscalar(val):
    # work as scalar
else:
    raise TypeError()
```

relentless/optimize/method.py
```diff
@@ -171,30 +183,139 @@ def optimize(self, objective):
...
-            res = objective.compute()
-            for x in dvars:
-                x.value -= self.step_size*res.gradient(x)
-            converged = self.has_converged(res)
+        cur_res = objective.compute()
+        while not self.has_converged(cur_res) and iter_num < self.max_iter:
+
+            #steepest descent update
+            for x in dvars: #is there any way to make this into a single line?
+                x.value = cur_res.design_variables[x] - self.step_size*cur_res.gradient[x]
```
mphoward, line 191: The design variables are not a KeyedArray, so this cannot be done in one line.

Remove the extra whitespace above though. :)

relentless/optimize/method.py
```diff
@@ -171,30 +183,139 @@ def optimize(self, objective):
...
+        """
+        ovars = {x: x.value for x in objective.design_variables()}
+
+        # compute normalized search direction
+        d = end.design_variables - start.design_variables
+        max_step = d.norm()
+        d /= max_step
+
+        # compute start and end target values
+        targets = np.array([d.dot(-start.gradient), d.dot(-end.gradient)])
```
mphoward, line 262: For a minor performance difference, but the minus sign outside of the dot. This avoids making a copy of the gradient array before dotting.
```suggestion
        targets = np.array([-d.dot(start.gradient), -d.dot(end.gradient)])
```

relentless/optimize/method.py
```diff
@@ -171,30 +183,139 @@ def optimize(self, objective):
...
+
+        # check if max step size acceptable
+        if targets[1] > 0 or np.abs(targets[1]) < self.abs_tol:
+            result = end
+
```
mphoward, line 267: No whitespace between `if-else`. Put the comment inside each block.

relentless/optimize/method.py
```diff
@@ -171,30 +183,139 @@ def optimize(self, objective):
...
+        # iterate to minimize target
+        else:
+            steps = np.array([0, max_step])
+            iter_num = 0
+            new_target = self.abs_tol
+            print(steps)
+            print(targets)
```
mphoward, line 274: Stray prints, remove.

relentless/optimize/method.py
```diff
@@ -171,30 +183,139 @@ def optimize(self, objective):
...
+            while np.abs(new_target) >= self.abs_tol and iter_num < self.max_iter:
+                # linear interpolation for step size
+                new_step = (steps[0]*targets[1] - steps[1]*targets[0])/(targets[1] - targets[0])
+
+                # adjust variables based on new step size, compute target
+                for x in ovars:
+                    x.value = ovars[x] + new_step*d[x]
+                new_res = objective.compute()
+                new_target = d.dot(-new_res.gradient)
```
mphoward, line 283:
```suggestion
                new_target = -d.dot(new_res.gradient)
```

tests/optimize/test_method.py
```diff
@@ -52,5 +66,66 @@ def test_run(self):
...
+        l = relentless.optimize.LineSearch(abs_tol=1e-8, max_iter=1000)
+        x = relentless.variable.DesignVariable(value=-3.0)
+        q = QuadraticObjective(x=x)
+        res_1 = q.compute()
+
+        #bracketing the minimum (find step size that takes function to minimum)
+        x.value = 3.0
+        res_2 = q.compute()
+        x.value = -3.0
+#        res_new = l.find(objective=q, start=res_1, end=res_2)
```
mphoward, line 113: Why are these commented out?

relentless/optimize/method.py
```diff
@@ -171,30 +183,139 @@ def optimize(self, objective):
...
+            new_target = self.abs_tol
+            print(steps)
+            print(targets)
+            while np.abs(new_target) >= self.abs_tol and iter_num < self.max_iter:
+                # linear interpolation for step size
+                new_step = (steps[0]*targets[1] - steps[1]*targets[0])/(targets[1] - targets[0])
+
+                # adjust variables based on new step size, compute target
+                for x in ovars:
+                    x.value = ovars[x] + new_step*d[x]
```
mphoward, line 281: I think this should be:
```suggestion
                    x.value = start.design_variables[x] + new_step*d[x]
```

tests/test_collections.py
```diff
@@ -263,5 +263,96 @@ def test_iteration(self):
         self.assertEqual(m['A','A'], {'energy':3.0, 'mass':2.0})
         self.assertEqual(m['B','B'], {'energy':3.0, 'mass':2.0})
 
+class test_KeyedArray(unittest.TestCase):
```
mphoward, line 266: These all look good, but you might need to add a couple more if you want to support both element-wise and scalar arithmetic in each op.

mphoward (changes requested):
This is looking great! I have a couple more suggestions, and then I'd like Zach to review one more time.

relentless/_collections.py
```diff
@@ -266,3 +268,254 @@ def __str__(self):
...
+        if isinstance(val, KeyedArray):
+            self._assert_same_keys(val)
+            k.update({x: self[x] + val[x] for x in self})
+        elif np.isscalar(val):
+            k.update({x: self[x] + val for x in self})
+        else:
+            raise TypeError('A KeyedArray can only add a scalar or a KeyedArray.')
+        return k
+
+    def __radd__(self, val):
```
mphoward, line 350: Just to check, you only need to implement `__radd__` for the case where the left `val` is a scalar because `__add__` will be invoked if the `val` is a `KeyedArray`?

relentless/_collections.py
```diff
@@ -266,3 +268,254 @@ def __str__(self):
...
+        """Element-wise subtraction of a scalar and an array."""
+        k = KeyedArray(keys=self.keys)
+        if np.isscalar(val):
+            k.update({x: val - self[x] for x in self})
+        else:
+            raise TypeError('A KeyedArray can only subtract a scalar or a KeyedArray.')
+        return k
+
+    def __isub__(self, val):
+        """In-place element-wise subtraction of two arrays."""
```
mphoward, line 394:
```suggestion
        """In-place element-wise subtraction of two arrays, or of an array and a scalar."""
```

relentless/_collections.py
```diff
@@ -266,3 +268,254 @@ def __str__(self):
...
+        return self
+
+    def __neg__(self):
+        """Element-wise negation of an array."""
+        k = KeyedArray(keys=self.keys)
+        k.update({x: -self[x] for x in self})
+        return k
+
+    def norm(self):
+        r"""Vector norm.
```
mphoward, line 481:
```suggestion
        r"""Vector :math:`\ell^2`-norm.
```

relentless/_collections.py
```diff
@@ -266,3 +268,254 @@ def __str__(self):
...
+    def dot(self, val):
+        r"""Vector dot product.
+
+        For two vectors :math:`\mathbf{x}=\left[x_1,\ldots,x_n\right]` and
+        :math:`\mathbf{y}=\left[y_1,\ldots,y_n\right]`, the vector dot product
+        :math:`\mathbf{x}\cdot\mathbf{y}` is computed as:
+
+            .. math::
+
+                \mathbf{x}\cdot\mathbf{y} = \sum_{k=1}^{n} {x_k y_k}
```
mphoward, line 507: Pull this out one indentation.

relentless/_collections.py
```diff
@@ -266,3 +268,254 @@ def __str__(self):
...
+
+    def norm(self):
+        r"""Vector norm.
+
+        For a vector :math:`\mathbf{x}=\left[x_1,\ldots,x_n\right]`, the
+        Euclidean 2-norm :math:`\lVert\mathbf{x}\rVert` is computed as:
+
+            .. math::
+
+                \lVert\mathbf{x}\rVert = \sqrt{\sum_{k=1}^{n} {x_k}^2}
```
mphoward, line 488: Pull this out one indentation.

relentless/optimize/method.py
```diff
@@ -121,11 +131,144 @@ def abs_tol(self, value):
         else:
             self._abs_tol = abs_tol
 
+class LineSearch:
+    r"""Line search algorithm.
+
+    For an :class:`~relentless.optimize.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
+    the line search algorithm seeks to take a single, optimally-sized step towards
+    the minimum of the function. (Does not necessarily converge to minimum.)
```
mphoward, line 139:
```suggestion
    the line search algorithm seeks to take a single, optimally-sized step along
    a direction.
```

relentless/optimize/method.py
```diff
@@ -121,11 +131,144 @@ def abs_tol(self, value):
...
+
+    A search interval :math:`\mathbf{d}` is specified, and normalized to :math:`\hat{\mathbf{d}}`:
+
+    .. math::
+
+        \mathbf{d} = \mathbf{x}_{end} - \mathbf{x}_{start}
+
+        \hat{\mathbf{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
+
+    Then, a 'target' value :math:`t` is defined as:
```
mphoward, line 149:
```suggestion
    Then, a "target" value :math:`t` is defined as:
```

relentless/optimize/method.py
```diff
@@ -121,11 +131,144 @@ def abs_tol(self, value):
...
+
+    .. math::
+
+        t = -\hat{\mathbf{d}} \cdot \nabla{f\left(\mathbf{x}\right)}
+
+    The target at the start of the search interval is always negative. If the
```
mphoward, line 155: Wait, I've got myself confused now... if d points from start to end, then for steepest descent, it points in the direction of -grad(f). This would mean t > 0 at the start of the search interval?

I feel like these signs don't match the code below.

relentless/optimize/method.py
```diff
@@ -121,11 +131,144 @@ def abs_tol(self, value):
...
+    target is negative (or within the tolerance) at the end of the search
+    interval, then the maximum step size is acceptable and the algorithm steps
+    to the end of the search interval. If the target is positive (outside of the
+    tolerance) at the end of the search interval, then the algorithm iteratively
+    computes a new step size by linear interpolation within the search interval
+    until the target at the new location is minimized to within the tolerance.
+
```
mphoward, line 162: I think we might also want to add a short note saying that this is applying the strong Wolfe criterion and link the wiki page, like we discussed on slack. We might also mention that we are not doing anything with criteria one.

relentless/optimize/method.py
```diff
@@ -121,11 +131,144 @@ def abs_tol(self, value):
...
+        # compute start and end target values
+        targets = np.array([-d.dot(start.gradient), -d.dot(end.gradient)])
+
+        # check if max step size acceptable, else iterate to minimize target
+        if targets[1] > 0 or np.abs(targets[1]) < self.abs_tol:
+            result = end
+        else:
+            steps = np.array([0, max_step])
+            iter_num = 0
+            new_target = self.abs_tol
```
mphoward, line 212: I see what you're doing... better set it to something like `np.inf` so the intention is clear.

Maybe also initialize `new_res = targets[1]` here? That way if the loop is never entered, `new_res` is still defined.

mphoward (approved):
I have only a few minor points (one main one in the documentation) that need addressing, and then I think this is good to go once Zach looks it over!

relentless/optimize/method.py
zeesherman: Yeah, Mike is correct. d is a descent direction, so the way we have defined t, t > 0 means we decrease the function by moving in the direction of d. t < 0 means we increase the function by moving opposite the direction of d. t will always be positive at the start point (if it wasn't d would be a very lousy choice for a search direction!)

zeesherman, line 155:
```suggestion
    Because :math:`\hat{\mathbf{d}}`: is a descent direction, the target at the start of the search interval is always positive. If the
```

relentless/optimize/method.py
```diff
@@ -121,11 +131,144 @@ def abs_tol(self, value):
...
+        \hat{\mathbf{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
+
+    Then, a 'target' value :math:`t` is defined as:
+
+    .. math::
+
+        t = -\hat{\mathbf{d}} \cdot \nabla{f\left(\mathbf{x}\right)}
+
+    The target at the start of the search interval is always negative. If the
+    target is negative (or within the tolerance) at the end of the search
```
zeesherman, line 156:
```suggestion
    target is positive (or within the tolerance) at the end of the search
```

relentless/optimize/method.py
```diff
@@ -121,11 +131,144 @@ def abs_tol(self, value):
...
+    interval, then the maximum step size is acceptable and the algorithm steps
+    to the end of the search interval. If the target is positive (outside of the
```
zeesherman, line 158:
```suggestion
    to the end of the search interval. If the target is negative (outside of the
```

relentless/_collections.py
adisr04: Yes!

PR #54, opened 2021-03-28
Implement fixed-step steepest descent algorithm

adisr04:
- Add `FixedStepDescent` class with documentation in `method.py`
- Add `FixedStepDescent` unit tests in `test_method.py`
- Separate step size and scaling parameters in `SteepestDescent` 
- Add `descent_amount()` method to `SteepestDescent`
- Modify `LineSearch` to accept relative tolerance
- Modify `LineSearch.find()` to return a normalized step size
- Add exponentiation functionality to `KeyedArray` in `_collections.py`


relentless/optimize/__init__.py
```diff
@@ -10,7 +10,8 @@
 
 """
 from .method import (Optimizer,
+                     LineSearch,
                      SteepestDescent,
-                     LineSearch)
+                     FixedStepDescent)
 
-from .objective import (ObjectiveFunction)
+from .objective import (ObjectiveFunction)
```
mphoward, line 17: Minor formatting things: import the classes from `method` in alphabetical order, and restore the missing newline character at the end of this file (that's what the red circle here means).

relentless/optimize/method.py
```diff
@@ -314,35 +323,78 @@ def __init__(self, abs_tol, max_iter, step_size, line_search=None):
...
+        update = alpha*cur_res.gradient/cur_res.gradient.norm()
+
         while not self.has_converged(cur_res) and iter_num < self.max_iter:
             #steepest descent update
             for x in dvars:
-                x.value = cur_res.design_variables[x] - self.step_size*cur_res.gradient[x]
+                x.value = cur_res.design_variables[x] - update[x]
             next_res = objective.compute()
+            alpha = self.descent_amount(cur_res.gradient)
+            update = alpha*cur_res.gradient/cur_res.gradient.norm()
```
mphoward, line 397: Can't you put this before the `for x in dvars` loop, and remove lines 388-390?

relentless/optimize/objective.py
```diff
@@ -6,7 +6,7 @@
 by adjusting the variables on which the function depends.
 
 This function, :math:`f`, is a scalar value that is defined as a function of :math:`n`
-problem :class:`DesignVariables<~relentless.variable.DesignVariable>`
+problem :class:`DesignVariables<~relentless.variable.DesignVariable>`                           #this doesn't make a hyperlink, do you know how to fix this?
```
mphoward, line 9: I'll need to investigate. I thought this pattern worked when I tried it before... the goal is to get it to show `DesignVariables` (plural) but link to `DesignVariable` (singular). If nothing else, you can replace it with a regular reference to `DesignVariable` and we can ignore the pluralization.

mphoward (changes requested):
This overall looks like it should work nicely. My only big question is about whether we really should convert `step_size` to a dict, or if we should leave it as a scalar and accept a separate dict of `normalization` for the design variables. Let's discuss on slack.

relentless/optimize/method.py
```diff
@@ -314,35 +323,78 @@ def __init__(self, abs_tol, max_iter, step_size, line_search=None):
...
         self.line_search = line_search
 
+    def descent_amount(self, gradient):
+        r"""Calculate the descent amount for the optimization.
+
+        The descent amount is:
+
+        .. math::
+
+            d = \alpha\lVert\nabla f\left(\mathbf{x}\right)\rVert
```
zeesherman, line 333: Do we want to use a different variable here since we have been using d for the step direction?

relentless/optimize/method.py
```diff
@@ -314,35 +323,78 @@ def __init__(self, abs_tol, max_iter, step_size, line_search=None):
...
     def optimize(self, objective):
-        """Perform the steepest descent optimization for the given objective function.
+        r"""Perform the steepest descent optimization for the given objective function.
+
+        A single steepest descent update is performed as:
+
+        .. math::
+
+            \mathbf{x}_{n+1} = \mathbf{x}_{n}-d\frac{\nabla f\left(\mathbf{x}\right)}
+                                                    {\lVert\nabla f\left(\mathbf{x}\right)\rVert}
```
zeesherman, line 362:
```suggestion
            \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha \nabla f\left(\mathbf{x}\right)
```

relentless/optimize/method.py
```diff
@@ -314,35 +323,78 @@ def __init__(self, abs_tol, max_iter, step_size, line_search=None):
...
+
+        :math:`d` is the descent amount returned by :meth:`descent_amount()`.
```
zeesherman, line 364: I'd prefer the `alpha*grad` version of the update rather than `d*grad/norm`. The latter looks like fixed step descent unless you knew that` d = alpha*norm`. And this is consistent with the documentation for the class.

relentless/optimize/method.py
```diff
@@ -177,41 +193,67 @@ def __init__(self, abs_tol, max_iter):
         self.abs_tol = abs_tol
         self.max_iter = max_iter
 
-    def find(self, objective, start, end):
+    def find(self, objective, start, end, grad_scale=1.0):
```
mphoward, line 196: I would prefer to call this parameter `scale` rather than `grad_scale`. Really, you are rescaling the **variables**, and the consequence is seen in the gradient.

relentless/optimize/method.py
```diff
@@ -145,15 +153,23 @@ class LineSearch:
...
+        \nabla f\left(\mathbf{\hat{x}}\right) = \left[{b_1}^2 \frac{\partial f}{\partial x_1},
+                                                      \cdots,
+                                                      {b_n}^2 \frac{\partial f}{\partial x_n}\right]
 
-    Because :math:`\hat{\mathbf{d}}` is a descent direction, the target at the
+        t = -\mathbf{\hat{d}} \cdot \nabla{f\left(\mathbf{\hat{x}}\right)}
+
+    :math:`\mathbf{b}=\left[b_1,\ldots,b_n\right]` is defined as the scaling
+    parameters for the gradient such that :math:`\hat{x}_i=\frac{x_i}{b_i}`. By
+    default, the gradient is left unscaled (:math:`b_i=1`).
```
mphoward, line 170: I think we should write this more clearly upfront. The variables are _scaled_, then the optimal step is chosen along the unit vector pointing between the _scaled_ variables.

relentless/optimize/method.py
```diff
@@ -145,15 +153,23 @@ class LineSearch:
...
 
-        \hat{\mathbf{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
+        \mathbf{\hat{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
 
     Then, a "target" value :math:`t` is defined as:
 
     .. math::
 
-        t = -\hat{\mathbf{d}} \cdot \nabla{f\left(\mathbf{x}\right)}
+        \nabla f\left(\mathbf{\hat{x}}\right) = \left[{b_1}^2 \frac{\partial f}{\partial x_1},
```
mphoward, line 162: Can we use a different symbol for the scale parameters that attaches a little more meaning? Maybe capital X? I also think we shouldn't use `\hat` to indicate the scaled variables, since it also is used to indicate the unit vector. The result of scaling the variables is not necessarily unity. I would maybe suggest `y = x/X`? Then you can also write like `\nabla_y` and `\nabla_x` pretty cleanly.

relentless/optimize/method.py
```diff
@@ -177,41 +193,67 @@ def __init__(self, abs_tol, max_iter):
...
             The objective function evaluated at the end of the search interval.
+        grad_scale : float or dict or :class:`~relentless._collections.KeyedArray`
+            The gradient scaling parameter or parameters (:math:`\mathbf{b}`)
+            keyed on the :class:`~relentless.optimize.objective.ObjectiveFunction`
+            design variables (defaults to ``1.0``).
 
         Raises
         ------
+        ValueError
+            If the gradient scaling parameters are not positive.
```
mphoward, line 218: @zeesherman Do they need to be positive? It seems like a negative value would be allowable, even if it's silly.

relentless/optimize/method.py
```diff
@@ -177,41 +193,67 @@ def __init__(self, abs_tol, max_iter):
...
+                err = grad_scale <= 0
+        if err:
+            raise ValueError('The gradient scaling parameters must be positive.')
+
         ovars = {x: x.value for x in objective.design_variables()}
 
         # compute normalized search direction
         d = end.design_variables - start.design_variables
         max_step = d.norm()
+        if max_step == 0:
```
mphoward, line 249: The case you're guarding against should return exactly equal, but I would use `isclose` since this is a floating-point value.
```suggestion
        if np.isclose(max_step,0):
```

relentless/optimize/method.py
```diff
@@ -280,57 +323,105 @@ def max_iter(self, value):
...
-        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\nabla f\left(\mathbf{x}\right)
+        \nabla f\left(\mathbf{\hat{x}}\right) = \left[{b_1}^2 \frac{\partial f}{\partial x_1},
+                                                      \cdots,
+                                                      {b_n}^2 \frac{\partial f}{\partial x_n}\right]
 
-    where :math:`\alpha` is defined as the step size hyperparameter. The step size
-    must be specified numerically, and a :class:`LineSearch` can be performed to find
+        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\nabla f\left(\mathbf{\hat{x}}\right)
+
+    :math:`\mathbf{b}=\left[b_1,\ldots,b_n\right]` is defined as the scaling
```
mphoward, line 340: Same comments about the docs here as for the line search.

relentless/optimize/method.py
```diff
@@ -280,57 +323,105 @@ def max_iter(self, value):
...
+
+    :math:`\alpha` is defined as the step size hyperparameter. The step size must
+    be specified numerically, and a :class:`LineSearch` can be performed to find
     an optimal step size value if desired.
 
+    Note that this update is equivalent to:
+
+    .. math::
+
+        \mathbf{x}_{n+1} = \mathbf{x}_{n}-c\frac{\nabla f\left(\mathbf{\hat{x}}\right)}
```
mphoward, line 352: I think Zach might have mentioned this before, but I think I would absorb the normalization of the gradient into `c`. It's cleaner to read that the direction of descent is down the gradient, and it also makes the `FixedStepDescent` clearer since it is the only class that will normalize by the gradient magnitude.

relentless/optimize/method.py
```diff
@@ -280,57 +323,105 @@ def max_iter(self, value):
...
     .. math::
 
-        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\nabla f\left(\mathbf{x}\right)
+        \nabla f\left(\mathbf{\hat{x}}\right) = \left[{b_1}^2 \frac{\partial f}{\partial x_1},
+                                                      \cdots,
+                                                      {b_n}^2 \frac{\partial f}{\partial x_n}\right]
 
-    where :math:`\alpha` is defined as the step size hyperparameter. The step size
-    must be specified numerically, and a :class:`LineSearch` can be performed to find
+        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\nabla f\left(\mathbf{\hat{x}}\right)
```
mphoward, line 338: See my comment about rewriting the `\nabla_y`, and also make sure we document earlier up that the variables are scaled first. I would also add the second "equivalent" equation below here... basically, show the update first in scaled variables (y) then in actual variables (x) with the scaling parameters explicit.

We do need to be clear about which gradient the tolerance is defined on. Is it going to be `\nabla_x` or `\nabla_y`?

relentless/optimize/method.py
```diff
@@ -339,14 +430,21 @@ def optimize(self, objective):
         iter_num = 0
         cur_res = objective.compute()
         while not self.has_converged(cur_res) and iter_num < self.max_iter:
+            grad_hat = cur_res.gradient*self.grad_scale**2
+            c = self.descent_amount(grad_hat)
+            update = c*grad_hat/grad_hat.norm()
```
mphoward, line 435: Along with my comment above, I'm in favor of not normalizing the gradient here, and making the `descent_amount` method decide to do it.

relentless/optimize/method.py
```diff
@@ -372,15 +470,36 @@ def max_iter(self, value):
...
 
+    @property
+    def grad_scale(self):
+        r"""float or :class:`~relentless._collections.KeyedArray`: The gradient
+        scaling parameter or parameters (:math:`\mathbf{b}`) keyed on the design
+        variables. Must be positive."""
+        return self._grad_scale
+
+    @grad_scale.setter
+    def grad_scale(self, value):
```
mphoward, line 490: I think we want to allow `grad_scale` to be a regular-old dict. Imagine the user only wants to scale *some* but not all variables... this would cause an issue because the `KeyedArray` you generate here will not have the same keys as the design variables.

I think the solution I would use is to apply the scale parameters when you loop over the design variables. Try to get it from the dict if it exists (and `scale` is a dict), set it to 1.0 if `scale` is a dict but it's missing the key, otherwise use the scalar value of `scale`.

relentless/optimize/method.py
```diff
@@ -391,3 +510,85 @@ def line_search(self, value):
...
         self._line_search = value
+
+class FixedStepDescent(SteepestDescent):
+    r"""Fixed-step steepest descent algorithm.
+
+    For an :class:`~relentless.optimize.objective.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
+    the fixed-step steepest descent algorithm seeks to approach a minimum of the
+    function.
+
+    With an initial numerical guess for all the design variables :math:`\mathbf{x}`,
```
mphoward, line 521: Same comments on docs here.

mphoward (changes requested):

relentless/optimize/method.py
```diff
@@ -145,15 +153,33 @@ class LineSearch:
 
         \mathbf{d} = \mathbf{x}_{end} - \mathbf{x}_{start}
 
-        \hat{\mathbf{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
+        \mathbf{\hat{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
```
mphoward, line 156: There is a stray `\hat\mathbf` above this that you should flip.

relentless/optimize/method.py
```diff
@@ -145,15 +153,33 @@ class LineSearch:
...
+    parameters for the gradient such that :math:`y_i=\frac{x_i}{X_i}`. By
+    default, the variables are left unscaled (:math:`X_i=1`).
+
+    With scaling, the gradient of the function becomes:
+
+    .. math::
+
+        \nabla f\left(\mathbf{y}\right) = \left[{X_1}^2 \frac{\partial f}{\partial x_1},
+                                                \cdots,
+                                                {X_n}^2 \frac{\partial f}{\partial x_n}\right]
```
mphoward, line 168: Shouldn't there be no `^2` here?
```
df/dy = df/dx dx/dy = X df/dx
```
if `x = Xy`

relentless/optimize/method.py
```diff
@@ -145,15 +153,33 @@ class LineSearch:
...
 
-        t = -\hat{\mathbf{d}} \cdot \nabla{f\left(\mathbf{x}\right)}
+        t = -\mathbf{\hat{d}} \cdot \nabla{f\left(\mathbf{y}\right)}
+
+    which is equivalent to:
+
+    .. math::
 
-    Because :math:`\hat{\mathbf{d}}` is a descent direction, the target at the
+        t = \sum_{i=1}^{n} \left(-\hat{d}_i {X_i}^2 \frac{\partial f}{\partial x_i}\right)
```
mphoward, line 180: I see this `^2` is coming up here and it looks like in the code. @zeesherman please double check.

I think we need to be careful about what we are defining as the direction `d` and the target `t`. `d` would pick up an extra factor of X to convert between x and y, but I am not following this clearly from the docs.

relentless/optimize/method.py
```diff
@@ -177,41 +203,80 @@ def __init__(self, abs_tol, max_iter):
...
 
         Returns
         -------
-        :class:`~relentless.optimize.ObjectiveFunctionResult`
-            The objective function evaluated at the new, 'optimal' location.
+        :class:`~relentless.optimize.objective.ObjectiveFunctionResult`
+            The objective function evaluated at the new, "optimal" location.
 
         """
+        # check if valid scaling
```
mphoward, line 243: I would suggest that you coerce the scaling parameters into a `KeyedArray` here so that you can use this math later.

relentless/optimize/method.py
```diff
@@ -177,41 +203,80 @@ def __init__(self, abs_tol, max_iter):
...
 
         # compute normalized search direction
         d = end.design_variables - start.design_variables
         max_step = d.norm()
+        if max_step == 0:
+            raise ValueError('The start and end of the search interval must be different.')
         d /= max_step
 
+        # compute scaled gradients
+        grad_y_start = start.gradient
```
mphoward, line 263: I find this notation confusing. This is actually `grad_x_start`. I would rather that you left these variables in their original form, and then multiply by the scale factors where they are needed. It will make the intent a lot cleaner, and you can do any error checking once above.

relentless/optimize/method.py
```diff
@@ -177,41 +203,80 @@ def __init__(self, abs_tol, max_iter):
...
+                    grad_y_start[x] *= sc[x]**2
+                    grad_y_end[x] *= sc[x]**2
+                except KeyError:
+                    raise KeyError('''The scaling parameters cannot be defined
+                                      on a variable which is not in the objective
+                                      function.''')
+
         # compute start and end target values
-        targets = np.array([-d.dot(start.gradient), -d.dot(end.gradient)])
+        targets = np.array([-d.dot(grad_y_start), -d.dot(grad_y_end)])
```
mphoward, line 279: For example:
```suggestion
        targets = np.array([-d.dot(scale**2*start.gradient), -d.dot(scale**2*end.gradient)])
```

relentless/optimize/method.py
```diff
@@ -227,11 +292,17 @@ def find(self, objective, start, end):
                 # linear interpolation for step size
                 new_step = (steps[0]*targets[1] - steps[1]*targets[0])/(targets[1] - targets[0])
 
-                # adjust variables based on new step size, compute target
+                # adjust variables based on new step size, compute scaled gradient and new target
                 for x in ovars:
                     x.value = start.design_variables[x] + new_step*d[x]
                 new_res = objective.compute()
-                new_target = -d.dot(new_res.gradient)
+                grad_y_new = new_res.gradient
```
mphoward, line 299: Same idea here.

relentless/optimize/method.py
```diff
@@ -280,57 +351,116 @@ def max_iter(self, value):
 class SteepestDescent(Optimizer):
     r"""Steepest descent algorithm.
 
-    For an :class:`~relentless.optimize.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
+    For an :class:`~relentless.optimize.objective.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
     the steepest descent algorithm seeks to approach a minimum of the function.
 
-    With an initial numerical guess for all the design variables :math:`\mathbf{x}`,
-    the following iterative calculation is performed:
+    :math:`\mathbf{X}=\left[X_1,\ldots,X_n\right]` is defined as the scaling
```
mphoward, line 357: I think alot of the same comments will apply here, so I am going to stop reviewing for now. Let's get feedback from Zach and resolve what's going on in the LineSearch, apply some of the simplifying changes I proposed above to both classes, and then I'll take a look at this again!

mphoward (changes requested):
I left some comments for discussion. Could you also merge `develop` into this branch when you get a chance? The new commit on it will shorten your CI testing time.

relentless/optimize/objective.py
mphoward: I can't figure out how to get this to work, so convert to:
```suggestion
problem :class:`~relentless.variable.DesignVariable>`\s
```
which will print the link with an s right after it.

relentless/optimize/method.py
zeesherman: Mike is correct, only one factor of X here. Since we don't actually need the scale factors for the LineSearch, let's remove anything involving the scaled variables and scaled gradient.

relentless/optimize/method.py
```diff
@@ -145,15 +153,33 @@ class LineSearch:
 
         \mathbf{d} = \mathbf{x}_{end} - \mathbf{x}_{start}
 
-        \hat{\mathbf{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
+        \mathbf{\hat{d}} = \frac{\mathbf{d}}{\lVert\mathbf{d}\rVert}
+
+    :math:`\mathbf{X}=\left[X_1,\ldots,X_n\right]` is defined as the scaling
+    parameters for the gradient such that :math:`y_i=\frac{x_i}{X_i}`. By
+    default, the variables are left unscaled (:math:`X_i=1`).
+
```
zeesherman, line 161: Ok so I think lines 145-160 can be simplified to something like "Given an ObjectiveFunction f(x) and a search interval x_start and x_end, LineSearch seeks an optimal step 0 < alpha < 1 such that f(x_start + alpha d) is minimized, were the search direction d and hat d is...."

relentless/optimize/method.py
```diff
@@ -177,41 +203,80 @@ def __init__(self, abs_tol, max_iter):
         self.abs_tol = abs_tol
         self.max_iter = max_iter
 
-    def find(self, objective, start, end):
+    def find(self, objective, start, end, scale=1.0):
```
zeesherman, line 206: Just to reiterate, we will remove `scale` everywhere from `LineSearch`

relentless/optimize/method.py
zeesherman: Now that we are not scaling, this can just be `grad_start`

relentless/optimize/method.py
```diff
@@ -177,41 +203,80 @@ def __init__(self, abs_tol, max_iter):
...
+            grad_y_end *= sc**2
+        else:
+            for x in sc:
+                try:
+                    grad_y_start[x] *= sc[x]**2
+                    grad_y_end[x] *= sc[x]**2
+                except KeyError:
+                    raise KeyError('''The scaling parameters cannot be defined
+                                      on a variable which is not in the objective
+                                      function.''')
```
zeesherman, line 276: Can get rid of all this now that we don't need to scale.

relentless/optimize/method.py
```diff
@@ -227,11 +292,17 @@ def find(self, objective, start, end):
...
                 for x in ovars:
                     x.value = start.design_variables[x] + new_step*d[x]
                 new_res = objective.compute()
-                new_target = -d.dot(new_res.gradient)
+                grad_y_new = new_res.gradient
+                if np.isscalar(sc):
+                    grad_y_new *= sc**2
+                else:
+                    for x in sc:
+                        grad_y_new[x] *= sc[x]**2
```
zeesherman, line 304: Also can get rid of this w/o scaling.

relentless/optimize/method.py
```diff
@@ -280,57 +351,116 @@ def max_iter(self, value):
...
+    default, the variables are left unscaled (:math:`X_i=1`).
+
+    With scaling, the gradient of the function becomes:
 
     .. math::
 
-        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\nabla f\left(\mathbf{x}\right)
+        \nabla f\left(\mathbf{y}\right) = \left[{X_1}^2 \frac{\partial f}{\partial x_1},
+                                                      \cdots,
+                                                {X_n}^2 \frac{\partial f}{\partial x_n}\right]
```
zeesherman, line 367:
```suggestion
        \nabla f\left(\mathbf{y}\right) = \left[X_1 \frac{\partial f}{\partial x_1},
                                                      \cdots,
                                                X_n \frac{\partial f}{\partial x_n}\right]
```

Just one factor of `X_i` for the gradient. You don't get the second factor of `X_i` until the update equation (that one coming from the conversion between `x_i` and `y_i`.)

relentless/optimize/method.py
```diff
@@ -280,57 +351,116 @@ def max_iter(self, value):
...
+    :math:`\alpha` is defined as the step size hyperparameter. The step size must
+    be specified numerically, and a :class:`LineSearch` can be performed to find
     an optimal step size value if desired.
 
+    Then, with an initial numerical guess for all the design variables :math:`\mathbf{x}`,
+    the following iterative calculation is performed:
+
+    .. math::
+
+        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\nabla f\left(\mathbf{y}\right)
```
zeesherman, line 378:
```suggestion
        \mathbf{y}_{n+1} = \mathbf{y}_{n}-\alpha\nabla f\left(\mathbf{y}\right)
```

The optimization is done w.r.t. the scaled variables

relentless/optimize/method.py
```diff
@@ -339,14 +469,38 @@ def optimize(self, objective):
         iter_num = 0
         cur_res = objective.compute()
         while not self.has_converged(cur_res) and iter_num < self.max_iter:
+            grad_y = cur_res.gradient
+            if np.isscalar(self.scale):
+                grad_y *= self.scale**2
```
zeesherman, line 474:
```suggestion
                grad_y *= self.scale
```

Just one scale factor in the gradient.

relentless/optimize/method.py
```diff
@@ -339,14 +469,38 @@ def optimize(self, objective):
...
+            else:
+                for x in self.scale:
+                    try:
+                        grad_y[x] *= self.scale[x]**2
```
zeesherman, line 478:
```suggestion
                        grad_y[x] *= self.scale[x]
```

relentless/optimize/method.py
```diff
@@ -339,14 +469,38 @@ def optimize(self, objective):
...
+                    except KeyError:
+                        raise KeyError('''The scaling parameters cannot be defined
+                                          on a variable which is not in the objective
+                                          function.''')
+            update = self.descent_amount(grad_y)*grad_y
```
zeesherman, line 483: Here is where we will need the second factor of the `scale`. So each element of `update` needs to have something like `self.descent_amount(grad_y)*grad_y[x]*scale[x]`. There is probably a way of handling things so you don't need an extra loop here.

FYI, we can't just stuff the scale^2 into grad_y because when we do FixedStep, we will need `descent_amount` to get the correct grad_y.

relentless/optimize/method.py
```diff
@@ -391,3 +564,88 @@ def line_search(self, value):
...
+    default, the variables are left unscaled (:math:`X_i=1`).
+
+    With scaling, the gradient of the function becomes:
+
+    .. math::
+
+        \nabla f\left(\mathbf{y}\right) = \left[{X_1}^2 \frac{\partial f}{\partial x_1},
+                                                      \cdots,
+                                                {X_n}^2 \frac{\partial f}{\partial x_n}\right]
+
```
zeesherman, line 586:
```suggestion
        \nabla f\left(\mathbf{y}\right) = \left[X_1 \frac{\partial f}{\partial x_1},
                                                      \cdots,
                                                X_n \frac{\partial f}{\partial x_n}\right]

```

relentless/optimize/method.py
```diff
@@ -391,3 +564,88 @@ def line_search(self, value):
...
+    be specified numerically, and a :class:`LineSearch` can be performed to find
+    an optimal step size value if desired.
+
+    Then, with an initial numerical guess for all the design variables :math:`\mathbf{x}`,
+    the following iterative calculation is performed:
+
+    .. math::
+
+        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\frac{\nabla f\left(\mathbf{y}\right)}
+                                                     {\lVert\nabla f\left(\mathbf{y}\right)\rVert}
```
zeesherman, line 597:
```suggestion
        \mathbf{y}_{n+1} = \mathbf{y}_{n}-\alpha\frac{\nabla f\left(\mathbf{y}\right)}
                                                     {\lVert\nabla f\left(\mathbf{y}\right)\rVert}
```

zeesherman (commented):
I know you went through a bunch of work to put the scaling parameters into the LineSearch. I'm sorry that I didn't realize this was unnecessary and that the LineSearch could be done purely in terms of unscaled variables. You'll now need to remove all the scaling stuff from LineSearch. I really appreciate all your effort on this extra work I've caused you!

relentless/optimize/method.py
mphoward: Even better, you can just use `start.gradient` without assigning a new variable :-)

relentless/optimize/method.py
```diff
@@ -134,7 +142,7 @@ def abs_tol(self, value):
 class LineSearch:
     r"""Line search algorithm.
 
-    For an :class:`~relentless.optimize.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
+    For an :class:`~relentless.optimize.objective.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
     the line search algorithm seeks to take a single, optimally-sized step along
     a search interval, which must be a descent direction (a vector which points
     towards or crosses a local minimum of the objective function).
```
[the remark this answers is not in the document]
mphoward: I agree.

relentless/optimize/method.py
mphoward: I like this, I think this explains what is going on nicely.

relentless/optimize/method.py
mphoward: Indeed, you do need to compute `grad_y = scale*grad_x` for `descent_amount`. The `KeyedArray` should allow you to do the update without a loop.

adisr04: Not sure this is possible, as the `scale` doesn't necessarily have all the design variables that `gradient` has, and we can't convert the input `scale` dictionary to a `KeyedArray` as the `SteepestDescent` isn't aware of the `ObjectiveFunction` design variables at the time of its construction...

mphoward: At the beginning of `optimize`, make a local `scale` variable with the right keys:
```py
scale = KeyedArray(dvars)
for x in dvars:
    if np.isscalar(self.scale):
        scale[x] = self.scale
    else:
        try:
            scale[x] = self.scale[x]
        except KeyError:
            scale[x] = 1.0
```

relentless/optimize/method.py
```diff
@@ -11,6 +11,7 @@
     :nosignatures:
 
     SteepestDescent
+    FixedStepDescent
```
mphoward, line 14: `FixedStepDescent` should come before `SteepestDescent` alphabetically.

relentless/optimize/method.py
```diff
@@ -37,16 +38,24 @@
...
-    :members: optimize,
+    :members: descent_amount,
+        scale_grad,
+        optimize,
         max_iter,
         step_size,
+        scale,
         line_search
 
+.. autoclass:: FixedStepDescent
```
mphoward, line 49: Reorder here as well.

relentless/optimize/method.py
```diff
@@ -37,16 +38,24 @@
 
 .. autoclass:: SteepestDescent
     :member-order: bysource
-    :members: optimize,
+    :members: descent_amount,
```
mphoward, line 41: Do you need to list all these members explicitly? (i.e., are there any you are excluding?) I think you can just use `:members:` to pull everything in by default.

relentless/optimize/method.py
```diff
@@ -219,15 +226,15 @@ def find(self, objective, start, end):
         if targets[1] > 0 or np.abs(targets[1]) < self.abs_tol:
             result = end
         else:
-            steps = np.array([0, max_step])
+            steps = np.array([0,1], dtype=np.float64)
```
mphoward, line 229: Just FYI, if you put `[0.,1.]` you should get `float64` as the default type.

relentless/optimize/method.py
```diff
@@ -219,15 +226,15 @@ def find(self, objective, start, end):
...
             iter_num = 0
             new_target = np.inf
             new_res = targets[1]
             while np.abs(new_target) >= self.abs_tol and iter_num < self.max_iter:
```
mphoward, line 233: I think we said we wanted to switch the stopping criterion for the `LineSearch` to use a relative tolerance. You can compute that tolerance once (outside the loop) as:
```py
tol = self.rel_tol*np.abs(targets[0])
```
See iii) of https://en.wikipedia.org/wiki/Wolfe_conditions

This would mean you need to change the `abs_tol` property throughout, as well.

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
 
-    For an :class:`~relentless.optimize.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
+    For an :class:`~relentless.optimize.objective.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
     the steepest descent algorithm seeks to approach a minimum of the function.
 
-    With an initial numerical guess for all the design variables :math:`\mathbf{x}`,
-    the following iterative calculation is performed:
+    :math:`\mathbf{X}=\left[X_1,\ldots,X_n\right]` is defined as the scaling
+    parameters for the gradient such that :math:`y_i=\frac{x_i}{X_i}`. By
+    default, the variables are left unscaled (:math:`X_i=1`).
```
mphoward, line 295:
```suggestion
    The optimization is performed using scaled variables **y**. Define **X**
    as the scaling parameters for each variable such that :math:`y_i=\frac{x_i}{X_i}`.
    (A variable can be left unscaled by setting :math:`X_i=1`).
```

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
+
+    With scaling, the gradient of the function becomes:
```
mphoward, line 297:
```suggestion
    The gradient of the function with respect to the scaled variables is:
```

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
+    :math:`\alpha` is defined as the step size hyperparameter. The step size must
+    be specified numerically, and a :class:`LineSearch` can be performed to find
     an optimal step size value if desired.
 
+    Then, with an initial numerical guess for all the design variables :math:`\mathbf{x}`,
+    the following iterative calculation is performed:
+
+    .. math::
+
+        \mathbf{y}_{n+1} = \mathbf{y}_{n}-\alpha\nabla f\left(\mathbf{y}\right)
```
mphoward, line 314: This is out of order, it should come after line 295 so that you define the update scheme. I also don't think it's necessary to include the second line up front.
```suggestion
    The function is iteratively minimized by taking successive steps down
    the gradient of the functional. If the scaled variables are :math:`\mathbf{y}_n`
    at iteration *n*, the next value of the variables is:

    .. math::

        \mathbf{y}_{n+1} = \mathbf{y}_{n}-\alpha\nabla f\left(\mathbf{y}\right)
```
Then put the part from lines 299-305.

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
+
+    such that each term is equivalent to:
+
+    .. math::
+
+        \left(x_i\right)_{n+1} = \left(x_i\right)_{n}-\alpha{X_i}^2 \frac{\partial f}{\partial x_i}
+
```
mphoward, line 321: Put this more as a final note:
```suggestion
    Note that this optimization procedure is equivalent to:

    .. math::

        \left(x_i\right)_{n+1} = \left(x_i\right)_{n}-\alpha{X_i}^2 \frac{\partial f}{\partial x_i}

```
for each unscaled design variable :math:`x_i`.

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
 
-        \mathbf{x}_{n+1} = \mathbf{x}_{n}-\alpha\nabla f\left(\mathbf{x}\right)
+        \nabla f\left(\mathbf{y}\right) = \left[X_1 \frac{\partial f}{\partial x_1},
+                                                      \cdots,
+                                                X_n \frac{\partial f}{\partial x_n}\right]
 
-    where :math:`\alpha` is defined as the step size hyperparameter. The step size
-    must be specified numerically, and a :class:`LineSearch` can be performed to find
+    :math:`\alpha` is defined as the step size hyperparameter. The step size must
+    be specified numerically, and a :class:`LineSearch` can be performed to find
```
mphoward, line 306:
```suggestion
    where :math:`\alpha` is the specified descent step size hyperparameter. A :class:`LineSearch`
    can be optionally be performed to optimize the choice of :math:`\alpha` between 0 and the
    specified value.
```

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
+        :class:`~relentless._collections.KeyedArray`
+            The descent amount, keyed on the objective function design variables.
+
+        """
+        k = _collections.KeyedArray(keys=gradient.keys)
+        for i in k:
+            k[i] = self.step_size
+        return k
+
+    def scale_grad(self, gradient, scale):
```
mphoward, line 373: Remove this function after you implement my suggested change to `optimize`.

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
+                                      function.''')
+        return grad_sc
+
     def optimize(self, objective):
-        """Perform the steepest descent optimization for the given objective function.
+        r"""Perform the steepest descent optimization for the given objective function.
 
         If specified, the line search is used as described in :class:`LineSearch`
         to place the objective function at a location such that the specified
         step size will reach a minimum.
```
mphoward, line 422:
```suggestion
        to choose an optimal step size.
```

relentless/optimize/method.py
```diff
@@ -280,57 +287,157 @@ def max_iter(self, value):
...
         bool or None
             ``True`` if converged, ``False`` if not converged, ``None`` if no
             design variables are specified for the objective function.
+
+        Raises
+        ------
+        ValueError
+            If the line search used has a tolerance greater than the tolerance
+            for the steepest descent.
+
```
mphoward, line 440: This will be removed once the relative tol is fixed.

relentless/optimize/method.py
```diff
@@ -339,13 +446,24 @@ def optimize(self, objective):
         iter_num = 0
         cur_res = objective.compute()
         while not self.has_converged(cur_res) and iter_num < self.max_iter:
+            grad_y = self.scale_grad(cur_res.gradient, self.scale)
```
mphoward, line 449: Once `scale` is a `KeyedArray`:
```suggestion
            grad_y = scale*cur_res.gradient
```

relentless/optimize/method.py
```diff
@@ -339,13 +446,24 @@ def optimize(self, objective):
...
+            update = self.descent_amount(grad_y)*self.scale_grad(grad_y, self.scale)
```
mphoward, line 450:
```suggestion
            update = self.descent_amount(grad_y)*grad_y
```

relentless/optimize/method.py
```diff
@@ -339,13 +446,24 @@ def optimize(self, objective):
...
             #if line search, attempt backtracking in interval
             if self.line_search is not None:
+                try:
+                    err = self.line_search.abs_tol > self.abs_tol
+                except TypeError:
+                    err = any([self.line_search.abs_tol > t for t in self.abs_tol.values()])
+                if err:
+                    raise ValueError('''The line search object must have a tolerance less
+                                        than or equal to the steepest descent object.''')
+
```
mphoward, line 466: Remove this

relentless/optimize/method.py
```diff
@@ -391,3 +528,88 @@ def line_search(self, value):
         if value is not None and not isinstance(value, LineSearch):
             raise TypeError('If defined, the line search parameter must be a LineSearch object.')
         self._line_search = value
+
+class FixedStepDescent(SteepestDescent):
+    r"""Fixed-step steepest descent algorithm.
+
+    For an :class:`~relentless.optimize.objective.ObjectiveFunction` :math:`f\left(\mathbf{x}\right)`,
```
mphoward, line 535: You could apply similar changes to these docs as I suggested to these docs.

Alternatively, you could say that this is a modification of `SteepestDescent` (and refer the user to those docs for details), then just give the equation in lines 558-561 here.

relentless/optimize/method.py
```diff
@@ -391,3 +528,88 @@ def line_search(self, value):
...
+    be specified numerically, and a :class:`LineSearch` can be performed to find
+    an optimal step size value if desired.
+
+    Then, with an initial numerical guess for all the design variables :math:`\mathbf{x}`,
+    the following iterative calculation is performed:
+
+    .. math::
+
+        \mathbf{y}_{n+1} = \mathbf{y}_{n}-\alpha\frac{\nabla f\left(\mathbf{y}\right)}
+                                                     {\lVert\nabla f\left(\mathbf{y}\right)\rVert}
```
mphoward, line 561: Lumping the norm with alpha makes the intent clearer:
```suggestion
        \mathbf{y}_{n+1} = \mathbf{y}_{n}-\frac{\alpha}{\lVert\nabla f\left(\mathbf{y}\right)\rVert} \nabla f\left(\mathbf{y}\right)
```

relentless/optimize/method.py
```diff
@@ -391,3 +528,88 @@ def line_search(self, value):
...
+
+    such that each term is equivalent to:
+
+    .. math::
+
+        \left(x_i\right)_{n+1} = \left(x_i\right)_{n}-\frac{\alpha{X_i}^2 \frac{\partial f}{\partial x_i}}
+                                                           {\sqrt{\sum\limits_{i=1}^{n} \left({X_i}^2 \frac{\partial f}{\partial x_i}\right)^2}}
```
mphoward, line 568: Per my comment above, I wouldn't repeat this here, I don't think it makes the method any clearer.

relentless/optimize/objective.py
```diff
@@ -6,8 +6,7 @@
 by adjusting the variables on which the function depends.
 
 This function, :math:`f`, is a scalar value that is defined as a function of :math:`n`
-problem :class:`DesignVariables<~relentless.variable.DesignVariable>`
-:math:`\mathbf{x}=\left[x_1,\ldots,x_n\right]`.
+problem :class:`~relentless.variable.DesignVariable`s :math:`\mathbf{x}=\left[x_1,\ldots,x_n\right]`.
```
mphoward, line 9: I think you need to escape the s with a backslash to get this to set correctly
```suggestion
problem :class:`~relentless.variable.DesignVariable`\s :math:`\mathbf{x}=\left[x_1,\ldots,x_n\right]`.
```

relentless/optimize/objective.py
```diff
@@ -50,7 +49,7 @@ class ObjectiveFunction(abc.ABC):
     """Abstract base class for the optimization objective function.
 
     An :class:`ObjectiveFunction` defines the objective function parametrized on
-    one or more adjustable :class:`DesignVariables<~relentless.variable.DesignVariable>`.
+    one or more adjustable :class:`~relentless.variable.DesignVariable`s.
```
mphoward, line 52: Same here

relentless/optimize/objective.py
```diff
@@ -70,13 +69,13 @@ def compute(self):
 
     @abc.abstractmethod
     def design_variables(self):
-        """Return all :class:`DesignVariables<~relentless.variable.DesignVariable>`
+        """Return all :class:`~relentless.variable.DesignVariable`s`
```
mphoward, line 72: And here

mphoward (changes requested):

relentless/optimize/method.py
adisr04: The properties weren't showing with just `:members:`.

relentless/optimize/method.py
```diff
@@ -280,68 +297,134 @@ def max_iter(self, value):
...
         self.line_search = line_search
 
+    def descent_amount(self, gradient):
+        r"""Calculate the descent amount for the optimization.
+
+        The amount that each update descends down the scaled gradient is:
+
+        .. math::
+
+            \alpha
```
mphoward, line 369:
```suggestion
        The amount that each update descends down the scaled gradient is a constant :math:`\alpha`.
```

relentless/optimize/method.py
```diff
@@ -391,3 +493,65 @@ def line_search(self, value):
         if value is not None and not isinstance(value, LineSearch):
             raise TypeError('If defined, the line search parameter must be a LineSearch object.')
         self._line_search = value
+
+class FixedStepDescent(SteepestDescent):
+    r"""Fixed-step steepest descent algorithm.
+
+    This is a modification of :class:`SteepestDescent` in which the function is
+    iteratively minimized by taking successive steps of fixed magnitude down
+    the normalized gradient of the functional.
```
mphoward, line 502:
```suggestion
    iteratively minimized by taking successive steps of fixed magnitude down
    the gradient of the function.
```

relentless/optimize/method.py
```diff
@@ -280,68 +297,134 @@ def max_iter(self, value):
...
+    can optionally be performed to optimize the value of :math:`\alpha` between
+    :math:`0` and the input value.
+
+    The function is iteratively minimized by taking successive steps down the
+    gradient of the functional. If the scaled variables are :math:`\mathbf{y}_n`
+    at iteration :math:`n`, the next value of the variables is:
+
+    .. math::
+
+        \mathbf{y}_{n+1} = \mathbf{y}_n-\alpha\nabla f\left(\mathbf{y}\right)
```
mphoward, line 318:
```suggestion
        \mathbf{y}_{n+1} = \mathbf{y}_n-\alpha\nabla f\left(\mathbf{y}_n\right)
```

relentless/optimize/method.py
```diff
@@ -391,3 +493,65 @@ def line_search(self, value):
...
+    iteratively minimized by taking successive steps of fixed magnitude down
+    the normalized gradient of the functional.
+
+    If the scaled variables are :math:`\mathbf{y}_n` at iteration :math:`n`, the
+    next value of the variables is:
+
+    .. math::
+
+        \mathbf{y}_{n+1} = \mathbf{y}_n
+                          -\frac{\alpha}{\lVert\nabla f\left(\mathbf{y}\right)\rVert}\nabla f\left(\mathbf{y}\right)
```
mphoward, line 510:
```suggestion
                          -\frac{\alpha}{\lVert\nabla f\left(\mathbf{y}\right)\rVert}\nabla f\left(\mathbf{y}_n\right)
```

relentless/optimize/method.py
```diff
@@ -391,3 +493,65 @@ def line_search(self, value):
...
+
+    def descent_amount(self, gradient):
+        r"""Calculate the descent amount for the optimization.
+
+        The amount that each update descends down the scaled gradient is:
+
+        .. math::
+
+            \frac{\alpha}{\lVert\nabla y\rVert}
+
```
mphoward, line 542: Add:

"which makes a step of constant magnitude."

mphoward (approved):
I have only a handful of small comments. Please address, and then I think we can go ahead and merge!

PR #59, opened 2021-04-13
Convergence tests for optimizers

adisr04:
- Add `ConvergenceTest` abstract base class with `LogicTest` abstract implementation, composed with class `Tolerance`
- Implement `GradientTest`, `ValueTest`,  `AllTest`, `AnyTest`, `OrTest`, and `AndTest` implementations
- Modify optimizers in `method.py` to use `stop` parameter for convergence criterion
- Add `DefaultDict` class in `_collections.py` with unit tests
- Document classes in `criteria.py` and add unit tests in `test_criteria.py`


relentless/optimize/criteria.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+        abs_tol : float or dict
+            The absolute tolerance or tolerances (keyed on the :class:`~relentless.optimize.objective.ObjectiveFunction`
+            design variables).
+        rel_tol : float or dict
+            The relative tolerance or tolerances (keyed on the :class:`~relentless.optimize.objective.ObjectiveFunction`
+            design variables). (Defaults to ``None``).
+
+    """
+    def __init__(self, abs_tol, rel_tol=None, mode='or'):
```
mphoward, line 15: Overall, I very much like this design, but I think this is a little too base-class heavy. I don't think all `ConvergenceTest`s should be required to support both absolute tolerance and relative tolerance or these different modes.

Rather, I would make `ConvergenceTest` be a light class:
```py
class ConvergenceTest(abc.ABC):
    @abc.abstractmethod
    def converged(self, result):
        pass
```

Then, you can subclass abstract classes for absolute and relative tolerances:
```py
class AbsoluteTolerance(ConvergenceTest):
    def __init__(self, tolerance):
        pass

class RelativeTolerance(ConvergenceTest):
    def __init__(self, tolerance):
        pass
```
and implement specific cases of each:
```py
class GradientAbsoluteTolerance(AbsoluteTolerance):
    pass
class GradientRelativeTolerance(RelativeTolerance):
    pass
```

In this way, `LogicTest` can also subclass from `ConvergenceTest`:
```py
class LogicTest(ConvergenceTest):
    pass
``
I think we voted not to do component-wise and/or tests, so you can remove mode an implement this with the `OrTest` and `AndTest`.

File: relentless/optimize/criteria.py [hunk 2]
```diff
@@ -0,0 +1,427 @@
...
+            elif x.atlow() and -grad > tol:
+                converged = False
+                break
+            elif x.isfree() and np.abs(grad) > tol:
+                converged = False
+                break
+
+        return converged
+
+class RelativeGradientTest(RelativeTolerance):
```
mphoward, line 243: I'm not sure that I actually get what this test is a relative tolerance with respect to. What should we be checking? @zeesherman

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,427 @@
...
+        for x in result.design_variables:
+            grad = result.gradient[x]
+            tol = self._get_tol(x)
+            if x.athigh() and -grad < -tol:
+                converged = False
+                break
+            elif x.atlow() and -grad > tol:
+                converged = False
+                break
+            elif x.isfree() and np.abs(grad) > tol:
```
mphoward, line 237:
```suggestion
            elif x.isfree() and not self.check(grad,0):
```

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,427 @@
...
+    design variables :math:`x_i`.
+
+    Parameters
+    ----------
+    tolerance : float or dict
+        The absolute tolerance or tolerances (keyed on the :class:`~relentless.optimize.objective.ObjectiveFunction`
+        design variables).
+
+"""
+    def __init__(self, tolerance):
```
mphoward, line 322: You should an add an argument that is the value you are trying to get to.

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,427 @@
...
+        Returns
+        -------
+        bool
+            ``True`` if the function is converged.
+
+        """
+        converged = True
+        for x in result.design_variables:
+            tol = self._get_tol(x)
+            if x.value > tol:
```
mphoward, line 342:
```suggestion
            if not self.check(x,self.value):
```

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,427 @@
...
+                converged = False
+                break
+
+        return converged
```
mphoward, line 346: There should also be a `ValueRelativeTolerance`.

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,427 @@
...
+    def __init__(self, *tests):
+        super().__init__(*tests)
+
+    def converged(self, result):
+        """Check if the function has converged by any of the specified tests.
+
+        Parameters
+        ----------
+        result : :class:`~relentless.optimize.objective.ObjectiveFunctionResult`
+            The location of the function at which to check for convergence.
```
mphoward, line 388:
```suggestion
            The result to check for convergence.
```

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,427 @@
...
+    def __init__(self, *tests):
+        super().__init__(*tests)
+
+    def converged(self, result):
+        """Check if the function is converged by all of the specified tests.
+
+        Parameters
+        ----------
+        result : :class:`~relentless.optimize.objective.ObjectiveFunctionResult`
+            The location of the function at which to check for convergence.
```
mphoward, line 419:
```suggestion
            The result to check for convergence.
```

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,427 @@
...
+
+        Returns
+        -------
+        bool
+            ``True`` if the function is converged by all tests.
+
+        """
+        return all([t.converged(result) for t in self.tests])
```
mphoward, line 427: They are just special cases, but people might expect to see `AndTest` and `OrTest`. I think you can add them like this:
```py
class AndTest(AllTest):
    def __init__(self, a, b):
        super().__init__(a,b)
class OrTest(AnyTest):
    def __init__(self, a, b):
        super().__init__(a,b)
```

mphoward (changes requested):
Thanks! This is a great start. I have commented on `criteria.py`. I have a few suggestions for revisions, and also a question about how we are defining the relative-tolerance gradient test. Once you have made these changes, I can review the rest of the PR later.

relentless/optimize/method.py
```diff
@@ -269,17 +227,17 @@ def find(self, objective, start, end):
     @property
     def rel_tol(self):
```
mphoward, line 228: You can use a `Tolerance` here now, with `absolute=0`. I would expose `self._tolerance.relative` as `tolerance` for consistency with the gradient test, or else as `relative`.

mphoward (changes requested):

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,487 @@
...
+It may be helpful for the class to be composed having a :class:`Tolerance`.
+
+.. autosummary::
+    :nosignatures:
+
+    ConvergenceTest
+    LogicTest
+    Tolerance
+
+.. autoclass:: ConvergenceTest
```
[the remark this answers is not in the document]
adisr04: I followed the template you used in `variable.py` and `pair.py` where you put related classes together, not necessarily alphabetically.

tests/optimize/test_method.py
```diff
@@ -41,6 +35,7 @@ def test_find(self):
         res_2 = q.compute()
         x.value = -3.0
         res_new = l.find(objective=q, start=res_1, end=res_2)
+        print(res_new)
```
mphoward, line 38:
```suggestion
```

mphoward (changes requested):
I think these changes will fix the issue you were having, and then this will be good to go. I will apply and test them now.

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,422 @@
...
+        """float: The value(s) to check."""
+        return self._value
+
+    @value.setter
+    def value(self, value):
+        self._value = value
+
+    @property
+    def absolute(self):
+        """`~relentless._collections.DefaultDict`: The absolute tolerance(s)."""
```
mphoward, line 288:
```suggestion
        """float: The absolute tolerance."""
```

relentless/optimize/criteria.py
```diff
@@ -0,0 +1,422 @@
...
+        return self._tolerance.absolute
+
+    @property
+    def relative(self):
+        """`~relentless._collections.DefaultDict`: The relative tolerance(s)."""
```
mphoward, line 293:
```suggestion
        """float: The relative tolerance."""
```

mphoward (approved):

PR #61, opened 2021-04-28
Implement relative entropy objective function

adisr04:
- Implement `RelativeEntropy` objective function class
- Document `RelativeEntropy` in `objective.py`
- Add unit tests in `test_objective.py`
- Add example script of relative entropy minimization to `examples/lj`


relentless/optimize/objective.py
```diff
@@ -165,3 +168,55 @@ def design_variables(self):
...
+
+        # compute the relative entropy gradient by integration
+        g_tgt = self.target.rdf
+        g_sim = sim_ens.rdf
+        gradient = {}
+        dvars = self.design_variables()
+
+        for x in dvars:
+            update = 0
+            for i,j in self.potentials.pair.potentials.coeff.pairs:
```
mphoward, line 193: We need a simpler way to do this... this could work:
```suggestion
            for i,j in self.target.pairs:
```

relentless/optimize/objective.py
```diff
@@ -165,3 +168,55 @@ def design_variables(self):
...
+                # evaluate derivative wrt x and interpolate
+                dudx = _math.Interpolator(self.potentials.pair.x, self.potentials.pair.derivative((i,j), x))
```
mphoward, line 195:
```suggestion
                dudx = _math.Interpolator(self.potentials.pair.r, self.potentials.pair.derivative((i,j), r))
```

relentless/optimize/objective.py
```diff
@@ -165,3 +168,55 @@ def design_variables(self):
...
+
+                # find common domain to compare rdfs
+                r0 = max(g_sim[i,j].table[0][0], g_tgt[i,j].table[0][0])
+                r1 = min(g_sim[i,j].table[-1][0], g_tgt[i,j].table[-1][0])
```
mphoward, line 199: You might also want to include `self.potentials.pair.rmax` in the min, as this is the bound up to which the potentials are evaluated.

tests/optimize/test_objective.py
```diff
@@ -74,5 +74,27 @@ def tearDown(self):
...
+        v_obj = relentless.volume.Cube(L=3.0)
+        target = relentless.ensemble.Ensemble(T=10, V=v_obj, N={'A':1, 'B':2})
+
+        thermo = relentless.simulate.dilute.AddEnsembleAnalyzer()
+        operations = [relentless.simulate.InitializeRandomly(seed=2,neighbor_buffer=0.4),
+                      relentless.simulate.AddNVTIntegrator(dt=0.1,tau_T=1.0),
+                      thermo]
+        simulation = relentless.simulate.dilute.Dilute(operations)
+
+        lj = relentless.potential.LennardJones(types=('1','2'))
```
mphoward, line 91: You need to define the parameters and design variables for the lj potential here. I'm surprised this even runs, actually... do the potentials not raise an error if one of their parameters is not set?

Also, make life easier on yourself and start with just one type for now.

relentless/optimize/objective.py
```diff
@@ -165,3 +168,55 @@ def design_variables(self):
...
+                r = np.arange(r0, r1+0.5*self.dr, self.dr)
+
+                # take integral by trapezoidal rule
+                sim_factor = sim_ens.N[i]*sim_ens.N[j]/sim_ens.V.volume
+                tgt_factor = self.target.N[i]*self.target.N[j]/self.target.V.volume
```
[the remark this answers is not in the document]
zeesherman: That's right! You can put beta into each of these factors.

relentless/optimize/objective.py
```diff
@@ -165,3 +168,55 @@ def design_variables(self):
...
+                mult = 2 if i == j else 4 # 2 if same, otherwise need i,j and j,i contribution
```
zeesherman, line 205: Just want to make sure: the loop over `self.potentials.pair.potentials.coeff.pairs` (or `self.target.pairs` like Mike suggests) does not double count pairs right? So if we have type A and B particles, we get three pairs (A,A), (A,B), and (B,A)?

Also small nit, maybe move a factor of 2 into the next line and make this statement pick between 1 and 2. It'll be more obvious that `mult` is referring to counting pairs that way.

mphoward: Yep, pairs are the _unique_ pairs of the system: (A,A), (A,B), and (B,B).

tests/optimize/test_objective.py
```diff
@@ -74,5 +74,27 @@ def tearDown(self):
         self.directory.cleanup()
         del self.directory
 
+class test_RelativeEntropy(unittest.TestCase):
+    """Unit tests for relentless.optimize.RelativeEntropy"""
+
+    def test_compute(self):
+        """Test compute method"""
+        v_obj = relentless.volume.Cube(L=3.0)
+        target = relentless.ensemble.Ensemble(T=10, V=v_obj, N={'A':1, 'B':2})
```
mphoward, line 83: Edit to add: you need to set a target RDF here! Again, not sure what the behavior is if you don't.

relentless/optimize/objective.py
```diff
@@ -165,3 +168,66 @@ def design_variables(self):
...
+        g_sim = sim_ens.rdf
+        dvars = self.design_variables()
+        gradient = _collections.KeyedArray(keys=dvars)
+
+        for var in dvars:
+            update = 0
+            for i,j in self.target.pairs:
+                # evaluate derivative wrt design variable and interpolate with r
+                rs = self.potentials.pair.r
+                dus = self.potentials.pair.derivative((i,j),var)
```
mphoward, line 199: I would evaluate u directly, since this is what is going to force g(r) = 0, to check for infinity rather than the derivative.

relentless/optimize/objective.py
```diff
@@ -165,3 +168,66 @@ def design_variables(self):
...
+                for k,duk in enumerate(dus):
+                    if np.isinf(duk):
+                        rs = np.delete(rs,k)
+                        dus = np.delete(dus,k)
+                        break
```
mphoward, line 204: This is the right idea, but I think you can do this faster:
```py
flags = np.isinf(u)
first_finite = 0
while not flags[first_finite] and first_finite < len(rs):
    first_finite += 1
rs = rs[first_finite:]
dus = dus[first_finite:]
```
(There might be a way to do this without a loop too, but at least here you are not copy the array repeatedly)

relentless/optimize/objective.py
```diff
@@ -165,3 +168,66 @@ def design_variables(self):
...
+                dudvar = _math.Interpolator(rs,dus)
+
+                # find common domain to compare rdfs
+                r0 = max(g_sim[i,j].table[0][0],g_tgt[i,j].table[0][0])
+                r1 = min(g_sim[i,j].table[-1][0],g_tgt[i,j].table[-1][0],self.potentials.pair.rmax)
+                r = np.arange(r0,r1+0.5*self.dr,self.dr)
```
mphoward, line 210: @zeesherman do we really need to use a separate array of rs, or could we use the ones we used to tabulate the pair potential (cut down to the same bounds as the rdfs)? It seems like this would work just fine. Then we could get rid of dr as an argument here.

relentless/optimize/objective.py
```diff
@@ -165,3 +168,66 @@ def design_variables(self):
...
+
+                # take integral by trapezoidal rule
+                sim_factor = sim_ens.N[i]*sim_ens.N[j]*sim_ens.beta/sim_ens.V.volume**2
+                tgt_factor = self.target.N[i]*self.target.N[j]*self.target.beta/self.target.V.volume**2
```
mphoward, line 214: These need to be asymmetric in their volumes, since we are normalizing the relative entropy by the volume of the target ensemble.
```suggestion
                sim_factor = sim_ens.N[i]*sim_ens.N[j]*sim_ens.beta/(self.target.V.volume*sim_ens.V.volume)
                tgt_factor = self.target.N[i]*self.target.N[j]*self.target.beta/self.target.V.volume**2
```

tests/optimize/test_objective.py
```diff
@@ -74,5 +76,37 @@ def tearDown(self):
...
+        target.rdf['1','1'] = relentless.ensemble.RDF(r=rs, g=gs)
+
+        thermo = relentless.simulate.dilute.AddEnsembleAnalyzer()
+        simulation = relentless.simulate.dilute.Dilute(operations=[thermo])
+
+        relent = relentless.optimize.RelativeEntropy(target,simulation,potentials,thermo,dr)
+        res = relent.compute()
+        self.assertIsNone(res.value)
+        assert not np.isinf(res.gradient[epsilon])
+        assert not np.isinf(res.gradient[sigma])
```
mphoward, line 108: Switch these to unit test asserts

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+    :math:`\mathbf{x}` given the target ensemble :math:`\mathbf{T}`. Given
+    the rdfs of the simulation and target ensembles, :math:`g(r|\mathbf{x})` and :math:`g_{tgt}(r)`,
+    respectively, the total pair potential energy, the gradient is:
+
+    .. math::
+
+        \nabla_\mathbf{x} L\left(\mathbf{x}|\mathbf{T}\right) = 2\pi N\rho\int{dr r^2 \left(g(r|\mathbf{x})-g_{tgt}(r)\right)\beta u(r|\mathbf{x})}
+
+    where :math:`\beta` is the thermodynamic beta, :math:`N` is the particular number,
+    and :math:`\rho=N/V` is the number density, all for the target ensemble.
```
mphoward, line 204: Here is a rewrite of these docs. I also propose that we make the normalization of S an option of the class so that you can still get the extensive version if you want. I propose adding an option `intensive` that defaults to True, but it could also be called `extensive`, and I don't have a strong preference what the default value is.

Also, somebody should check my math on the equations :)

```suggestion
    The value of the relative entropy is not readily determined in molecular simulations,
    so this :class:`ObjectiveFunction` does not return a value.

    However, its gradient with respect to design variables **x** is much easier to compute.
    Currently, the :class:`RelativeEntropy` objective function supports
    :class:`~relentless.potential.pair.PairPotential` interactions. These interactions
    are characterized by an :class:`~relentless.ensemble.RDF` for each pair
    of interacting types in each :class:`~relentless.ensemble.Ensemble`,
    :math:`g_{ij}(r)`. The gradient of :class:`S_{\rm rel}` is then:

    .. math::

        \nabla_\mathbf{x} S_{\rm rel} = \frac{1}{2} \sum_{i,j} \int{dr 4\pi r^2 \left[\frac{\beta N_i N_j}{V} g_{ij}(r)- \frac{\beta_0 N_{i,0} N_{j,0}}{V_0} g_{ij,0}(r)\right] \nabla_\mathbf{x} u_{ij}(r)}

    where :math:`\beta = 1/(k_{\rm B}T)`, :math:`N_i` is the number of particles of
    type *i*, *V* is the volume, and :math:`u_{ij}` is the pair potential in the *model*
    ensemble.
    
    The relative entropy :math:`S_{\rm rel}` is extensive as written, meaning that it
    depends on the size of the system. This can be undesirable for optimization because
    it means certain hyperparameters are system-size dependent. To circumvent this,
    use ``intensive=True`` (default) to normalize :math:`s_{\rm rel} = S_{\rm rel}/V_0`.
    The extensive relative entropy can still be computed by setting ``intensive=False``.
```

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
         return self._design_variables
+
+class RelativeEntropy(ObjectiveFunction):
+    r"""Relative entropy.
+
+    The relative entropy quantifies how different two probability distributions are.
+    This can be applied to the probability distributions of a target ensemble
+    and the ensemble of a simulated molecular system, which is parametrized
+    by some pair potential functions.
+
```
mphoward, line 192: I think we should add another paragraph here describing how this works. The two key points are:
* The target is known in advance.
* The model ensemble is generated by a simulation for a given set of parameters.

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+        self.simulation = simulation
+        self.potentials = potentials
+        self.thermo = thermo
+        self.dr = dr
+        self.communicator = communicator
+
+    def compute(self, directory=None):
+        """Evaluate the value and gradient of the relative entropy function.
+
+
```
mphoward, line 235: You should add a comment here about what is going to be written and how the files will be named.

I would also add a comment making it clear that this will require running a simulation, so it can be expensive to compute.

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+        Parameters
+        ----------
+        directory : :class:`~relentless.data.Directory`
+            The directory to which the values of the pair potential design
+            variables at the time of computation are saved (defaults to ``None``).
+
+        Returns
+        -------
+        :class:`ObjectiveFunctionResult`
+            The result of :meth:`ObjectiveFunction.make_result()`
```
mphoward, line 245:
```suggestion
            The result, which has unknown value ``None`` but known gradient.
```

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+    ----------
+    target : :class:`~relentless.ensemble.Ensemble`
+        The target ensemble (must have specified N and V).
+    simulation :class:`~relentless.simulate.Simulation`
+        The simulation engine to use, with specified simulation operations.
+    potentials :class:`~relentless.simulate.Potentials`
+        The pair potentials to use in the simulations.
+    thermo : :class:`~relentless.simulate.SimulationOperation`
+        The thermodynamic analyzer for the simulation ensemble and rdf (usually
+        :meth:`AddEnsembleAnalyzer()` for the specified simulation type.
```
mphoward, line 216:
```suggestion
        The thermodynamic analyzer operation for the simulation ensemble and rdf (usually
        :meth:`AddEnsembleAnalyzer()`). The model ensemble will be extracted from this
        operation.
```

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+    dr : float
+        The radial step size.
+    communicator : :class:`~relentless.mpi.Communicator`
+        The communicator used for saving outputs to file ##?## (defaults to
+        ``None``).
```
mphoward, line 221:
```suggestion
        The communicator used for running the ``simulation`` (defaults to
        ``None``).
```

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+
+    def compute(self, directory=None):
+        """Evaluate the value and gradient of the relative entropy function.
+
+
+        Parameters
+        ----------
+        directory : :class:`~relentless.data.Directory`
+            The directory to which the values of the pair potential design
+            variables at the time of computation are saved (defaults to ``None``).
```
mphoward, line 240:
```suggestion
            The output directory. In addition to simulation output, the pair potential design
            variables at the time of computation are saved (defaults to ``None``).
```

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+                while not flags[first_finite] and first_finite < len(rs):
+                    first_finite += 1
+                rs = rs[first_finite+1:]
+                dus = dus[first_finite+1:]
+
+                #interpolate derivative wrt design variable with r
+                dudvar = _math.Interpolator(rs,dus)
+
+                # find common domain to compare rdfs
+                r0 = max(g_sim[i,j].table[0][0],g_tgt[i,j].table[0][0])
```
mphoward, line 276: Also need `rs[0]` here, since it could be bigger than either of these.
```suggestion
                r0 = max(g_sim[i,j].table[0,0],g_tgt[i,j].table[0,0],dudvar.table[0,0])
```

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+                r1 = min(g_sim[i,j].table[-1][0],g_tgt[i,j].table[-1][0],self.potentials.pair.rmax)
```
mphoward, line 277: I think this will be clearer if you use the interpolated derivative:
```suggestion
                r1 = min(g_sim[i,j].table[-1,0],g_tgt[i,j].table[-1,0],dudvar.table[-1,0])
```

relentless/optimize/objective.py
mphoward: We do still need to discuss this point. I am in favor of simplifying the number of parameters (i.e., removing `dr` if it is not really needed).

relentless/optimize/objective.py
```diff
@@ -165,3 +181,150 @@ def design_variables(self):
...
+
+        # optionally write output to directory
+        if directory is not None:
+            if self.communicator is None or self.communicator.rank == self.communicator.root:
+                n = 0
+                for p in self.potentials.pair.potentials:
+                    fname = 'potential_{}.log'.format(n)
+                    with open(directory.file(fname),'w') as f:
+                        p.save(f.name)
+                    n += 1
```
mphoward, line 297: This is mostly correct but can be simplified (see below). Our only big issue is that we don't have an obvious mapping between potentials and parameters (like, how do you know what object is 0 or 1), but this is not easy to solve here. Let's just stick with this for now:
```suggestion
        if directory is not None and (self.communicator is None or self.communicator.rank == self.communicator.root):
            for n,p in enumerate(self.potentials.pair.potentials):
                p.save(directory.file('potential.{}.json'.format(n)))
```

tests/optimize/test_objective.py
```diff
@@ -74,5 +78,125 @@ def tearDown(self):
...
+        sim_ens = self.thermo.extract_ensemble(sim)
+        rs = self.potentials.pair.r[1:]
+        dus = self.potentials.pair.derivative(('1','1'),var)[1:]
+        dudvar = relentless._math.Interpolator(rs,dus)
+        sim_factor = sim_ens.N['1']**2*sim_ens.beta/(sim_ens.V.volume*self.target.V.volume)
+        tgt_factor = self.target.N['1']**2*self.target.beta/(self.target.V.volume**2)
+
+        r = np.arange(0.05, 3.55, self.dr)
+        y = 2*np.pi*r**2*(sim_factor*sim_ens.rdf['1','1'](r)-tgt_factor*self.target.rdf['1','1'](r))*dudvar(r)
+        return scipy.integrate.trapz(x=r, y=y)
```
mphoward, line 145: This is a nice idea, but I think it would be good to make this function a little less dependent on some of the internals of the relentless code. For example, you could hard-code the integrands to the functional forms you expect (rather than use the simulation & ensemble). We want to try to get as independent of a verification as possible.

mphoward (changes requested):

tests/optimize/test_objective.py
```diff
@@ -74,5 +78,152 @@ def tearDown(self):
...
+        self.target = relentless.ensemble.Ensemble(T=1.5, V=v_obj, N={'1':50})
+        rs = np.arange(0.05,5.0,0.1)
+        gs = np.exp(-lj.energy(('1','1'),rs))
+        self.target.rdf['1','1'] = relentless.ensemble.RDF(r=rs, g=gs)
+
+        self.thermo = relentless.simulate.dilute.AddEnsembleAnalyzer()
+        self.simulation = relentless.simulate.dilute.Dilute(operations=[self.thermo])
+
+    def relent_grad(self, var, ext=False):
+        rs = np.arange(0,3.6,3.6/1000)[1:]
```
mphoward, line 106: If you use linspace, you can be sure to hit the end point:
```suggestion
        rs = np.linspace(0,3.6,1001)[1:]
```

tests/optimize/test_objective.py
```diff
@@ -74,5 +78,152 @@ def tearDown(self):
...
+        r6_inv = np.power(0.9*1./rs, 6)
+        gs = np.exp(-(1/1.5)*4.*1.0*(r6_inv**2 - r6_inv))
+        sim_rdf = relentless._math.Interpolator(rs,gs)
+
+        rs = np.arange(0.05,5.0,0.1)
+        r6_inv = np.power(0.9/rs, 6)
+        gs = np.exp(-4.*1.0*(r6_inv**2 - r6_inv))
+        tgt_rdf = relentless._math.Interpolator(rs,gs)
+
+        rs = np.arange(0,3.6,3.6/1000)[1:]
```
mphoward, line 116:
```suggestion
        rs = np.linspace(0,3.6,1001)[1:]
```

tests/optimize/test_objective.py
```diff
@@ -74,5 +78,152 @@ def tearDown(self):
...
+        dudvar = relentless._math.Interpolator(rs,dus)
+
+        if ext:
+            norm_factor = 1.
+        else:
+            norm_factor = 1000.
+        sim_factor = 50**2*(1/1.5)/(1000*norm_factor)
+        tgt_factor = 50**2*(1/1.5)/(1000*norm_factor)
+
+        r = np.arange(0.05, 3.55, 0.00355)
```
mphoward, line 131:
```suggestion
        r = np.linspace(0.05, 3.55, 1001)
```

tests/optimize/test_objective.py
```diff
@@ -74,5 +78,152 @@ def tearDown(self):
...
+                                                     self.thermo)
+
+        res = relent.compute()
+        self.assertIsNone(res.value)
+        grad_eps = self.relent_grad(self.epsilon)
+        grad_sig = self.relent_grad(self.sigma)
+        err_grad_eps = 1.-np.abs(grad_eps/res.gradient[self.epsilon])
+        err_grad_sig = 1.-np.abs(grad_sig/res.gradient[self.sigma])
+        self.assertLess(err_grad_eps, 0.01) #<1% error in floating point rounding
+        self.assertLess(err_grad_sig, 0.01)
```
mphoward, line 173: If you want to get agreement to a certain number of decimal places, use the `places` option. For example, if the numbers are order 1, 1% would be 2 places:
```suggestion
        self.assertAlmostEqual(res.gradient[self.epsilon], grad_eps, places=2)
        self.assertAlmostEqual(res.gradient[self.sigma], grad_sig, places=2)
```
You might want to make places smaller depending on what the values of these is, though.

There is also numpy.testing.assert_allclose, which will let you set an atol and rtol for closeness, if you prefer that route.

tests/optimize/test_objective.py
```diff
@@ -74,5 +78,152 @@ def tearDown(self):
...
+                                                     self.simulation,
+                                                     self.potentials,
+                                                     self.thermo,
+                                                     extensive=True)
+
+        res = relent.compute()
+        self.assertIsNone(res.value)
+        grad_eps = self.relent_grad(self.epsilon, ext=True)
+        grad_sig = self.relent_grad(self.sigma, ext=True)
+        err_grad_eps = 1.-np.abs(grad_eps/res.gradient[self.epsilon])
```
mphoward, line 187: Same check down here, may need different number of places because extensive.

mphoward (changes requested):
This looks like it's almost ready! Here are my last set of requested changes.

examples/lj/project.py
```diff
@@ -2,31 +2,32 @@
...
-# change parameters and setup optimization
-lj.coeff['1','1'] = {'epsilon': 1.0, 'sigma': 1.0}
-lj.variables['1','1'] = (relentless.Variable('sigma', low=0.8, high=1.2),)
-
-# lammps simulation engine
-tab = relentless.potential.Tabulator(nbins=1000, rmin=0.0, rmax=3.6, fmax=100., fcut=1.e-6)
-sim = relentless.engine.LAMMPS(ensemble=tgt, table=tab, lammps='lmp_mpi', template='nvt.in', policy=relentless.environment.Policy(procs=2), potentials=lj)
+# dilute molecular simulation
+thermo = relentless.simulate.AddEnsembleAnalyzer(check_thermo_every=5, check_rdf_every=5, rdf_dr=dr)
+simulation = relentless.simulate.dilute.Dilute([thermo])
```
mphoward, line 23:
```suggestion
simulation = relentless.simulate.Dilute([thermo])
```

PR #70, opened 2021-09-22
Expand documentation for modules

adisr04:
- Implement and extend documentation (docstrings and Sphinx) for core (`data`, `ensemble`, `volume`, `_collections`, `_math`) and `simulate` modules
- Fix typos and hyperlinks in all module docs

### DIFF
--- a/relentless/_math.py
+++ b/relentless/_math.py
@@ -1,3 +1,19 @@
+"""
+Math functions
+==============
+
+Contains a interpolator using `Akima 1D splines <https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.Akima1DInterpolator.html>`_.
+The function's value and derivatives can be evaluated.
+
+.. autosummary::
+    :nosignatures:
+
+    Interpolator
+
+.. autoclass:: Interpolator
+    :members:
+
+"""
 import numpy as np
 import scipy.interpolate
 
--- a/relentless/_math.py
+++ b/relentless/_math.py
@@ -2,8 +2,8 @@
 Math functions
 ==============
 
-Contains a interpolator using `Akima 1D splines <https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.Akima1DInterpolator.html>`_.
-The function's value and derivatives can be evaluated.
+This module contains a interpolator using Akima 1D splines. The function's value
+and derivatives can be evaluated.
 
 .. autosummary::
     :nosignatures:
@@ -18,7 +18,7 @@ import numpy as np
 import scipy.interpolate
 
 class Interpolator:
-    """Interpolating function.
+    r"""Interpolating function.
 
     Interpolates through a function :math:`y(x)` on the domain
     :math:`a \le x \le b` using Akima splines. Outside this domain, `y` is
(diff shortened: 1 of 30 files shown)


relentless/_math.py
```diff
@@ -1,35 +1,51 @@
+"""
+Math functions
+==============
+
+This module contains a interpolator using Akima 1D splines. The function's value
+and derivatives can be evaluated.
```
mphoward, line 6: I think go a little more generic. We might reorg `_math` and `_collections`, so something like this would be fine:
```suggestion
This module implements some convenience objects for mathematical operations.
```

relentless/data.py
```diff
@@ -1,8 +1,23 @@
-"""Core data management.
+"""
+Data management
```
mphoward, line 2: I opened issue #71 to remove the `Project` class, which is largely unused. This module then only contains `Directory`. Maybe we should think about combining `data` with `_collections` or something. You can leave this description as-is since we might want to change it later anyway.

relentless/ensemble.py
```diff
@@ -1,3 +1,21 @@
+"""
+Ensemble
+========
+
+This module defines a statistical mechanical thermodynamic ensemble. It is parametrized
+by a temperature, a particle type number, and either the volume or pressure of
+the system. For each pair in the ensemble, the radial distribution function (RDF)
+is also defined using an Akima spline.
```
mphoward, line 8: I think we should probably make this description a little more generic, since it is reproducing a lot of the `Ensemble` documentation. Maybe better would be to describe the intent of these objects more broadly (to define properties and statistics of simulated systems). For example, I think eventually we will add other types of distribution functions here, e.g., bond-length distribution.

relentless/simulate/__init__.py
```diff
@@ -1,3 +1,29 @@
+"""
```
mphoward, line 1: In my opinion, we should document the `generic` operations here (rather than in a separate module) because we'd like people to use them in this namespace, like:
```
relentless.simulate.AddNVTIntegrator()
```
I'm not sure if you can force sphinx to document those here (using the proper namespace), or if you just need to copy the docs from there to here.

I think these module docs should also be expanded a bit to describe to describe the general philosophy with a short example of creating a specific `Simulation` but then use the generic operations. We also want to add some developer-type docs about how to add new operations or backends.

relentless/simulate/dilute.py
```diff
@@ -1,3 +1,18 @@
+"""
+Dilute system
+=============
+
+Provides the :meth:`AddEnzembleAnalyzer()` operation for a dilute system.
```
mphoward, line 5: This should describe what a dilute simulation is. Alternatively, put it in the `Dilute` class documentation and include that here. (That might be better.)

relentless/simulate/hoomd.py
```diff
@@ -1,3 +1,54 @@
+"""
+HOOMD operations
+================
+
+Implements the generic simulation operations using HOOMD-blue.
```
mphoward, line 5: Same here as for `Dilute`.

relentless/simulate/lammps.py
```diff
@@ -1,3 +1,44 @@
+"""
+LAMMPS operations
+=================
+
+Implements the generic simulation operations using LAMMPS.
```
mphoward, line 5: Same here as for `Dilute`.

mphoward (changes requested):
Here are some high level comments!

Issue #67, opened 2021-09-16
Import dependencies by full name


mphoward:
To avoid polluting namespaces, import packages by full name in the main code. Ex:
```
import numpy
```
rather than
```
import numpy as np
```

PR #74, opened 2021-10-12
Fix dependency imports

adisr04:
Import all packages by full name. Addresses #67


tests/test_mpi.py
```diff
@@ -3,11 +3,11 @@
 
 import os
 import tempfile
-import numpy as np
+import numpy
 
 import relentless
 try:
-    import mpi4py.MPI as MPI
+    from mpi4py.MPI import MPI
```
mphoward, line 10:
```suggestion
    from mpi4py import MPI
```

PR #76, opened 2021-10-18
Fix bug in Dilute pressure calculation

### LINKED ISSUE #63: Dilute pressure with inf
Warnings like:
```
relentless/simulate/dilute.py:106: RuntimeWarning: invalid value encountered in multiply
  ens.P += (2.*np.pi/3.)*rho_a*rho_b*np.trapz(y=f*gr*r**3,x=r)
```
are produced if `f` has an inf when gr is zero. We should omit those values from the integrand.

adisr04:
Omit values where `f` is infinite from the integrand to calculate pressure. Fixes #63

### DIFF
--- a/relentless/simulate/dilute.py
+++ b/relentless/simulate/dilute.py
@@ -19,7 +19,7 @@ It can be accessed using the corresponding :class:`~relentless.simulate.generic.
         extract_ensemble
 
 """
-import numpy as np
+import numpy
 
 from relentless.ensemble import RDF
 from relentless._math import Interpolator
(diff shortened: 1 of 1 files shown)


relentless/simulate/dilute.py
```diff
@@ -135,8 +135,9 @@ def __call__(self, sim):
                 r = sim.potentials.pair.r
                 u = sim.potentials.pair.energy((a,b))
                 f = sim.potentials.pair.force((a,b))
+                f[f==numpy.inf] = 0.                          # values omitted from integrand
```
mphoward, line 138: This will work, but your intention here is confusing since you are zeroing out the force. I would rather that you do something like in the `RelativeEntropy` class
https://github.com/mphowardlab/relentless/blob/02da640a9b7622762540aeb49bc0bd8d3b333edb/relentless/optimize/objective.py#L298-L303

Search for the first value of `u` that is finite (not `f`), then only integrate for r values from this point onward.

Issue #65, opened 2021-05-12
Move neighbor buffer to pair potentials


mphoward:
The `neighbor_buffer` option for initialization does not make a ton of sense. It would be better to add it as a required argument of the tabulated pair potentials.

PR #77, opened 2021-10-20
Refactor neighbor buffer parameter

adisr04:
Move `neighbor_buffer` parameter from `simulate` initialization operations to `PairPotentialTabulator`. Addresses #65

### DIFF
--- a/relentless/simulate/simulate.py
+++ b/relentless/simulate/simulate.py
@@ -402,10 +402,7 @@ class PairPotentialTabulator(PotentialTabulator):
 
     @neighbor_buffer.setter
     def neighbor_buffer(self, val):
-        if val is not None:
-            self._neighbor_buffer = val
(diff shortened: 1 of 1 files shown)


relentless/simulate/simulate.py
```diff
@@ -391,6 +394,19 @@ def rmax(self):
...
+        """float: The amount to be added to ``rmax`` to search for particles while
+        computing the neighbor list."""
+        return self._neighbor_buffer
+
+    @neighbor_buffer.setter
+    def neighbor_buffer(self, val):
+        if val is not None:
+            self._neighbor_buffer = val
+        else:
+            self._neighbor_buffer = None
```
mphoward, line 408: This branch is redundant:
```suggestion
        self._neighbor_buffer = val
```

PR #84, opened 2021-11-03
Enable arithmetic operations on Variables

### LINKED ISSUE #62: Variables as floats
We should implement all of the dunder functions needed to treat a variable as a floating-point value. That way, operations like:
```
a = DesignVariable(2.0)
>>> a.value + 3.0
5.0
```
can be done more simply as
```
>>> a + 3.0
5.0
```

adisr04:
- Implement dunder functions to allow `Variable` values to be used in arithmetic operations
- Implement dunder functions to allow `IndependentVariable` values to be used in in-place arithmetic operations
- Enable creation of `DependentVariable`s through arithmetic operations (`Sum`, `Difference`, `Product`, `Quotient`, `Power`, `Negation`)
- Add unit tests and documentation for new features

Resolves #62


relentless/variable.py
```diff
@@ -305,6 +305,65 @@ def athigh(self):
         """
         return self.state is self.State.HIGH
 
+    def __add__(self, val):
```
mphoward, line 308: These operations are actually more generic than just the `DesignVariables`. Hence, I recommend moving the methods as follows:

I think the binary operators (`__add__`, etc.) should be defined at the `Variable` level. If a value is not a `Variable`, you should coerce it into an `IndependentVariable`.

I think the unary infix operators (`__iadd__`, etc.) should be defined at the `IndependentVariable` level, and they should only allow regular numbers (and not `Variable`), as this would otherwise mean that we need to convert the variable into a `DependentVariable` However, infix operators may not change the type of the object itself. I think it is reasonable to scope this separation.

mphoward (changes requested):
This is a good start! I left a general comment on the organization of the methods. After you address this, and implement any appropriate tests, I will review more closely for correctness and corner cases.

relentless/variable.py
```diff
@@ -74,13 +86,79 @@ class Variable(abc.ABC):
...
         pass
 
+    def __add__(self, val):
+        """Addition of two variables, or of a variable and a scalar."""
+        v = val if isinstance(val, Variable) else IndependentVariable(val)
+        return Sum(self, val)
+
+    def __radd__(self, val):
+        """Addition of a scalar and a variable."""
+        return Sum(IndependentVariable(val), self)
```
mphoward, line 124: Is the reason `val` is always cast to an `IndependentVariable` because it must be something other than a `Variable` if `__radd__` gets called? Just asking because I don't remember when this is actually invoked in python.

relentless/variable.py
```diff
@@ -74,13 +86,79 @@ class Variable(abc.ABC):
...
     @property
     @abc.abstractmethod
     def value(self):
         """float: Value of the variable."""
         pass
 
+    def __add__(self, val):
+        """Addition of two variables, or of a variable and a scalar."""
+        v = val if isinstance(val, Variable) else IndependentVariable(val)
+        return Sum(self, val)
```
mphoward, line 120: Here and thereafter, you need to use `v` not `val`, or this typecast does nothing :)

relentless/variable.py
```diff
@@ -74,13 +86,79 @@ class Variable(abc.ABC):
...
 
+    Examples
+    --------
+    Create variables with specified value::
+
+        x = relentless.variable.IndependentVariable(value=2.0)
+
+    If the value is numerical, arithmetic operations are available::
+
+        >>> print(x+2.0)
```
mphoward, line 97: Should we add `__str__` to the `Variable`? I think these examples don't work as shown, since we haven't written a string converter. Maybe you can test and confirm? It would be something small like:
```
def __str__(self):
    return str(self.value)
```

relentless/variable.py
```diff
@@ -74,13 +86,79 @@ class Variable(abc.ABC):
...
+        4.0
+        >>> print(5.0-x)
+        1.0
+        >>> print(2.0*x)
+        2.0
+        >>> print(x/4.0)
+        0.5
+        >>> print(x^2.0)
```
mphoward, line 105: This is not valid python syntax
```suggestion
        >>> print(x**2)
```

relentless/variable.py
```diff
@@ -103,6 +181,21 @@ class IndependentVariable(Variable):
         >>> print(x.value)
         -1.0
 
+    Perform in-place arithmetic operations::
+
+        >>> x += 4.0
+        >>> print(x.value)
```
mphoward, line 187: If you do the string conversion, you can remove `.value` here

relentless/variable.py
```diff
@@ -559,6 +702,144 @@ class BinaryOperator(DependentVariable):
...
+
+    def _derivative(self, param):
+        if param == 'a':
+            try:
+                return 1.0/self.b.value
+            except ZeroDivisionError:
+                return numpy.inf
+        elif param == 'b':
+            try:
+                return -self.a.value/self.b.value**2.0
```
mphoward, line 808:
```suggestion
                return -self.a.value/self.b.value**2
```

relentless/variable.py
```diff
@@ -559,6 +702,144 @@ class BinaryOperator(DependentVariable):
...
+        Second value.
+
+    """
+    @property
+    def value(self):
+        return self.a.value**self.b.value
+
+    def _derivative(self, param):
+        if param == 'a':
+            return self.b.value*self.a.value**(self.b.value-1.0)
```
mphoward, line 833:
```suggestion
            return self.b.value*self.a.value**(self.b.value-1)
```
There should also be a special case for `b == 0`, right?

relentless/variable.py
```diff
@@ -559,6 +702,144 @@ class BinaryOperator(DependentVariable):
...
+    ----------
+    a : :class:`Variable`
+        First value.
+    b : :class:`Variable`
+        Second value.
+
+    """
+    @property
+    def value(self):
+        return self.a.value**self.b.value
```
mphoward, line 829: You might want to replace all of these with `np.power`:
```
numpy.power(self.a.value,self.b.value)
```
I think there was something less robust about `**`, but I don't remember what.

mphoward (changes requested):
This also looks really good! Here are a couple small things to address, and then I think we can merge this too.

relentless/variable.py
adisr04: Yes, `1+var` is interpreted as `var.__radd__(1)` . If both operands are `Variable`s then the usual `__add__` method can be used.

mphoward: OK, that's what I thought, thanks!

tests/test_variable.py
```diff
@@ -5,6 +5,60 @@
...
+        self.assertAlmostEqual(x.value, 16.0)
+        x = v**3.0
+        self.assertAlmostEqual(x.value, 8.0)
+
+        #negation
+        x = -w
+        self.assertAlmostEqual(x.value, -4.0)
+
+        #string conversion
+        self.assertEqual(str(x), str(x.value))
```
mphoward, line 60: Small thing, make this check the expected string:
```suggestion
        self.assertEqual(str(x), "4")
```
(or however this formats)

mphoward (changes requested):
One small thing on your new unit test, then this can be merged. Thanks!

PR #85, opened 2021-12-02
Use LAMMPS 29 Sep 2021 stable version

adisr04:
- Make all features compatible with LAMMPS 29 Sep 2021 stable version


relentless/simulate/lammps.py
```diff
@@ -820,8 +820,8 @@ def to_commands(self, sim):
                  ' v_ensemble_Lx v_ensemble_Ly v_ensemble_Lz'
                  ' v_ensemble_xy v_ensemble_xz v_ensemble_yz'
                  ' mode scalar ave running'
-                 ' file {filename} overwrite format "%.16e"').format(every=self.check_thermo_every,
-                                                                     filename=sim[self].thermo_file)
+                 ' file {filename} overwrite format "% .16e "').format(every=self.check_thermo_every,
```
mphoward, line 823: This looks funny, I think it should be like this? (To prepend the white space)
```suggestion
                 ' file {filename} overwrite format " %.16e"').format(every=self.check_thermo_every,
```

relentless/simulate/lammps.py
```diff
@@ -840,9 +840,9 @@ def to_commands(self, sim):
         cmds += ['compute ensemble_rdf all rdf {bins} {pairs}'.format(bins=sim[self].num_bins,pairs=' '.join(_pairs)),
                  ('fix ensemble_rdf_avg all ave/time {every} 1 {every}'
                   ' {computes} mode vector ave running off 1'
-                  ' file {filename} overwrite format "%.16e"').format(every=self.check_rdf_every,
-                                                                      computes=' '.join(_computes),
-                                                                      filename=sim[self].rdf_file)
+                  ' file {filename} overwrite format " % .16e "').format(every=self.check_rdf_every,
```
mphoward, line 843: Same here:
```suggestion
                  ' file {filename} overwrite format " %.16e"').format(every=self.check_rdf_every,
```

tests/simulate/test_generic.py
```diff
@@ -52,8 +53,9 @@ def test_invalid(self):
             sim = relentless.simulate.Simulation(self.ops)
             sim.run(self.ensemble, self.potentials, self.directory)
 
-    @unittest.skipIf(not relentless.simulate.lammps._lammps_found,
-                     "LAMMPS not installed")
+    @unittest.skipIf(not relentless.simulate.lammps._lammps_found or
```
mphoward, line 56: Which part of this test fails under python 3.8?

I am not sure that we should keep this test around anyway. LAMMPS has a Brownian integrator in more recent versions, so we might need to drop this.

tests/simulate/test_lammps.py
```diff
@@ -72,15 +71,13 @@ def test_initialize(self):
         op = relentless.simulate.lammps.InitializeFromFile(filename=file_)
         l = relentless.simulate.lammps.LAMMPS(operations=op, quiet=False)
         sim = l.run(ensemble=ens, potentials=pot, directory=self.directory)
-        pl = lammps.PyLammps(ptr=sim.lammps)
-        self.assertIsNotNone(pl.system)
```
mphoward, line 76: I'm not sure how this test is equivalent to what was here before. It might be good to do some assertion on `sim.lammps` first, before you then check if the number of atoms is nonzero.

tests/simulate/test_lammps.py
```diff
@@ -112,11 +107,11 @@ def test_integrators(self):
         lgv_r = relentless.simulate.lammps.RemoveLangevinIntegrator(add_op=lgv)
         l.operations = [init, lgv]
         sim = l.run(ensemble=ens, potentials=pot, directory=self.directory)
-        pl = lammps.PyLammps(ptr=sim.lammps)
-        self.assertEqual(pl.fixes[1]['style'], 'nve')
-        self.assertEqual(pl.fixes[2]['style'], 'langevin')
+        self.assertTrue(sim.lammps.has_id('fix',str(lgv._fix_nve)))
```
mphoward, line 110: Nice, this is much cleaner

.github/workflows/test.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
```
mphoward, line 26: Just to check: do these need to be strings because the yml template lops the trailing zero off 3.10 (treats as float)? That is good to know.

mphoward (changes requested):
Great work, just a few small things.

.github/workflows/test.yml
adisr04: Yes! See commit `f95bbd6`, it attempted to test in Python 3.1

tests/simulate/test_generic.py
adisr04: Ok, I have removed this test.

mphoward (approved):
Awesome, thanks for the fixes!

PR #86, opened 2021-12-02
Refactor stub modules

### LINKED ISSUE #72: Refactor stub modules
Right now `_math` and `data` are stubs. We might consider combining these back into some sort of `core` module or something like that rather than needing to remember so many different ones. It could be OK to leave it alone, though.

adisr04:
- Rename `_collections` and `_math` modules to `collections` and `math`
- Move `KeyedArray` class to `math` module
- Update all unit tests and documentation
- Combine documentation pages of `data`, `math`, and `collections` to "Developer tools"

Fixes #72


relentless/collections.py
```diff
@@ -0,0 +1,339 @@
+"""
+Collections
+===========
+
+This module contains some data structures used in :mod:`relentless` as an
+alternative to Python's general purpose containers, for ease of use in constructing
+potentials and performing computations during the optimization workflow.
```
mphoward, line 7: In the docs, this `:mod:` link is coming back to the developer page, not sure why. Suggesting this reword to avoid it:
```suggestion
This module contains some customized data structures used for constructing
potentials and performing computations during the optimization workflow.
```

(Edit: the link issue might get fixed once you fix the docstring per my comment above. In that case, you might not need to do anything here.)

doc/source/api/index.rst
```diff
@@ -1,5 +1 @@
-Overview
```
mphoward, line 1: Please revert this, we eventually do want to give some type of overview description of the documentation.

relentless/__init__.py
```diff
@@ -1,7 +1,20 @@
...
+Developer tools
+===============
+
+.. automodule:: relentless.collections
+
+.. automodule:: relentless.data
+
+.. automodule:: relentless.math
+
+"""
```
mphoward, line 11: This docstring is not correct, as it is supposed to document the whole `relentless` package but instead is documenting a subpart of it. Instead, can you create a page in the docs like `doc/source/api/devtools.rst` and put this content there? Then, you can link that page into the index, at the end of the list ideally.

mphoward (changes requested):
Thank you Adithya! Just a few small things to address with how the documentation is being rendered.

mphoward (approved):
Looks good, thanks!

PR #89, opened 2022-01-31
Exclude self-interactions in HOOMD RDF computation

### LINKED ISSUE #88: Test if HOOMD RDF excludes self interaction
When using freud in another script, I found that it was not properly excluding the self-interaction, so the RDF was getting a huge peak in the first bin. We should test if this bug is also present in `simulate.hoomd` and correct. The workaround I found was to do something like this:

```
query_args = dict(rdf[i,j].default_query_args)
query_args.update(exclude_ii=(i == j))
rdf.compute(
    system=(box,s.position[s.typeid == j]),
    query_points=s.position[s.typeid == i],
    neighbors=query_args,
    reset=False,
    )
```

because it seems to be necessary to respecify all of the arguments to neighbors

adisr04:
- Exclude self-interactions in RDF computations using HOOMD simulations.
- Add unit test to ensure there isn't a peak in the first bin.

Resolves #88

### DIFF
--- a/tests/simulate/test_hoomd.py
+++ b/tests/simulate/test_hoomd.py
@@ -248,20 +248,18 @@ class test_HOOMD(unittest.TestCase):
         ens = relentless.ensemble.Ensemble(T=2.0, V=relentless.volume.Cube(L=8.0), N={'A':2,'B':2})
         _,pot = self.ens_pot()
         init = relentless.simulate.hoomd.InitializeFromFile(filename=f.file.name)
-        lgv = relentless.simulate.hoomd.AddLangevinIntegrator(dt=0.1,
-                                                              friction=0.9,
-                                                              seed=2)
-        analyzer = relentless.simulate.hoomd.AddEnsembleAnalyzer(check_thermo_every=5,
-                                                                 check_rdf_every=5,
-                                                                 rdf_dr=1.0)
-        run = relentless.simulate.hoomd.Run(steps=500)
-        op = [init,lgv,analyzer,run]
+        ig = relentless.simulate.hoomd.AddVerletIntegrator(dt=0.0)
+        analyzer = relentless.simulate.hoomd.AddEnsembleAnalyzer(check_thermo_every=1,
+                                                                 check_rdf_every=1,
+                                                                 rdf_dr=0.1)
+        run = relentless.simulate.hoomd.Run(steps=1)
+        op = [init,ig,analyzer,run]
         h = relentless.simulate.hoomd.HOOMD(operations=op)
         sim = h.run(ensemble=ens, potentials=pot, directory=self.directory)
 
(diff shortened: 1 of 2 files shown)


tests/simulate/test_hoomd.py
```diff
@@ -234,6 +234,35 @@ def test_analyzer(self):
...
+            s.particles.position = [[-1,-1,-1],[1,1,1],[1,-1,1],[-1,1,-1]]
+            s.configuration.box = [8,8,8,0,0,0]
+            f.append(s)
+
+        ens = relentless.ensemble.Ensemble(T=2.0, V=relentless.volume.Cube(L=8.0), N={'A':2,'B':2})
+        _,pot = self.ens_pot()
+        init = relentless.simulate.hoomd.InitializeFromFile(filename=f.file.name)
+        lgv = relentless.simulate.hoomd.AddLangevinIntegrator(dt=0.1,
+                                                              friction=0.9,
+                                                              seed=2)
```
mphoward, line 253: For this test, I would recommend a trick to add an integrator that does nothing:
```suggestion
        ig = relentless.simulate.hoomd.AddNVEIntegrator(dt=0.0)
```

tests/simulate/test_hoomd.py
```diff
@@ -234,6 +234,35 @@ def test_analyzer(self):
...
+        analyzer = relentless.simulate.hoomd.AddEnsembleAnalyzer(check_thermo_every=5,
+                                                                 check_rdf_every=5,
+                                                                 rdf_dr=1.0)
+        run = relentless.simulate.hoomd.Run(steps=500)
```
mphoward, line 257: Then change this to run only 1 step. I am also recommending you decrease the RDF bin size.
```suggestion
        analyzer = relentless.simulate.hoomd.AddEnsembleAnalyzer(check_thermo_every=1,
                                                                 check_rdf_every=1,
                                                                 rdf_dr=0.1)
        run = relentless.simulate.hoomd.Run(steps=1)
```

tests/simulate/test_hoomd.py
```diff
@@ -234,6 +234,35 @@ def test_analyzer(self):
...
+        op = [init,lgv,analyzer,run]
+        h = relentless.simulate.hoomd.HOOMD(operations=op)
+        sim = h.run(ensemble=ens, potentials=pot, directory=self.directory)
+
+        ens_ = analyzer.extract_ensemble(sim)
+        for i,j in ens_.rdf:
+            self.assertLess(ens_.rdf[i,j].table[0,1], 20.0)
```
mphoward, line 264: This should test if it is identically zero, as in your initial configuration there are no particles on top of each other.

mphoward (changes requested):

mphoward (approved):
Thanks for the fix! Once the unit tests all pass, this is good to merge.

PR #90, opened 2022-02-17
Implement generic interface for energy minimizer

### LINKED ISSUE #49: Energy minimization operation
We should consider how we want to define the interface for the energy minimization operation. Different packages have drastically different methods for doing this minimization (sometimes supplying multiple options), and there are additional options required based on the method. What should be the "generic" implementation of this? How many args should we allow in the constructor, and how should we handle forwarding?

adisr04:
- Modify generic interface for backend `MinimizeEnergy` operation to allow for `options` dictionary argument
- Add keys to `options`: required `max_displacement` for HOOMD and optional `max_evaluations` for LAMMPS, `steps_per_iteration` for HOOMD
- Fix unit tests and documentation

Resolves #49.

### DIFF
--- a/relentless/simulate/hoomd.py
+++ b/relentless/simulate/hoomd.py
@@ -401,17 +401,22 @@ class MinimizeEnergy(simulate.SimulationOperation):
         Force convergence criterion.
     max_iterations : int
         Maximum number of iterations to run the minimization.
-    max_displacement : float
-        Maximum step size.
+    options : dict
+        Additional options for energy minimizer.
+
+    Raises
+    ------
+    ValueError
+        If a value for the maximum displacement is not provided.
 
     """
-    def __init__(self, energy_tolerance, force_tolerance, max_iterations, max_displacement):
+    def __init__(self, energy_tolerance, force_tolerance, max_iterations, options):
         self.energy_tolerance = energy_tolerance
         self.force_tolerance = force_tolerance
         self.max_iterations = max_iterations
-        self.max_displacement = max_displacement
-        if self.max_displacement is None:
-           raise ValueError('The HOOMD energy minimizer requires a value for the maximum displacement (dt).')
+        self.options = options
+        if 'max_displacement' not in self.options:
+           raise ValueError('HOOMD energy minimizer requires max_displacement option.')
 
     def __call__(self, sim):
         """Performs the energy minimization operation.
@@ -430,7 +435,7 @@ class MinimizeEnergy(simulate.SimulationOperation):
         """
         with sim.context:
             # setup FIRE minimization
-            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.max_displacement,
+            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.options['max_displacement'],
                                                          Etol=self.energy_tolerance,
                                                          ftol=self.force_tolerance)
             all_ = hoomd.group.all()
--- a/relentless/simulate/hoomd.py
+++ b/relentless/simulate/hoomd.py
@@ -401,15 +401,17 @@ class MinimizeEnergy(simulate.SimulationOperation):
         Force convergence criterion.
     max_iterations : int
         Maximum number of iterations to run the minimization.
-    dt : float
+    max_displacement : float
         Maximum step size.
 
     """
-    def __init__(self, energy_tolerance, force_tolerance, max_iterations, dt):
+    def __init__(self, energy_tolerance, force_tolerance, max_iterations, max_displacement):
         self.energy_tolerance = energy_tolerance
         self.force_tolerance = force_tolerance
         self.max_iterations = max_iterations
-        self.dt = dt
+        self.max_displacement = max_displacement
+        if self.max_displacement is None:
+           raise ValueError('The HOOMD energy minimizer requires a value for the maximum displacement (dt).')
 
     def __call__(self, sim):
         """Performs the energy minimization operation.
@@ -428,7 +430,7 @@ class MinimizeEnergy(simulate.SimulationOperation):
         """
         with sim.context:
             # setup FIRE minimization
-            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.dt,
+            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.max_displacement,
                                                          Etol=self.energy_tolerance,
                                                          ftol=self.force_tolerance)
             all_ = hoomd.group.all()
--- a/relentless/simulate/hoomd.py
+++ b/relentless/simulate/hoomd.py
@@ -396,9 +396,9 @@ class MinimizeEnergy(simulate.SimulationOperation):
     Valid ``options`` include:
 
     - ``max_displacement`` (`float`) - the maximum time step size the minimizer
-    is allowed to use.
+      is allowed to use.
     - ``max_evaluations`` (`int`) - the maximum number of time steps the minimizer
-    is allowed to run per iteration. Defaults to 100.
+      is allowed to run per iteration. Defaults to 100.
 
     Parameters
     ----------
--- a/relentless/simulate/hoomd.py
+++ b/relentless/simulate/hoomd.py
@@ -393,6 +393,13 @@ class InitializeRandomly(Initialize):
 class MinimizeEnergy(simulate.SimulationOperation):
     """Runs an energy minimzation until converged.
 
+    Valid ``options`` include:
+
+    - ``max_displacement`` (`float`) - the maximum time step size the minimizer
+    is allowed to use.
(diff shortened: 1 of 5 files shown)


relentless/simulate/hoomd.py
```diff
@@ -401,15 +401,17 @@ class MinimizeEnergy(simulate.SimulationOperation):
...
     """
-    def __init__(self, energy_tolerance, force_tolerance, max_iterations, dt):
+    def __init__(self, energy_tolerance, force_tolerance, max_iterations, max_displacement):
         self.energy_tolerance = energy_tolerance
         self.force_tolerance = force_tolerance
         self.max_iterations = max_iterations
-        self.dt = dt
+        self.max_displacement = max_displacement
+        if self.max_displacement is None:
+           raise ValueError('The HOOMD energy minimizer requires a value for the maximum displacement (dt).')
```
mphoward, line 414:
```suggestion
        self.options = options
        if 'max_displacement' not in self.options:
            raise KeyError('HOOMD energy minimizer requires max_displacement option.')
```

relentless/simulate/hoomd.py
```diff
@@ -428,7 +430,7 @@ def __call__(self, sim):
         """
         with sim.context:
             # setup FIRE minimization
-            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.dt,
+            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.max_displacement,
```
mphoward, line 433:
```suggestion
            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.options['max_displacement'],
```

relentless/simulate/lammps.py
```diff
@@ -445,11 +445,11 @@ class MinimizeEnergy(LAMMPSOperation):
...
     max_iterations : int
         Maximum number of iterations to run the minimization.
-    dt : float
-        Maximum step size.
+    max_displacement : float (defaults to `None`).
+        The maximum step size is not required by the LAMMPS energy minimizer.
 
     """
-    def __init__(self, energy_tolerance, force_tolerance, max_iterations, dt):
+    def __init__(self, energy_tolerance, force_tolerance, max_iterations, max_displacement=None):
```
mphoward, line 452:
```suggestion
    def __init__(self, energy_tolerance, force_tolerance, max_iterations, options=None):
```

tests/simulate/test_hoomd.py
```diff
@@ -82,11 +82,18 @@ def test_minimization(self):
               relentless.simulate.hoomd.MinimizeEnergy(energy_tolerance=1e-7,
                                                        force_tolerance=1e-7,
                                                        max_iterations=1000,
-                                                       dt=0.01)
+                                                       max_displacement=0.01)
```
mphoward, line 85:
```suggestion
                                                       options={'max_displacement': 0.01})
```

mphoward (changes requested):
Thanks! I think there was some confusion about `options` should be incorporated in the implementations. See below for suggested changes.

relentless/simulate/hoomd.py
```diff
@@ -401,15 +401,22 @@ class MinimizeEnergy(simulate.SimulationOperation):
...
         Maximum number of iterations to run the minimization.
-    dt : float
-        Maximum step size.
+    options : dict
+        Additional options for energy minimizer.
+
+    Raises
+    ------
+    ValueError
+        If a value for the maximum displacement is not provided.
```
mphoward, line 410: You should add a list of valid `options` to the documentation of this argument, and document what their type should be and what they do. You might be able to do that here with the argument by adding a bulleted list (Just be careful to render the docs and check it looks right), or else you could add a new section below.

relentless/simulate/lammps.py
```diff
@@ -445,14 +445,22 @@ class MinimizeEnergy(LAMMPSOperation):
         Force convergence criterion.
     max_iterations : int
         Maximum number of iterations to run the minimization.
-    dt : float
-        Maximum step size.
+    options : dict
```
mphoward, line 448: Same thing here as for the docs above

relentless/simulate/hoomd.py
```diff
@@ -401,15 +401,22 @@ class MinimizeEnergy(simulate.SimulationOperation):
...
     max_iterations : int
         Maximum number of iterations to run the minimization.
-    dt : float
-        Maximum step size.
+    options : dict
+        Additional options for energy minimizer.
+
+    Raises
+    ------
+    ValueError
```
mphoward, line 409: This should be a `KeyError` (not a `ValueError`) because the option is a key of the dictionary.

relentless/simulate/lammps.py
```diff
@@ -471,7 +479,7 @@ def to_commands(self, sim):
         cmds = ['minimize {etol} {ftol} {maxiter} {maxeval}'.format(etol=self.energy_tolerance,
                                                                     ftol=self.force_tolerance,
                                                                     maxiter=self.max_iterations,
-                                                                    maxeval=100*self.max_iterations)]
+                                                                    maxeval=self.options['max_evaluations'])]
```
mphoward, line 482: Do we want to give `max_evaluations` a default of 100 so that the user does not necessarily need to specify it? This would be the "typical" behavior in lammps. It would be done above like:
```
if 'max_evaluations' not in self.options:
    self.options['max_evaluations'] = 100
```
rather than raising a `KeyError`. I'm OK if you prefer this to be an explicit option.

relentless/simulate/hoomd.py
```diff
@@ -428,7 +435,7 @@ def __call__(self, sim):
         """
         with sim.context:
             # setup FIRE minimization
-            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.dt,
+            fire = hoomd.md.integrate.mode_minimize_fire(dt=self.options['max_displacement'],
```
mphoward, line 438: I can't tag it below, but line 447 is also like `max_evaluations` in that it runs 100 steps per iteration. Do you want to make this an option too? You could still use `max_evaluations`, and just document that it means the minimization runs for 100 steps each iteration.

mphoward (changes requested):

relentless/simulate/lammps.py
```diff
@@ -437,6 +437,11 @@ def to_commands(self, sim):
 class MinimizeEnergy(LAMMPSOperation):
     """Runs an energy minimization until converged.
 
+    Valid **options** include:
+
+    - **max_evaluations** (`int`) - the maximum number of time steps the minimizer
+      is allowed to run per iteration. Defaults to 100.
```
mphoward, line 443: This documentation is not quite right for LAMMPS. In LAMMPS, `max_evaluations` is:

> maxeval = max number of force/energy evaluations

I would recommend that the default is not fixed at 100. Instead, maybe set it to 100 / iteration (i.e., `max_evaluations = 100 * max_iterations`).

mphoward (changes requested):
One small final request!

relentless/simulate/lammps.py
mphoward: (You would need to modify the code for setting the default below too.)

relentless/simulate/lammps.py
```diff
@@ -445,14 +450,17 @@ class MinimizeEnergy(LAMMPSOperation):
...
 
     """
-    def __init__(self, energy_tolerance, force_tolerance, max_iterations, dt):
+    def __init__(self, energy_tolerance, force_tolerance, max_iterations, options):
         self.energy_tolerance = energy_tolerance
         self.force_tolerance = force_tolerance
         self.max_iterations = max_iterations
+        self.options = options
+        if 'max_evaluations' not in self.options:
+            self.options['max_evaluations'] = 100*self.max_iterations
```
mphoward, line 463: Small issue: this default will not update if `max_iterations` is changed after the operation is made. I suggest setting this to `None` here, then in the `to_commands` method compute `maxeval` if `self.options['max_evaluations'] is None`.

tests/simulate/test_hoomd.py
```diff
@@ -82,11 +82,26 @@ def test_minimization(self):
               relentless.simulate.hoomd.MinimizeEnergy(energy_tolerance=1e-7,
                                                        force_tolerance=1e-7,
                                                        max_iterations=1000,
-                                                       dt=0.01)
+                                                       options={'max_displacement':0.5,
+                                                                'max_evaluations':50})
```
mphoward, line 86:
```suggestion
                                                                'steps_per_iteration':50})
```

mphoward (changes requested):
OK, one final renaming, which I think I left as suggestions you can batch commit, then this should be good to merge.

mphoward (approved):
All done, thanks!

PR #93, opened 2022-03-07
Use cutoff in LAMMPS pair potential

adisr04:
Modify LAMMPS `Initialize.pair_map()` method to use per-pair cutoff values in order to remove unnecessary rows from the tabulated pair potentials

### DIFF
--- a/relentless/simulate/lammps.py
+++ b/relentless/simulate/lammps.py
@@ -333,21 +333,19 @@ class Initialize(LAMMPSOperation):
                         rcut[(i,j)] = r[1]
         else:
             rcut = None
-            rcut = sim.communicator.bcast(rcut)
+        rcut = sim.communicator.bcast(rcut)
 
         # process all lammps commands
         cmds = ['neighbor {skin} multi'.format(skin=sim.potentials.pair.neighbor_buffer)]
         cmds += ['pair_style table linear {N}'.format(N=Nr)]
 
-        cutoff = {}
         for i,j in sim.ensemble.pairs:
             # get lammps type indexes, lowest type first
             id_i,id_j = pair_map(sim,(i,j))
-            cutoff[(i,j)] = rcut[(i,j)] if rcut is not None else ''
             cmds += ['pair_coeff {id_i} {id_j} {filename} TABLE_{id_i}_{id_j} {cutoff}'.format(id_i=id_i,
                                                                                                id_j=id_j,
                                                                                                filename=file_,
-                                                                                               cutoff=cutoff[(i,j)])]
+                                                                                               cutoff=rcut[(i,j)])]
 
         return cmds
 


relentless/simulate/lammps.py
```diff
@@ -304,6 +304,7 @@ def pair_map(sim,pair):
         if sim.communicator.rank == sim.communicator.root:
             with open(file_,'w') as fw:
                 fw.write('# LAMMPS tabulated pair potentials\n')
+                rcut = {}
```
mphoward, line 307: This variable needs to be made available on all ranks, right now it is only defined on the root rank. To do this, you either need to broadcast from `root`, or you need to adapt the code so that `u` and `f` become available on all ranks. The first option requires fewer changes:
```py
if sim.communicator.rank == sim.communicator.root:
    rcut = {}
    with open(file_,'w') as fw:
        # all the existing code here
else:
    rcut = None
rcut = sim.communicator.bcast(rcut)
```

relentless/simulate/lammps.py
```diff
@@ -321,14 +322,24 @@ def pair_map(sim,pair):
                     f = sim.potentials.pair.force((i,j))[flags]
                     for idx in range(Nr):
                         fw.write('{idx} {r} {u} {f}\n'.format(idx=idx+1,r=r[idx],u=u[idx],f=f[idx]))
+                        # find the largest r with a nonzero potential
+                        if abs(u[idx])>0.0 and u[idx+1]==0.0:
+                            rcut_val = r[idx]
+
+                    key = '{id_i},{id_j}'.format(id_i=id_i,id_j=id_j)
+                    rcut[key] = rcut_val
```
mphoward, line 330: Just key it on `i,j` since below you are going to do the same loop
```suggestion
                    rcut[(i,j)] = rcut_val
```

relentless/simulate/lammps.py
```diff
@@ -321,14 +322,24 @@ def pair_map(sim,pair):
...
         cmds += ['pair_style table linear {N}'.format(N=Nr)]
         for i,j in sim.ensemble.pairs:
             # get lammps type indexes, lowest type first
             id_i,id_j = pair_map(sim,(i,j))
-            cmds += ['pair_coeff {id_i} {id_j} {filename} TABLE_{id_i}_{id_j}'.format(id_i=id_i,id_j=id_j,filename=file_)]
+            key = '{id_i},{id_j}'.format(id_i=id_i,id_j=id_j)
+            cmds += ['pair_coeff {id_i} {id_j} {filename} TABLE_{id_i}_{id_j} {rcut}'.format(id_i=id_i,
+                                                                                             id_j=id_j,
+                                                                                             filename=file_,
+                                                                                             rcut=rcut[key])]
```
mphoward, line 342:
```suggestion
            cmds += ['pair_coeff {id_i} {id_j} {filename} TABLE_{id_i}_{id_j} {rcut}'.format(id_i=id_i,
                                                                                             id_j=id_j,
                                                                                             filename=file_,
                                                                                             rcut=rcut[(i,j)])]
```

relentless/simulate/lammps.py
```diff
@@ -321,14 +322,24 @@ def pair_map(sim,pair):
                     f = sim.potentials.pair.force((i,j))[flags]
                     for idx in range(Nr):
                         fw.write('{idx} {r} {u} {f}\n'.format(idx=idx+1,r=r[idx],u=u[idx],f=f[idx]))
+                        # find the largest r with a nonzero potential
+                        if abs(u[idx])>0.0 and u[idx+1]==0.0:
+                            rcut_val = r[idx]
```
mphoward, line 327: This will sort of work, but it is inefficient and has some bad corner cases. For example, `rcut_val` will be undefined if this condition is never met. I would suggest instead doing this outside the loop with numpy. Something like this might work:
```py
# find r where potential and force are both zero
nonzero_r = numpy.flatnonzero(numpy.logical_and(~np.isclose(u,0), ~np.isclose(f,0)))
if len(nonzero_r) > 1:
    # cutoff at last nonzero r, must not be first r or LAMMPS will raise error
    rcut[(i,j)] = r[nonzero_r[-1]]
else:
    # 0 or 1 points are nonzero, cutoff at second r point
    rcut[(i,j)] = r[1]
```

mphoward (changes requested):
Good start! Here are some suggestions

relentless/simulate/lammps.py
```diff
@@ -322,13 +323,31 @@ def pair_map(sim,pair):
...
+                    nonzero_r = numpy.flatnonzero(numpy.logical_and(~numpy.isclose(u,0),~numpy.isclose(f,0)))
+                    if len(nonzero_r) > 1:
+                        # cutoff at last nonzero r (cannot be first r)
+                        rcut[(i,j)] = r[nonzero_r[-1]]
+                    else:
+                        # if first or second r is nonzero, cutoff at second
+                        rcut[(i,j)] = r[1]
+        else:
+            rcut = None
+            rcut = sim.communicator.bcast(rcut)
```
mphoward, line 336: This needs to be outside the `else` branch, as it must be called on all ranks
```suggestion
        rcut = sim.communicator.bcast(rcut)
```

relentless/simulate/lammps.py
```diff
@@ -322,13 +323,31 @@ def pair_map(sim,pair):
...
+
         # process all lammps commands
         cmds = ['neighbor {skin} multi'.format(skin=sim.potentials.pair.neighbor_buffer)]
         cmds += ['pair_style table linear {N}'.format(N=Nr)]
+
+        cutoff = {}
```
mphoward, line 342: This seems like a hack to fix the `rcut` being in the wrong place above. It should be removed
```suggestion
```

relentless/simulate/lammps.py
```diff
@@ -322,13 +323,31 @@ def pair_map(sim,pair):
...
         for i,j in sim.ensemble.pairs:
             # get lammps type indexes, lowest type first
             id_i,id_j = pair_map(sim,(i,j))
-            cmds += ['pair_coeff {id_i} {id_j} {filename} TABLE_{id_i}_{id_j}'.format(id_i=id_i,id_j=id_j,filename=file_)]
+            cutoff[(i,j)] = rcut[(i,j)] if rcut is not None else ''
```
mphoward, line 346:
```suggestion
```

relentless/simulate/lammps.py
```diff
@@ -322,13 +323,31 @@ def pair_map(sim,pair):
...
+            cmds += ['pair_coeff {id_i} {id_j} {filename} TABLE_{id_i}_{id_j} {cutoff}'.format(id_i=id_i,
+                                                                                               id_j=id_j,
+                                                                                               filename=file_,
+                                                                                               cutoff=cutoff[(i,j)])]
```
mphoward, line 350:
```suggestion
                                                                                               cutoff=rcut[(i,j)])]
```

mphoward (changes requested):

mphoward (approved):
Thank you!

PR #94, opened 2022-03-22
Refactor RelativeEntropy compute method

adisr04:
- Modify `RelativeEntropy.compute()` method to save simulation `Ensemble` to file
- Create `RelativeEntropy.compute_gradient()` method for direct computation of gradient using Ensemble
- Modify corresponding documentation and unit tests.

### DIFF
--- a/relentless/optimize/objective.py
+++ b/relentless/optimize/objective.py
@@ -51,7 +51,7 @@ To implement your own objective function, create a class that derives from
 .. autoclass:: RelativeEntropy
     :member-order: bysource
     :members: compute,
-        gradient,
+        compute_gradient,
         design_variables,
         target
 
@@ -266,7 +266,8 @@ class RelativeEntropy(ObjectiveFunction):
         as defined in :meth:`~relentless.simulate.simulate.Simulation.run()`,
         namely the simulation-generated ensemble, which is written to
         `ensemble.json`, and the values of the pair potential design variables,
-        which are written to ``potential.i.json`` for the :math:`i`\th pair potential.
+        which are written to ``pair_potential.i.json`` for the :math:`i`\th pair
+        potential.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


relentless/optimize/objective.py
```diff
@@ -279,9 +281,35 @@ def compute(self, directory=None):
...
+        # optionally write output to directory
+        if directory is not None and (self.communicator is None or self.communicator.rank == self.communicator.root):
+            for n,p in enumerate(self.potentials.pair.potentials):
+                p.save(directory.file('potential.{}.json'.format(n)))
+            sim_ens.save(directory.file('ensemble.json'))
+
+        # relative entropy *value* is None
+        return self.make_result(None, gradient, directory)
+
+    def gradient(self, sim_ens):
```
mphoward, line 298: Question: do we want this to be called `gradient` or `compute_gradient`? The former is shorter but also seems more like a property than a method.

In any case, suggesting a nicer argument name:
```suggestion
    def gradient(self, ensemble):
```

relentless/optimize/objective.py
```diff
@@ -279,9 +281,35 @@ def compute(self, directory=None):
...
+        """Computes the relative entropy gradient from a specified ensemble,
+        relative to the target.
```
mphoward, line 300:
```suggestion
        """Compute the relative entropy gradient for an ensemble.
```

tests/optimize/test_objective.py
```diff
@@ -218,6 +218,17 @@ def test_directory(self):
         self.assertAlmostEqual(x["('1', '1')"]['sigma'], self.sigma.value)
         self.assertAlmostEqual(x["('1', '1')"]['rmax'], 2.7)
 
+        sim = self.simulation.run(self.target, self.potentials, self.directory.name)
```
mphoward, line 221: We should also add a test (probably in its own test function, or else maybe in `test_compute`) that checks that the `gradient()` method works properly.

relentless/optimize/objective.py
```diff
@@ -279,9 +281,35 @@ def compute(self, directory=None):
...
         """
+        # run simulation and use result to compute gradient
         sim = self.simulation.run(self.target, self.potentials, directory, self.communicator)
         sim_ens = self.thermo.extract_ensemble(sim)
+        gradient = self.gradient(sim_ens)
+
+        # optionally write output to directory
+        if directory is not None and (self.communicator is None or self.communicator.rank == self.communicator.root):
+            for n,p in enumerate(self.potentials.pair.potentials):
+                p.save(directory.file('potential.{}.json'.format(n)))
```
mphoward, line 292: Looking at this, I am realizing that calling this "potential" is probably not the best idea, in case we add other types of potentials in future. Maybe we should rename the file to `pair_potential.*.json` or something like that?
```suggestion
                p.save(directory.file('pair_potential.{}.json'.format(n)))
```

mphoward (changes requested):
This looks very good, thank you! I have a couple comments and questions for you. We can discuss on Slack if you want, but I think we can get this wrapped up soon.

relentless/optimize/objective.py
```diff
@@ -261,10 +262,12 @@ def compute(self, directory=None):
...
         computationally expensive.
 
-        Optionally, a directory can be specified both to write the simulation
-        output as defined in :meth:`~relentless.simulate.simulate.Simulation.run()`,
-        and the values of the pair potential design variables, which are written
-        to ``potential.i.json`` for the :math:`i`\th pair potential.
+        Optionally, a directory can be specified to write the simulation output
+        as defined in :meth:`~relentless.simulate.simulate.Simulation.run()`,
+        namely the simulation-generated ensemble, which is written to
+        `ensemble.json`, and the values of the pair potential design variables,
```
mphoward, line 268:
```suggestion
        ``ensemble.json``, and the values of the pair potential design variables,
```

mphoward (approved):
Looks good, thank you! I will commit this minor edit to the docs and merge.

PR #114, opened 2022-10-14
Remove ensemble argument to simulations

### LINKED ISSUE #109: Remove ensemble from simulation run method
There isn't really a good reason to use this, I think, and it makes it awkward if you just want to do a run. I am going to inspect the code and see if there's a way to remove it.

### LINKED ISSUE #111: Silence freud warning about 2d boxes
freud generates a warning creating a 2d box with a z component. (`hoomd.py:761`). We should silence this warning by removing the z components of 2d boxes.

### LINKED ISSUE #113: LAMMPS integrator does not actually use timestep
This command is missing from the integrators, so it defaults to 0.005 in LJ units.

mphoward:
This PR refactors the simulation tools so that an `Ensemble` is not longer an argument of `run`. This is more convenient for the user, as the protocol is what defines the simulated ensemble and not some user input.

In order to enable this, I needed to refactor how the simulations asked the types in the simulations. The easiest way to do this was to split out the initialization operation from the rest of the operations. The initialization is now part of creating the `SimulationInstance`, i.e., after this instance is created, the system coordinates are loaded and the potentials are attached, so it is is "ready to run".

This also fixes some miscellaneous bugs:
* LAMMPS did not actually set the timestep in its integrators, so that is fixed here.
* LAMMPS may not have been setting the triclinic box properly. I think the definition is now correct, but we should verify it by visualization of the LAMMPS box from a dump file.
* freud's warnings about 2d boxes are now gone.

Fixes #109
Fixes #111 
Fixes #113


mphoward:
This needs a little bit more testing and to clean up the documentation, but otherwise it is ready to look at.

mphoward:
I would like to merge this on Monday afternoon, please provide any feedback by then.

adisr04 (approved):
Looks good! All the changes are clear and make sense to me. Was the LAMMPS triclinic box initialization also checked visually?

relentless/simulate/dilute.py
```diff
@@ -135,9 +164,24 @@ class Dilute(simulate.Simulation):
...
+        elif isinstance(ens.V, extent.Area):
+            sim.dimension = 2
+        sim.types = ens.types
+
+        return sim
+
     # initialization
-    InitializeFromFile = NullOperation
-    InitializeRandomly = NullOperation
+    # InitializeFromFile = simulate.NotImplementedOperation
```
clpetix, line 183: Should this be commented out?

mphoward:
Thanks! I did not, but we can ask @clpetix to do this in conjunction with PR #118!

clpetix (approved):
Looks good to me! I think this will help with ease-of-use!

relentless/simulate/dilute.py
mphoward: I commented it out to make it clear that this was intentional not to support InitializeFromFile as an operation. We could uncomment it to be explicit, but it is slightly redundant because it is already NotImplemented in the base class. Which do you prefer?

clpetix: That makes sense. I think this is fine since it still has NotImplemented in the base class!

Issue #81, opened 2021-11-01
Run LAMMPS outside python


mphoward:
The LAMMPS python API is a moving, unstable target. Since this API mostly just invokes a sequence of normal LAMMPS commands, it would be good if we could also compile these into a normal LAMMPS input file and invoke the LAMMPS binary. This should be much more stable, and also a common installation.

For many of our operations, this is as simple as calling `to_commands` and writing them to file. However, there are some operations (e.g., `AddLangevinIntegrator`) that call some LAMMPS python commands without generating an equivalent LAMMPS command. We would need to identify all of these spots and provide alternative script implementations (if possible) that can be run inside LAMMPS.

We also need to make sure that LAMMPS can be invoked properly from inside python when it is run under MPI (i.e., that there are no illegal process forks).

PR #118, opened 2022-10-16
Improve random initialization procedure

mphoward:
This PR enables smarter random initialization (i.e., without hard-sphere overlaps). I decided to write a method that fills the box by random sampling of a close-packed lattice. The particles are placed from biggest to smallest into random lattice sites that do not overlap any previously inserted particles. This should cover a lot of common use cases, but if dense enough packings are still not achieved, we could also write a separate method that uses packmol to do this.

This PR drops support for arbitrarily oriented Parallelepipeds and Parallelograms because we were essentially never using them, and I'm unaware of any packages that do. We could add support for initializing from files with rotation, if needed, at a later time.

This PR adds support for setting particle masses during random initialization, with thermalization handled correctly. Before, it only gave unit mass.

`lammpsio` is now a dependency for writing the Data File to initialize the LAMMPS simulation. This opens up the possibility of (1) getting the particle masses out of a file and (2) configuring the Langevin integrator using these masses, rather than extracting from a running simulation. This would allow us to use 100% LAMMPS script commands to do the simulation, and hence run from the command line (see #81).

This PR should be rebased and merged after PR #114 .

### DIFF
--- a/relentless/simulate/hoomd.py
+++ b/relentless/simulate/hoomd.py
@@ -90,9 +90,19 @@ class InitializeFromFile(Initialize):
 class InitializeRandomly(Initialize):
     """Initialize a randomly generated simulation box.
 
-    Particles are randomly placed in the box according to the specification.
-    No checks are performed for overlaps, so this can produce very poor starting
-    states for denser systems.
+    If ``diameters`` is ``None``, the particles are randomly placed in the box.
+    This can work pretty well for low densities, particularly if
+    :class:`MinimizeEnergy` is used to remove overlaps before starting to run a
+    simulation. However, it will typically fail for higher densities, where
+    there are many overlaps that are hard to resolve.
+
+    If ``diameters`` is specified for each particle type, the particles will
+    be randomly packed into sites of a close-packed lattice. The insertion
+    order is from big to small. No particles are allowed to overlap based on
+    the diameters, which typically means the initially state will be more
+    favorable than using random initialization. However, the packing procedure
+    can fail if there is not enough room in the box to fit particles using
+    lattice sites.
 
     Parameters
     ----------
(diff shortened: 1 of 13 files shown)


relentless/simulate/simulate.py
```diff
@@ -736,11 +755,100 @@ class InitializeRandomly(GenericOperation):
...
+            if dimension == 3:
+                # fcc lattice
+                a = numpy.sqrt(2.)*di
+                lattice = a*numpy.array([1.,1.,1.])
+                unitcell = a*numpy.array([[0.,0.,0.],[0.5,0.5,0.],[0.5,0.,0.5],[0.,0.5,0.5]])
+            elif dimension == 2:
+                a = di
+                b = numpy.sqrt(3.)*di
+                lattice = numpy.array([a,b])
+                unitcell = numpy.array([[0.,0.],[0.5*a,0.5*b]])
```
mphoward, line 816: I need to revisit these lattices and the shifting procedure to make sure I did it right.

mphoward:
Noting here: we are going to have to drop support for python 3.7 because `lammpsio` requires python 3.8 or higher. This is not a big loss because most packages are going to drop support any day now.

adisr04 (commented):
The PR looks good as of now! The FCC lattice constants and coordinate vectors are correct, although I am less sure about the subsequent shifting procedure (lines 819-825 in ``simulate.py``). Also, I think you also missed updating the ``/.github/workflows/test.yml`` to python 3.8.

relentless/simulate/hoomd.py
```diff
@@ -90,111 +90,104 @@ def initialize(self, sim):
...
-                    box_ = hoomd.data.boxdim(Lx=self.V.a[0],
-                                             Ly=self.V.b[1],
-                                             Lz=1,
-                                             xy=self.V.b[0]/self.V.b[1],
-                                             xz=0,
-                                             yz=0,
-                                             dimensions=2)
-                    box = freud.Box.from_box([box_.Lx, box_.Ly, 0, box_.xy, 0, 0], dimensions=2)
+            # randomly place particles in fractional coordinates
+            if mpi.world.rank == 0:
```
mphoward, line 153: This should be:
```suggestion
            if mpi.world.rank_is_root:
```

relentless/simulate/simulate.py
mphoward: This is correct as written, so I am cleaning up and commenting.

mphoward:
I think this PR is ready to go once Levi reviews it and does the offline testing of the initial configurations.

relentless/simulate/hoomd.py
```diff
@@ -90,111 +90,104 @@ def initialize(self, sim):
...
+                # generate the positions and types
+                if self.diameters is not None:
+                    positions, all_types = simulate.InitializeRandomly._pack_particles(self.seed, self.N, self.V)
```
clpetix, line 156: positions, all_types = simulate.InitializeRandomly._pack_particles(self.seed, self.N, self.V, self.diameters)
?

relentless/simulate/simulate.py
```diff
@@ -736,11 +754,103 @@ class InitializeRandomly(GenericOperation):
...
         Masses of each particle type. Defaults to None, which means particles
         have unit mass.
+    diameters : dict
+        Diameter of each particle type. Defaults to None, which means particles
+        are randomly inserted without checking their sizes.
 
     """
-    def __init__(self, seed, N, V, T=None, masses=None):
+    def __init__(self, seed, N, V, T=None, masses=None, diameters=None):
         super().__init__(seed, N, V, T, masses)
```
clpetix, line 763: super().__init__(seed, N, V, T, masses, diameters)

clpetix (changes requested):
I think the two changes I requested are required. With those, I was able to get it to work and used my dump class to view that it was properly putting down particles on lattice points. 

I imagine that it wasn't triggering the unittests because they are dilute enough to be put in randomly without overlap.

relentless/simulate/simulate.py
mphoward: Oh whoops! This is a big mistake, thanks for spotting it! I think the other code path you marked was probably never taken because it was defaulting the diameters to `None`. Otherwise, it should have been an error to miss the diameter argument. Let me take a look at the unit tests to see if we can add something to make sure this triggers.

mphoward:
> With those, I was able to get it to work and used my dump class to view that it was properly putting down particles on lattice points.

Thank you! Did you try visualizing both orthorhombic and triclinic boxes? I just want to make sure that the particles are being properly placed in the skewed boxes, since we had to make some changes in how those boxes were defined in LAMMPS. 

As an additional check on this, could you also verify that HOOMD and LAMMPS produce the same random state for a given orthorhombic or triclinic box? Since the seed to the generator is the same, the configurations should look identical after being dumped and loaded in OVITO.

clpetix:
738 in lammps.py
def __init__(self, initializer, operations=None, dimension=3, quiet=True):
This is defaulting the dimension that gives a mismatch error

relentless/simulate/lammps.py
```diff
@@ -144,74 +147,112 @@ def initialize(self, sim):
...
+        tilt = numpy.array([xy,xz,yz])
+        if not numpy.all(numpy.isclose(tilt,0)):
+            box = lammpsio.Box(lo,hi,tilt)
+        else:
+            box = lammpsio.Box(lo,hi)
+        snap = lammpsio.Snapshot(N=sum(self.N.values()), box=box)
+
+        # generate the positions and types
+        if self.diameters is not None:
+            positions, all_types = simulate.InitializeRandomly._pack_particles(self.seed, self.N, self.V)
```
clpetix, line 220: same thing here, needs diameters

clpetix:
I checked both ObliqueArea and TriclinicBox and it packed correctly. I also checked LAMMPS versus HOOMD on an ObliqueArea and it packed the same configuration.

clpetix:
> 

I checked both ObliqueArea and TriclinicBox and it packed correctly. I also checked LAMMPS versus HOOMD on an ObliqueArea and it packed the same configuration.

mphoward:
OK, yes, I think that is the designed behavior: the dimensionality of the LAMMPS simulation is fixed when the `LAMMPS` object gets created, and it is an error to try to change it afterwards (e.g., if dimension is 3, then you cannot use an `ObliqueArea`. if dimension is 2, then you cannot use a `TriclinicBox`).

mphoward:
With Levi's visual confirmation and our discussion at meeting today, I think this PR is finalized, so I am going to go ahead and merge so I can check on the other small PRs I opened.

PR #123, opened 2022-11-02
Fix error in pair potential derivative

mphoward:
I was unable to come up with a minimal unit test that reproduces the error Levi was seeing, but I am pretty sure this `if-if-else` should be an `if-elif-else` so that exactly one branch is taken when the parameter being differentiated is `rmin`.


mphoward:
Try to add a test where LJ = WCA + LJ tail, with rmax & rmin set on both

PR #124, opened 2022-11-02
Require explicit rmin for LAMMPS tabulated pair potential

### LINKED ISSUE #107: LAMMPS should not change rmin silently in tabulator
Currently, LAMMPS will automatically drop out the `r=0` entry from the `PairPotentialTabulator` when `rmin=0`. This may lead to confusing errors if particles get too close to each other, since the user will think that the potential is being tabulated to zero. Instead, `LAMMPS` should throw an error when it tries to attach the potentials if `rmin == 0`, and we should document this requirement in that class. (Explicit is better than implicit!)

mphoward:
This PR explicitly requires the user to set a nonzero rmin when using LAMMPS.

Fixes #107

### DIFF
--- a/relentless/simulate/hoomd.py
+++ b/relentless/simulate/hoomd.py
@@ -749,7 +749,7 @@ class HOOMD(simulate.Simulation):
     .. warning::
 
         HOOMD requires that tabulated pair potentials be finite. A common place to have an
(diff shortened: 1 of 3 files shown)


mphoward:
This is a small change, merging without review.

PR #125, opened 2022-11-02
Automatically generate LAMMPS IDs

### LINKED ISSUE #116: LAMMPS ensemble analyzer should not hard code ids
We should add methods to generate unique ids for computes, variables, groups, etc. to the `LAMMPSOperation`. There is no reason these need to be human-readable, and autogenerating will relax the requirement there is only one analyzer in the simulation.

mphoward:
This PR automatically generates IDs for computes, groups, and variables. This allows us to relax the requirement that there is only one ensemble analyzer, and lowers developer burden to carefully pick these IDs.

Fixes #116

### DIFF
--- a/relentless/simulate/lammps.py
+++ b/relentless/simulate/lammps.py
@@ -19,6 +19,7 @@ To implement your own LAMMPS operation, create an operation that derives from
 
 """
 import abc
+import uuid
 
 import lammpsio
 import numpy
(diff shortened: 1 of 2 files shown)


adisr04 (commented):
The changes look fine to me. Do you think adding a unit test case to showcase the functionality of multiple analyzers is needed?

mphoward:
Yes, I think we definitely should add that test, as your question made me realize a potential bug. (The same filename is used for the analyzer, so if two are created, they will overwrite each other.)

adisr04 (approved):
Looks good!

PR #126, opened 2022-11-03
Improve output order and content for RelativeEntropy

mphoward:
This PR improves the output of the `RelativeEntropy` compute. It now dumps the pair potential parameters *before* a run, which can be useful for debugging. It also adds a save method for the `ObjectiveFunctionResult`, which can be used to inspect the gradient after the fact. Last, it also fixes a bug where trying to use the compute without a directory could cause a crash if the backend needed a `Directory`. Now, it creates its own temporary directory if one is not specified, and tears it down after the simulation.

### DIFF
--- a/relentless/optimize/objective.py
+++ b/relentless/optimize/objective.py
@@ -54,6 +54,8 @@ To implement your own objective function, create a class that derives from
 
 """
 import abc
+import json
+import tempfile
 
 import numpy
 import scipy.integrate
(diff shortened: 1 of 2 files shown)


relentless/optimize/objective.py
```diff
@@ -176,6 +178,25 @@ def directory(self, value):
...
+
+        Parameters
+        ----------
+        filename : str
+            The name of the file to save data in.
+
+        """
+        data = {'variables': {x.name: v for x,v in self.variables.items()},
+                'value': self.value,
+                'gradient': {x.name: self.gradient[x] for x in self.variables},
```
mphoward, line 192:
```suggestion
                'gradient': {x.name: v for x,v in self.gradient.items()},
```

relentless/optimize/objective.py
```diff
@@ -176,6 +178,25 @@ def directory(self, value):
...
+                'directory': self.directory.path if self.directory is not None else None
```
mphoward, line 193: I need to check this in the unittest, I forgot it was a key.

PR #127, opened 2022-11-06
Refactor simulation operations

mphoward:
This PR aims to reduce the complexity of the simulation operations, possibly making it easier to extend to other simulation packages in future and possibly add other features like temperature annealing too.

The first big change is to combine the `Run` command with the various AddIntegrator-type commands. An integrator is really only meaningful in the context of a run, so there didn't seem to be a good reason to keep them separate. Hence, I fused these into three new run commands: `RunBrownianDynamics`, `RunLangevinDynamics`, and `RunMolecularDynamics`.

The only remaining add-style operation was then the `AddEnsembleAnalyzer`. This average is also only meaningful in the context of a run, so I added the concept of an `AnalysisOperation`, which is attached to a run and is invoked at the appropriate points of the run by two entry points: the `__call__` method, which handles normal setup stuff, and the `finalize` method, which allows the analysis to do any post-processing cleanup that is necessary. For example, the averaged ensemble is now accessible in the simulation data like `sim[analyzer].ensemble` rather than needing to be manually pulled by `analyzer.extract_ensemble(sim)`.

In support of the simplified interface, I refactored some of the generic operations into separate modules. The main reason for this is readability.


adisr04 (approved):
Thanks for this - looks ready to merge! Looking forward to using this streamlined simulation protocol.

PR #128, opened 2022-11-07
Refactor model subpackage

mphoward:
This PR refactors several related concepts into one `relentless.model` subpackage. The potentials are considered a subpackage of `model`, `relentless.model.potential`. The tests are refactored to match this structure.


adisr04 (approved):
Thanks, looks great!

PR #132, opened 2022-12-08
Trajectory dump operations

clpetix:
Add operators to dump trajectory files. Operation saves a trajectory to file while the simulation is running, for visualization/analysis purposes.


relentless/simulate/analyze.py
```diff
@@ -18,3 +18,29 @@ class EnsembleAverage(simulate.GenericAnalysisOperation):
...
+    ----------
+    filename : str
+        Name of the .gsd file to be written.
+    every : int
+        Interval of time steps at which to write a snapshot of the simulation.
+    velocity : bool
+        Log particle velocity.
+    image : bool
+        Log particle image.
+    typeid : bool
```
mphoward, line 36: We call these "types" in relentless. However, `type` is reserved... maybe use plurals like "velocities", "types", etc.?

relentless/simulate/analyze.py
```diff
@@ -18,3 +18,29 @@ class EnsembleAverage(simulate.GenericAnalysisOperation):
 
     def __init__(self, check_thermo_every, check_rdf_every, rdf_dr):
         super().__init__(check_thermo_every, check_rdf_every, rdf_dr)
+
+
+class WriteTrajectory(simulate.GenericAnalysisOperation):
+    """Writes a .gsd file.
```
mphoward, line 24: Just noting we will want to generalize these docs later.

relentless/simulate/hoomd.py
```diff
@@ -801,6 +801,53 @@ def reset(self):
...
+        if self.velocity is True or self.image is True:
+            _dynamic.append("momentum")
+
+        with sim.hoomd:
+            all_ = hoomd.group.all()
+            hoomd.dump.gsd(
+                self.filename,
+                period=self.every,
+                group=all_,
+                overwrite=True,
```
mphoward, line 843: This should be clearly documented so there are no nasty surprises.

relentless/simulate/hoomd.py
```diff
@@ -801,6 +801,53 @@ def reset(self):
...
+
+    def __call__(self, sim):
+        _dynamic = ["property"]
+        if self.velocity is True or self.image is True:
+            _dynamic.append("momentum")
+
+        with sim.hoomd:
+            all_ = hoomd.group.all()
+            hoomd.dump.gsd(
+                self.filename,
```
mphoward, line 840: This needs to be wrapped by the directory most of the time:
```suggestion
                sim.directory.file(self.filename),
```
We should decide if we want to optionally allow an absolute path that does not do this.

We should also make sure `sim.directory` is not None.

relentless/simulate/hoomd.py
```diff
@@ -801,6 +801,53 @@ def reset(self):
...
+                period=self.every,
+                group=all_,
+                overwrite=True,
+                dynamic=_dynamic,
+            )
+
+    def finalize(self, sim):
+        return super().finalize(sim)
```
mphoward, line 848: No need to reimplement
```suggestion
```

mphoward (changes requested):
Good start! Here are some comments.

relentless/simulate/hoomd.py
clpetix: If this not implemented I get an error. 

`Traceback (most recent call last):
  File "/Users/levipetix/Documents/research/projects/relentless_inverese_design/relentless/tests/simulate/test_hoomd.py", line 267, in test_writetrajectory
    sim = h.run(pot, self.directory)
  File "/Users/levipetix/Documents/research/projects/relentless_inverese_design/relentless/relentless/simulate/simulate.py", line 190, in run
    op(sim)
  File "/Users/levipetix/Documents/research/projects/relentless_inverese_design/relentless/relentless/simulate/simulate.py", line 71, in __call__
    return self._op(sim)
  File "/Users/levipetix/Documents/research/projects/relentless_inverese_design/relentless/relentless/simulate/hoomd.py", line 413, in __call__
    analyzer(sim)
  File "/Users/levipetix/Documents/research/projects/relentless_inverese_design/relentless/relentless/simulate/simulate.py", line 70, in __call__
    self._ensure_op(sim)
  File "/Users/levipetix/Documents/research/projects/relentless_inverese_design/relentless/relentless/simulate/simulate.py", line 86, in _ensure_op
    self._op = BackendOp(*self._args, **self._kwargs)
TypeError: Can't instantiate abstract class WriteTrajectory with abstract method finalize`

mphoward: Oh I see. This is because `finalize` is abstract in the parent class, so it needs to be implemented. However, the body should just be `pass` to mean "do nothing" rather than `super()`. I'm actually not sure what the latter behavior even is.

relentless/simulate/lammps.py
```diff
@@ -803,6 +803,75 @@ def finalize(self, sim):
         sim[self].ensemble = ens
 
 
+class WriteTrajectory(simulate.AnalysisOperation):
+    """Writes a LAMMPS dump file. When all options are set to True the file has
+    the following format:
+    ITEM: ATOMS x y z vx vy vz ix iy iz id mass
```
clpetix, line 809: id should be type. I will fix along with any other issues identified.

relentless/simulate/lammps.py
```diff
@@ -803,6 +803,75 @@ def finalize(self, sim):
...
+        if self.velocity is True:
+            _dynamic += "vx vy vz "
+        if self.image is True:
+            _dynamic += "ix iy iz "
+        if self.typeid is True:
+            _dynamic += "type "
+        if self.mass is True:
+            _dynamic += "mass "
+
+        cmds = ["dump myDump all custom " + str(self.every) + " " + filename + _dynamic]
```
clpetix, line 868: Probably better to do this with .format()
```suggestion
        cmds = ["dump myDump all custom {} {} {}".format(str(self.every), filename, _dynamic)]
```

mphoward (approved):

PR #139, opened 2023-01-09
Fix lammpsio to only raise import error when using LAMMPS.

clpetix:
An import error triggers if lammpsio is not installed, even if LAMMPS is not installed. The lammpsio import statement needs to be changed to only try to import lammpsio if LAMMPS is installed.


mphoward (approved):
Looks good, thanks!

PR #140, opened 2023-01-10
Fix LAMMPS pair potential

### LINKED ISSUE #135: LAMMPS miscellaneous fixes
* [x] The potential table needs to use r^2 rather than r. Otherwise, LAMMPS does cubic spline interpolation, which can produce very wrong results when the potential becomes large.
* [x] The cutoff may be off by one r value when potentials have cutoffs set at their minimum (e.g., WCA). We should scan these cutoffs when deciding where to cut the potential.
* [x] The default delay for rebuilding the neighbor list is 10, but it should be 0.
* [x] The `EnsembleAnalyzer` should return a 2d box for 2d simulations. It always returns a `TriclinicBox`.

mphoward:
Fixes #135

### DIFF
--- a/relentless/simulate/lammps.py
+++ b/relentless/simulate/lammps.py
@@ -915,17 +915,13 @@ class LAMMPS(simulate.Simulation):
         """
         if sim.potentials.pair.start == 0:
             raise ValueError("LAMMPS requires start > 0 for pair potentials")
-        r = sim.potentials.pair.x
+        rsq = sim.potentials.pair.xsquared
+        r = numpy.sqrt(rsq)
         Nr = len(r)
         if Nr == 1:
             raise ValueError(
                 "LAMMPS requires at least two points in the tabulated potential."
             )
-
-        # check that all r are equally spaced
-        dr = r[1:] - r[:-1]
-        if not numpy.all(numpy.isclose(dr, dr[0])):
-            raise ValueError("LAMMPS requires equally spaced r in pair potentials.")
 
         def pair_map(sim, pair):
             # Map lammps type indexes as a pair, lowest type first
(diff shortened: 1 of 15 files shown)


relentless/simulate/lammps.py
```diff
@@ -951,28 +955,42 @@ def pair_map(sim, pair):
...
                         )
 
                     # find r where potential and force are zero
-                    nonzero_r = numpy.flatnonzero(
-                        numpy.logical_and(~numpy.isclose(u, 0), ~numpy.isclose(f, 0))
-                    )
-                    if len(nonzero_r) > 1:
-                        # cutoff at last nonzero r (cannot be first r)
-                        rcut[(i, j)] = r[nonzero_r[-1]]
+                    all_rmax = [
```
clpetix, line 976: all_rmax's type is relentless.model.variable.Product

relentless/simulate/lammps.py
```diff
@@ -951,28 +955,42 @@ def pair_map(sim, pair):
...
+                        pair_pot.coeff[i, j]["rmax"]
+                        for pair_pot in sim.potentials.pair.potentials
+                    ]
+                    if None not in all_rmax:
+                        # use rmax if set for all potentials
+                        rcut[(i, j)] = min(max(all_rmax), sim.potentials.pair.stop)
```
clpetix, line 982: Can't take the min of float & product
`TypeError: '<' not supported between instances of 'float' and 'Product'`

relentless/simulate/lammps.py
mphoward: We need to make sure we evaluate these if it's a Variable.

mphoward:
The doc builds are failing, I think because RTD is using a newer version of sphinx than we do. It may not be possible to actually pin the sphinx version on RTD, so I will updated our requirements and see if I can reproduce & fix.

clpetix (approved):
Looks good!

PR #143, opened 2023-01-10
Add Python 3.11 to testing

mphoward:
* Add testing for Python 3.11
* Refactor tests to run those for simulation engines in isolation. This allows us to tailor the dependencies for each package with finer detail (e.g., not use a certain version of Python), and we make sure that everything can run when only zero or one engines are installed. LAMMPS does not have a conda package for 3.11, which is why this change was needed.
* Test the newest stable release of LAMMPS in addition to the oldest supported.
* Supply simulation dependencies using miniconda with explicit channel. mamba would fail to configure the environment repeatedly.
* Refactor code into a `src/` directory so that testing can be done inside a virtual environment. Run `pip install -e .` in your environment to get an editable installation for testing.


tests/simulate/test_hoomd.py
```diff
@@ -2,27 +2,19 @@
...
 
-from parameterized import parameterized_class
-
-try:
-    import gsd.hoomd
-
-    _found_gsd = True
-except ImportError:
-    _found_gsd = False
+import gsd.hoomd
```
clpetix, line 5: Won't this trigger an import error if not installed? Do we want the test to skip if it doesn't have hoomd or freud but raise an error without gsd.

tests/requirements.txt
```diff
@@ -1,5 +1,2 @@
-hoomd=2.9.7
-freud>=2
 gsd
```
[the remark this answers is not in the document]
mphoward: GSD is a testing-only dependency (whereas the others are runtime dependencies) and it is a small package, so I thought it's OK to always include it in the testing requirements. We expect developers to install all the packages here, so then we can safely assume we can import it later.

tests/simulate/test_hoomd.py
mphoward: Yes, we would want an error to be raised so the developer realizes they should have installed the testing dependencies.

clpetix (approved):
Looks good to me!

PR #146, opened 2023-01-13
Refactor simulation operations (again)

mphoward:
This PR fixes several clunky designs of the simulation operations:
* The operations we want the user to call no longer wrap around a backend operation. Instead, they construct a delegated operation at the time they are called. This means that these operations guarantee that they will construct the right delegated operation.
* This fix helps resolve issues with data storage, as the meaning of "self" is ambiguous in the data. Data created by delegated operations is now pilfered and destroyed so that the operation the user creates owns the data.
* To make this simpler, the data is now stored in plain dictionaries rather than classes. That way, we don't have to worry about creating a specific data type.
* The delegate operations are private members of the Simulation classes.
* Analysis operations are refactored into three phases: pre-run, post-run, and process. The pre-run/post-run call should be performed right before/after the parent operation executes a "run". These should mostly handle setup/destruction. The process call occurs after all simulation operations have executed, which gives time to process files, etc. at the end of a workflow.
* Analysis operations are now made aware of the parent operation calling them. This is not currently used for anything but could help optimize some analysis.
* Initializers actually handle the full initialization now. Simulations should only override `_new_instance` if they need to do something wonky (like LAMMPS).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
           python -m unittest discover -v -s ./tests -p 'test_*.py'
 
   test-hoomd:
-    name: unit test [python-${{ matrix.python-version }},hoomd-2.9.7]
+    name: unit test [python-${{ matrix.python-version }}, hoomd-2.9.7]
     needs: test
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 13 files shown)


clpetix (approved):
This looks good to me!

PR #147, opened 2023-01-13
Expand random filling size

### LINKED ISSUE #130: Random initialization box volume too small?
The current code may be too conservative in choosing the orthorhombic box. It may be possible to use the diagonal of the matrix instead. We should check this and, if allowed, expand the box.

mphoward:
The equivalent orthorhombic box _should_ be given by the diagonal of the matrix of lattice vectors because the tilt factors make the lattice vectors longer.

Fixes #130


mphoward:
Levi verified this works in some 2d tests.

PR #148, opened 2023-01-13
Run tests under MPI

mphoward:


mphoward:
I'm not happy with these MPI runners, I think we may need to investigate using a container or another solution if we want to do MPI testing. Closing.

PR #149, opened 2023-01-16
Run tests with MPI

### LINKED ISSUE #121: Run unit tests under MPI
All sorts of bugs creep in when you don't do this automatically... we desperately need to get this set up on GitHub actions. We might need to build our own container that has HOOMD and LAMMPS stuffed in it, along with a compatible MPI library. We should then be able to install all of our python requirements from pip.

### LINKED ISSUE #150: Possible I/O race condition
There may be a race condition in MPI-enabled simulations when temporary directories are used. The cleanup() method is called only by the root rank. This means the directory contents could possibly be wiped out while another rank is still using them. We should put an MPI barrier in place to ensure everyone is done with the contents before cleanup is called.

This is mostly present in unit tests, but there is at least one usage in the code (RelativeEntropy).

I am not sure if this might explain some spurious CI failures under MPI.

mphoward:
This PR enables MPI testing of:

- [X] Core features
- [x] LAMMPS

HOOMD is not able to be tested with MPI because the conda build lacks support. We would need to use a container built with MPI to test this in future.

In the process of running the tests, I noticed race-like conditions on the I/O. I have refactored the `Directory` object and tests to try to address these issues. The caller now has more control over who makes the directories, and there are some additional synchronization barriers to ensure required files are in place before they are consumed. We may need to keep an eye on the code to see if there are any other I/O points that have similar issues, but the tests seem much more robust now.

Fixes #121 
Fixes #150

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+      - name: Setup MPI
+        uses: mpi4py/setup-mpi@v1
+        with:
+          mpi: 'openmpi'
       - name: Setup Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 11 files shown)


clpetix (approved):
This all looks good to me!

PR #153, opened 2023-01-18
Load potentials from JSON files

mphoward:
This PR improves the way that potentials are saved to file, and enables loading them from JSON data or from a JSON file.


clpetix:
I have tested this out by reloading a potential that I saved, and it worked perfectly! The code all looks good to me!

clpetix (approved):

PR #154, opened 2023-01-20
Fix moving contents of directory

clpetix:
This fixes the `move_contents` function to ensure that it can overwrite, if desired. The `move_contents` testing has been expanded to include a case where it will overwrite a file and directory. 

 `SteepestDescent`'s  `optimize` method was changed to overwrite the directory, if specified, before writing the outputs. This ensures that the directory is always initially empty. This will prevent a confusing situation that could occur when running an optimization and then increasing the tolerance and rerunning. In this scenario, the optimizer will converge in fewer iterations, only rewriting the data until convergence. This would leave mixed data from the two runs which could cause confusion to users.

### DIFF
--- a/src/relentless/data.py
+++ b/src/relentless/data.py
@@ -249,15 +249,12 @@ class Directory:
                 shutil.copytree(entry.path, os.path.join(dest.path, entry.name))
 
     def is_empty(self):
-        """Returns whether the directory is empty.
+        """Check if the directory is empty.
 
         Returns
         -------
-        :Bool
+        :bool
             True if the directory is empty.
 
         """
-        empty = True
-        if len(os.listdir(self.path)) != 0:
-            empty = False
-        return empty
+        return len(os.listdir(self.path)) != 0
--- a/src/relentless/data.py
+++ b/src/relentless/data.py
@@ -12,6 +12,7 @@ Data management (`relentless.data`)
 
 """
 import os
+import pathlib
 import shutil
 
 
@@ -198,27 +199,38 @@ class Directory:
             elif entry.is_dir():
                 shutil.rmtree(entry.path)
 
-    def move_contents(self, dest, overwrite=True):
+    def move_contents(self, dest):
         """Move the contents of the directory.
 
         Parameters
         ----------
         dest : :class:`Directory` or :class:`str`
             Destination directory.
-        overwrite : bool
-            Clear destination directory before moving contents.
+
+        Raises
+        ------
+        OSError
+            If the destination exists and does not match the type of the source.
 
         """
         dest = Directory.cast(dest, create=True)
-        if overwrite is True:
-            dest.clear_contents()
-        try:
-            for entry in os.scandir(self.path):
-                shutil.move(entry.path, dest.path)
-        except Exception:
-            raise FileExistsError(
-                "File and/or directory already exists and overwrite is False"
-            )
+        for entry in os.scandir(self.path):
+            dest_entry = pathlib.Path(os.path.join(dest.path, entry.name))
+            if dest_entry.exists():
+                if dest_entry.is_dir() and entry.is_dir():
+                    shutil.copytree(entry.path, dest_entry, dirs_exist_ok=True)
+                    shutil.rmtree(entry.path)
+                elif dest_entry.is_file() and entry.is_file():
+                    shutil.move(entry.path, dest_entry)
+                else:
+                    raise OSError(
+                        "Destination "
+                        "{} exists and does not match type of source {}".format(
+                            dest_entry, entry.name
+                        )
+                    )
+            else:
+                shutil.move(entry.path, dest_entry)
 
     def copy_contents(self, dest):
         """Copy the contents of the directory.
--- a/src/relentless/data.py
+++ b/src/relentless/data.py
@@ -198,18 +198,27 @@ class Directory:
             elif entry.is_dir():
                 shutil.rmtree(entry.path)
 
-    def move_contents(self, dest):
+    def move_contents(self, dest, overwrite=True):
         """Move the contents of the directory.
 
         Parameters
         ----------
         dest : :class:`Directory` or :class:`str`
             Destination directory.
+        overwrite : bool
+            Clear destination directory before moving contents.
 
         """
         dest = Directory.cast(dest, create=True)
-        for entry in os.scandir(self.path):
-            shutil.move(entry.path, dest.path)
+        if overwrite is True:
+            dest.clear_contents()
+        try:
+            for entry in os.scandir(self.path):
+                shutil.move(entry.path, dest.path)
+        except Exception:
+            raise FileExistsError(
+                "File and/or directory already exists and overwrite is False"
+            )
 
     def copy_contents(self, dest):
         """Copy the contents of the directory.
--- a/src/relentless/data.py
+++ b/src/relentless/data.py
@@ -215,7 +215,7 @@ class Directory:
         """
         dest = Directory.cast(dest, create=True)
         for entry in os.scandir(self.path):
-            dest_entry = pathlib.Path(os.path.join(dest.path, entry.name))
+            dest_entry = pathlib.Path(dest.path, entry.name)
             if dest_entry.exists():
                 if dest_entry.is_dir() and entry.is_dir():
                     shutil.copytree(entry.path, dest_entry, dirs_exist_ok=True)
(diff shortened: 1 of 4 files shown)


src/relentless/data.py
```diff
@@ -198,18 +198,27 @@ def clear_contents(self):
             elif entry.is_dir():
                 shutil.rmtree(entry.path)
 
-    def move_contents(self, dest):
+    def move_contents(self, dest, overwrite=True):
```
mphoward, line 201: I thought we were initially planning to leave this function alone? One issue with the current `overwrite` implementation is that if the destination has some extra content (that would not be overwritten), it gets wiped out.

I propose to remove the `overwrite` option and make this function have some nicer behavior that is more typical of Unix commands, since we are looking at it now:
```
for entry in os.scandir(self.path):
    dest_entry = pathlib.Path(os.path.join(dest.path, entry.name))
    if dest_entry.exists():
        if dest_entry.is_dir() and entry.is_dir():
            shutil.copytree(entry.path, dest_entry.path, dirs_exist_ok=True)
            shutil.rmtree(entry.path)
        elif dest_entry.is_file() and entry.is_file():
            shutil.move(entry.path, dest_entry.path)
        else:
           raise OSError('Destination {} exists and does not match type of source {}'.format(dest_entry.path, entry.name))
    else:        
        shutil.move(entry.path, dest_entry.path)
```

src/relentless/optimize/method.py
```diff
@@ -351,6 +352,7 @@ def optimize(self, objective, design_variables, directory=None):
 
         if directory is not None:
             directory = data.Directory.cast(directory, create=mpi.world.rank_is_root)
+            directory.clear_contents()
```
mphoward, line 355: Do we want to have an `overwrite` option here? We could check if `directory` is not empty, then if `overwrite` is True, clear contents; otherwise, raise an error.

mphoward (changes requested):
Thanks! I have two suggestions to look into.

src/relentless/optimize/method.py
```diff
@@ -335,13 +337,20 @@ def optimize(self, objective, design_variables, directory=None):
         directory : str or :class:`~relentless.data.Directory`
             Directory for writing output during optimization. Default of `None`
             requests no output is written.
+        overwrite : bool
```
mphoward, line 340: You should add `overwrite` to the base class `optimize()` method as well. I think it would also then be a good idea to move the logic for setting up the directory into a private method of `_Optimizer` so it can be reused if we implement other classes in future. Something like:
```
def _setup_directory(self, directory, overwrite):
    ...
```
doing what's in lines 361-372.

src/relentless/optimize/method.py
```diff
@@ -351,6 +360,15 @@ def optimize(self, objective, design_variables, directory=None):
 
         if directory is not None:
             directory = data.Directory.cast(directory, create=mpi.world.rank_is_root)
+            if len(os.listdir(directory.path)) != 0:
+                if overwrite is True:
+                    directory.clear_contents()
```
mphoward, line 365: I think this needs to be guarded so that only the root rank calls it. It will be safe to not put another barrier before it because we doing `create` on the root only.
```suggestion
                    if mpi.world.rank_is_root:
                        directory.clear_contents()
```

tests/test_data.py
```diff
@@ -189,6 +189,25 @@ def test_move_contents(self):
...
+
+        # move file and directory from foo to baz
+        if relentless.mpi.world.rank_is_root:
+            dfoo.move_contents(baz)
+        relentless.mpi.world.barrier()
+        self.assertFalse(os.path.isfile(os.path.join(foo, "spam.txt")))
+        self.assertFalse(os.path.isdir(os.path.join(foo, "fizz")))
+        self.assertTrue(os.path.isfile(os.path.join(baz, "spam.txt")))
+        self.assertTrue(os.path.isdir(os.path.join(baz, "fizz")))
+
```
mphoward, line 210: There should be a short unit test added for the `overwrite` option of the optimizers in that file.

src/relentless/data.py
```diff
@@ -226,3 +247,17 @@ def copy_contents(self, dest):
...
+
+        Returns
+        -------
+        :Bool
+            True if the directory is empty.
+
+        """
+        empty = True
+        if len(os.listdir(self.path)) != 0:
+            empty = False
```
mphoward, line 262:
```suggestion
        return len(os.listdir(self.path)) != 0
```

src/relentless/data.py
```diff
@@ -226,3 +247,17 @@ def copy_contents(self, dest):
                 shutil.copy2(entry.path, dest.path)
             elif entry.is_dir():
                 shutil.copytree(entry.path, os.path.join(dest.path, entry.name))
+
+    def is_empty(self):
+        """Returns whether the directory is empty.
+
+        Returns
+        -------
+        :Bool
```
mphoward, line 256:
```suggestion
        bool
```

src/relentless/data.py
```diff
@@ -226,3 +247,17 @@ def copy_contents(self, dest):
                 shutil.copy2(entry.path, dest.path)
             elif entry.is_dir():
                 shutil.copytree(entry.path, os.path.join(dest.path, entry.name))
+
+    def is_empty(self):
+        """Returns whether the directory is empty.
```
mphoward, line 252:
```suggestion
        """Check if the directory is empty.
```

src/relentless/optimize/method.py
```diff
@@ -38,6 +38,8 @@ def optimize(self, objective, design_variables, directory=None):
         directory : str or :class:`~relentless.data.Directory`
             Directory for writing output during optimization. Default of ``None``
             requests no output is written.
+        overwrite : bool
+            Overwrites the directory, if specified, before begining optimization.
```
mphoward, line 42:
```suggestion
            If ``True``, overwrite the directory before beginning optimization.
```

src/relentless/optimize/method.py
```diff
@@ -54,6 +56,21 @@ def stop(self, value):
             raise TypeError("The stopping criterion must be a ConvergenceTest.")
         self._stop = value
 
+    def setup_directory(self, directory=None, overwrite=False):
```
mphoward, line 59: Let's make this private since we don't want the user to call it. Also, let's require the user to specify both arguments:
```suggestion
    def _setup_directory(self, directory, overwrite):
```

src/relentless/optimize/method.py
```diff
@@ -54,6 +56,21 @@ def stop(self, value):
...
+                if overwrite is True:
+                    if mpi.world.rank_is_root:
+                        directory.clear_contents()
+                else:
+                    raise OSError(
+                        "Directory {} is not empty and overwrite is not True".format(
+                            directory.path
+                        )
+                    )
+            mpi.world.barrier()
```
mphoward, line 72: I think the error might be because you are trying to check the directory contents on other ranks before it is created. By moving the barrier up, we can ensure it exists before calling the other operations. Then, we clear the contents only the root rank and put up another barrier if we need to clear.
```suggestion
        directory = data.Directory.cast(directory, create=mpi.world.rank_is_root)
        mpi.world.barrier()
        if not directory.is_empty():
            if overwrite is True:
                if mpi.world.rank_is_root:
                    directory.clear_contents()
                mpi.world.barrier()
            else:
                raise OSError(
                    "Directory {} is not empty and overwrite is not True".format(
                        directory.path
                    )
                )
        
```

src/relentless/optimize/method.py
```diff
@@ -335,23 +353,28 @@ def optimize(self, objective, design_variables, directory=None):
...
         design_variables = variable.graph.check_variables_and_types(
             design_variables, variable.DesignVariable
         )
         if len(design_variables) == 0:
             return None
 
-        if directory is not None:
-            directory = data.Directory.cast(directory, create=mpi.world.rank_is_root)
-            mpi.world.barrier()
+        self.setup_directory(directory, overwrite)
```
mphoward, line 377:
```suggestion
        if directory is not None:
            self._setup_directory(directory, overwrite)
```

mphoward (changes requested):
I think this should fix the MPI issue but let me know if it doesn't. I also caught a couple typos.

tests/test_data.py
```diff
@@ -189,6 +189,47 @@ def test_move_contents(self):
...
+        )
+        dbazfizz = relentless.data.Directory(
+            bazfizz, create=relentless.mpi.world.rank_is_root
+        )
+
+        if relentless.mpi.world.rank_is_root:
+            open(dfoofizz.file("buzz.txt"), "w").close()
+            open(dfoofizz.file("fred.txt"), "w").close()
+            open(dbazfizz.file("buzz.txt"), "w").close()
+            open(dbazfizz.file("jim.txt"), "w").close()
```
mphoward, line 217:
```suggestion
            open(dbazfizz.file("jim.txt"), "w").close()
        relentless.mpi.world.barrier()
```

mphoward (approved):
I had one small tweak that I committed, so once tests pass this is good to go!

PR #160, opened 2023-02-02
Simplify parameter setting

### LINKED ISSUE #158: Simplify pair parameters
The shared and per-type parameters are basically never going to need to be used because they can handled more flexibly at the user level (e.g., binding to a variable or a dict), and they make the code unnecessarily complex. I will remove these code paths.

mphoward:
Removes shared parameters from `Parameters` and `PairParameters`. Also removes per-type parameters from `PairParameters`. These features don't really add much when you can code them in normal Python anyway, and it makes the objects harder to understand / load from file.

Fixes #158


clpetix:
This all looks good to me!

clpetix (approved):

PR #161, opened 2023-02-02
Allow LAMMPS to run as an executable

### LINKED ISSUE #81: Run LAMMPS outside python

mphoward:
This PR enables running LAMMPS as an executable rather than a Python library. MPI is supported with the restriction that relentless itself cannot be launched with MPI. This should be fine since only the simulation really needs it.

To implement this, the following major changes were made:

* Create an `InitializationOperation` type that guarantees certain data will be present after it runs.
* Set the masses of different particle types during initialization. This was the only spot where LAMMPS was requiring immediate feedback from the MPI library.
* Add a LAMMPSAnalysisOperation type that breaks its calls into running and commands, e.g. `pre_run` and `pre_run_commands`.
* Only run LAMMPS commands through the Python interface when we are in that run configuration. Otherwise, just save the commands.
* Add a `_post_run` method to the simulations. By default, this executes the `process` command of all detected analyzers. However, LAMMPS now leverages this step to write all the commands to an input file, then run it with the executable.
* To avoid some nasty fork stuff when `mpi4py` is installed but the process is not being run under MPI, `mpi4py` is only loaded when we detect the program is being run under MPI due to one of the environment variables.
* Tests are run on multiple MPI implementations, as what works with MPICH does not always work with OpenMPI.
* LAMMPS is tested as both a Python library and executable.

Closes #81

### DIFF
--- a/src/relentless/simulate/lammps.py
+++ b/src/relentless/simulate/lammps.py
@@ -1109,9 +1109,12 @@ class LAMMPS(simulate.Simulation):
             file_ = mpi.world.bcast(file_)
 
             # only launch process on the root rank with MPI launcher
-            run_cmd = sim[sim.initializer]["_lammps"] + ["-i", file_]
             if mpi.world.rank_is_root:
-                result = subprocess.run("mpirun " + " ".join(run_cmd), shell=True)
+                run_cmd = sim[sim.initializer]["_lammps"] + ["-i", file_]
+                run_cmd = " ".join(run_cmd)
+                if mpi.world.enabled:
+                    run_cmd = "mpiexec " + run_cmd
+                result = subprocess.run(run_cmd, shell=True)
                 returncode = result.returncode
             else:
                 returncode = None
--- a/src/relentless/simulate/lammps.py
+++ b/src/relentless/simulate/lammps.py
@@ -628,7 +628,7 @@ class RunLangevinDynamics(_MDIntegrator):
         if Ntypes > 1:
             scale = damp / damp_ref
             scale_str = " ".join(
-                ["scale {} {}".format(i + 1, s) for i, s in enumerate(scale[1:])]
+                ["scale {} {}".format(i + 2, s) for i, s in enumerate(scale[1:])]
             )
         else:
             scale_str = ""
--- a/src/relentless/simulate/lammps.py
+++ b/src/relentless/simulate/lammps.py
@@ -1,5 +1,5 @@
 import abc
-import shutil
+import datetime
 import subprocess
 import uuid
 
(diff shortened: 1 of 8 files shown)


src/relentless/simulate/lammps.py
```diff
@@ -555,30 +607,28 @@ def __init__(self, steps, timestep, T, friction, seed, analyzers):
...
+
         # compute per-type damping parameter and rescale if multiple types
         damp = numpy.divide(mass, friction, where=(friction > 0))
-        damp_ref = damp[1]
+        damp_ref = damp[0]
         if Ntypes > 1:
             scale = damp / damp_ref
             scale_str = " ".join(
-                ["scale {} {}".format(i + 1, s) for i, s in enumerate(scale[2:])]
+                ["scale {} {}".format(i + 1, s) for i, s in enumerate(scale[1:])]
```
mphoward, line 631: This might need to be `i+2` to account for the shift by 1 for the index, plus another 1 for the slice.
```suggestion
                ["scale {} {}".format(i + 2, s) for i, s in enumerate(scale[1:])]
```

src/relentless/simulate/simulate.py
```diff
@@ -20,6 +20,24 @@ def __call__(self, sim):
...
 
 
+class InitializationOperation(SimulationOperation):
+    """Operation that initializes a :class:`Simulation`.
+
+    An initialization operation must do the following:
+
+    * Set `sim.dimension`, the dimensionality of the simulation.
+    * Set `sim.types`, the list of particle types in the simulation.
+    * Attach the :class:`Potentials` given by `sim.potentials`.
```
mphoward, line 30:
```suggestion
    * Set `sim.masses`, the dictionary of masses for each particle type in the simulation.
    * Attach the :class:`Potentials` given by `sim.potentials`.
```

tests/simulate/test_lammps.py
```diff
@@ -18,12 +19,32 @@
...
-        cls.__name__, params_dict["dim"]
+# parametrize testing fixture
+test_params = []
+if _has_modules:
+    test_params += [(2, None), (3, None)]
+if _lammps_executable is not None:
+    test_params += [(2, _lammps_executable), (3, _lammps_executable)]
+
+
+@unittest.skipIf(
```
mphoward, line 38: Separate this into two skips so we can be more specific about why we skipped?

clpetix (approved):
This all looks good to me!

PR #164, opened 2023-02-10
Add error messages to EnsembleAnalyzer

### LINKED ISSUE #157: Error handle HOOMD EnsembleAverage
Since there is intermixing of the thermo callback and the RDF, you get an error if you try to create the Ensemble before the RDF callback has run at least once to get `N`.

At minimum, we should catch this condition and throw a sensible message.

mphoward:
This PR throws a meaningful error message if the EnsembleAnalyzer fails during the `process` step. For HOOMD, this could occur if a short run was performed to increment the timestep to something that isn't a multiple of the reporting period, then another very short run is done.

Fixes #157


clpetix (approved):
This all looks good to me!

PR #169, opened 2023-02-15
Support HOOMD 3

### LINKED ISSUE #152: Support HOOMD 3
We need to add support for HOOMD 3, as only having 2.9.x is challenging when it is no longer supported.

I think a good way to do this would be to create a `HOOMDOperation` base class and give it abstract `_call_v2()` and `_call_v3()` methods. The `__call__` method can then invoke the right one based on the version of HOOMD, and the contents of the existing `__call__` methods can move to `_call_v2`. This will require deriving operations to either implement both, or explicitly opt out.

mphoward:
Fixes #152

### DIFF
--- a/src/relentless/simulate/hoomd.py
+++ b/src/relentless/simulate/hoomd.py
@@ -31,7 +31,7 @@ else:
 
 if _hoomd_found and _version.major == 3:
     from hoomd.custom import Action
-elif not _hoomd_found or _version.major == 2:
+else:
     # we need to spoof this Action class to use later in v2 callbacks
     class Action:
         pass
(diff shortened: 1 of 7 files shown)


src/relentless/simulate/hoomd.py
```diff
@@ -28,45 +45,102 @@
...
+        # create the potentials, defer attaching until later
+        neighbor_list = hoomd.md.nlist.Tree(buffer=sim.potentials.pair.neighbor_buffer)
+        pair_potential = hoomd.md.pair.Table(nlist=neighbor_list)
+        for i, j in sim.pairs:
+            r = sim.potentials.pair.x
+            u = sim.potentials.pair.energy((i, j))
+            f = sim.potentials.pair.force((i, j))
+            if numpy.any(numpy.isinf(u)) or numpy.any(numpy.isinf(f)):
+                raise ValueError("Pair potential/force is infinite at evaluated r")
+            pair_potential.params[(i, j)] = dict(r_min=r[0], U=u, F=f)
```
clpetix, line 126: For HOOMD 3 this needs to be: 
```
pair_potential.params[(i, j)] = dict(r_min=r[0], U=u[:-1], F=f[:-1])
pair_potential.r_cut[(i, j)] = r[-1]
```
We need to check that this is most consistent with the way HOOMD 2 works.

mphoward: This should be fixed. If everything seems to be working with the change applied, can you approve the PR? We'll merge and deal with the LAMMPS table issue, and trimming the `r_cut` down in a separate PR.

clpetix (approved):
All seems to be working now!

PR #170, opened 2023-02-16
Refactor operation types

mphoward:
To break diamond inheritance patterns and make a clear hierarchy of operations:

- `InitializationOperation`
- `SimulationOperation`
- `AnalysisOperation`

This should be rebased after PR #169 is merged.

PR #171, opened 2023-02-17
Support Brownian dynamics in LAMMPS

### LINKED ISSUE #166: Support Brownian dynamics in LAMMPS
Brownian dynamics was added in a more recent release. We should identify the first stable version that had the bug fix I added to ensure the friction coefficients were set correctly, then support this operation for LAMMPS. We can do a version check to make sure LAMMPS is compatible, and error out at call time if it is not.

mphoward:
This PR adds conditional / experimental support for Brownian dynamics in LAMMPS. There are several checks needed to make sure LAMMPS can run this type of simulation, and the CI does not currently include the `BROWNIAN` package so it is not very well tested. However, we can probably test in production since the changes are not too complicated.

This should be rebased and merged after PR #170.

Fixes #166

### DIFF
--- a/src/relentless/simulate/lammps.py
+++ b/src/relentless/simulate/lammps.py
@@ -557,7 +557,9 @@ class RunBrownianDynamics(_Integrator):
         self.seed = seed
 
     def call_commands(self, sim):
-        if sim["engine"]["version"] < 20220623:
+        if "BROWNIAN" not in sim["engine"]["packages"]:
+            raise NotImplementedError("LAMMPS BROWNIAN package is not installed.")
+        elif sim["engine"]["version"] < 20220623:
             raise NotImplementedError(
                 "LAMMPS versions prior to 23Jun2022 stable release do not"
                 " properly support Brownian dynamics."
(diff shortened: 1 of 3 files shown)


mphoward:
Noting: we should parse the packages that are installed with LAMMPS for `BROWNIAN` to make sure this command can succeed.

mphoward:
This is ready for review.

src/relentless/simulate/lammps.py
```diff
@@ -528,6 +528,95 @@ def _make_T(thermostat):
             return (thermostat.T, thermostat.T)
 
 
+class RunBrownianDynamics(_Integrator):
+    """Perform a Langevin dynamics simulation.
```
clpetix, line 532: Should this be Brownian dynamics?

clpetix (approved):
Besides the small issue in the documentation, this all looks good to me!

src/relentless/simulate/lammps.py
mphoward: Yep!

PR #172, opened 2023-02-17
Refactor interpolators and splines

mphoward:
This PR refactors `Interpolator` and `PairSpline` to give future flexibility in defining the spline potential. For example, it would now be possible to implement B splines in the pair potential rather than Akima splines. This was not previously possible.

It also modifies the way the knot positions are stored so that they are now differences. This would allow these variables to be optimized in future.

One question is if we should rename "mode" to "style" or something like that. We could also separate them into a "mode" (which is value or diff) AND a "style" / "basis" (which is akima, bspline, etc.). That way we would not need to use the "-" that I currently chose. Ex:
```
mode="diff", style="akima"
```
vs.
```
mode="akima-diff"
```

@clpetix what do you think?

### DIFF
--- a/src/relentless/math.py
+++ b/src/relentless/math.py
@@ -9,6 +9,7 @@ Math functions (`relentless.math`)
     :toctree: generated/
 
     Interpolator
+    AkimaSpline
     KeyedArray
 
 """
@@ -22,7 +23,7 @@ class Interpolator:
     r"""Interpolating function.
 
     Interpolates through a function :math:`y(x)` on the domain
-    :math:`a \le x \le b` using Akima splines. Outside this domain, ``y`` is
+    :math:`a \le x \le b`. Outside this domain, ``y`` is
     extrapolated as a constant, i.e., :math:`y(x < a) = y(a)` and
     :math:`y(x > b) = y(b)`\.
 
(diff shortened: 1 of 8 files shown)


clpetix:
> This PR refactors `Interpolator` and `PairSpline` to give future flexibility in defining the spline potential. For example, it would now be possible to implement B splines in the pair potential rather than Akima splines. This was not previously possible.
> 
> One question is if we should rename "mode" to "style" or something like that. We could also separate them into a "mode" (which is value or diff) AND a "style" / "basis" (which is akima, bspline, etc.). That way we would not need to use the "-" that I currently chose. Ex:
> 
> ```
> mode="diff", style="akima"
> ```
> 
> vs.
> 
> ```
> mode="akima-diff"
> ```
> 
> @clpetix what do you think?

I like the mode & style being separate arguments; that method seems more logical to me!

mphoward:
OK! Then what we will do for now is revert to `mode` for `value` vs. `diff`, and we will internally store the `style="akima"` in a variable. If we decide we want to support B splines, we can then come back and expose the `style` parameter publicly.

mphoward:
This should be ready now. I reverted the changes to `mode` so that it is just `value` or `diff`. I decided it didn't make sense to introduce the `style` variable now if we are not going to do anything with it, but I think this is a clean way to add the option later if we decide we want to support other types of splines.

clpetix (approved):
This all looks good to me!

PR #175, opened 2023-02-23
Pack in order of descending diameters

clpetix:


src/relentless/simulate/initialize.py
```diff
@@ -123,11 +124,16 @@ def _pack_particles(seed, N, V, diameters):
         types = []
         trees = {}
         Nadded = 0
+        evaluated_diameters = copy.deepcopy(diameters)
```
clpetix, line 127: If I did not deepcopy, then when I evaluated the diameters they were being permanently evaluated. So when I returned the diameter variable, it was not longer an object. I'm not sure if there is a better way to handle this.

src/relentless/simulate/initialize.py
```diff
@@ -165,7 +171,7 @@ def _pack_particles(seed, N, V, diameters):
             if len(trees) > 0:
                 mask = numpy.ones(sites.shape[0], dtype=bool)
                 for j, treej in trees.items():
-                    dj = diameters[j]
+                    dj = variable.evaluate(diameters[j])
```
clpetix, line 174: This wasn't caught in the unit tests but failed when testing my own script with one diameter value being a design variable.

src/relentless/simulate/initialize.py
mphoward: Since `diameters` is a `dict`, it is a reference and will be modified if you update its contents in some way. See below for a compact way I think can achieve what we want without `copy`.

src/relentless/simulate/initialize.py
mphoward: Thank you! Yes, that needs to be evaluated here. Can you modify / add to the unit tests so that this case is covered?

src/relentless/simulate/initialize.py
```diff
@@ -123,11 +124,16 @@ def _pack_particles(seed, N, V, diameters):
...
         Nadded = 0
+        evaluated_diameters = copy.deepcopy(diameters)
+        for keys in evaluated_diameters:
+            evaluated_diameters[keys] = variable.evaluate(evaluated_diameters[keys])
         # insert the particles, big to small
-        sorted_N = sorted(N.items(), key=lambda x: x[1], reverse=True)
-        for i, Ni in sorted_N:
+        sorted_diameters = sorted(
+            evaluated_diameters.items(), key=lambda x: x[1], reverse=True
+        )
```
mphoward, line 133: I think you can use a generator comprehension in `sorted`, as `N.items()` should be a generator of tuples:
```suggestion
        sorted_diameters = sorted(
            (i, variable.evaluate(diameters[i])) for i in N,
            key=lambda x: x[1],
            reverse=True
        )
```
Then you would not need the stuff above.

mphoward (changes requested):
Thank you! I have a suggestion for how to avoid the use of `deepcopy`. Let me know if it doesn't work, it definitely should be possible.

mphoward:
Also, while we were looking at this, I was wondering if this line is too conservative:
https://github.com/mphowardlab/relentless/blob/86531555f9daf5015f234f6e379620b9c7cc6b78/src/relentless/simulate/initialize.py#L149
Should it be `0.5*di`? What you want to do is subtract the distance a sphere sticks out of the FCC unit cell. Since we shift be `0.5*di` below, I am thinking that is what this amount is. I can't remember if I had issues getting overlaps if I used that amount here, though.

src/relentless/simulate/initialize.py
clpetix: Since it now packs by diameters and those diameters stay objects, the test fails since the tree will include type B (the design variable). I don't think the tests need modified now.

mphoward (approved):
Looks good to me!

PR #178, opened 2023-03-01
Fix tabulated pair potentials

### LINKED ISSUE #173: LAMMPS simulations failing when rmax is different than potential's cutoff
LAMMPS simulations are blowing up when the potential's stop value is larger than the rmax value. 

We should figure out what the largest rcut is for any pair, then use this as the cutoff for everyone & to reduce the number of points in the interpolation table

mphoward:
This PR should fix the way that potentials get their cutoff trimmed down. Before, the trimming was causing the tabulated potentials to be interpolated in LAMMPS when the cutoff didn't match the end of the table.

Here, we add a method to compute the pairwise energies and forces. Then, all of them are trimmed simultaneously to the biggest required cutoff.

Fixes #173

### DIFF
--- a/src/relentless/simulate/hoomd.py
+++ b/src/relentless/simulate/hoomd.py
@@ -63,21 +63,18 @@ class InitializationOperation(simulate.InitializationOperation):
         # create the potentials, defer attaching until later
         neighbor_list = hoomd.md.nlist.Tree(buffer=sim.potentials.pair.neighbor_buffer)
         pair_potential = hoomd.md.pair.Table(nlist=neighbor_list)
+        r, u, f = sim.potentials.pair.pairwise_energy_and_force(
+            sim.types, tight=True, minimum_num=2
+        )
         for i, j in sim.pairs:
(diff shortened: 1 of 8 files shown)


clpetix:
This fixes the errors I was having in issue #173!

src/relentless/simulate/hoomd.py
```diff
@@ -1006,8 +1005,10 @@ def process(self, sim, sim_op):
 
     def _get_rdf_params(self, sim):
         rdf_params = PairMatrix(sim.types)
-        rmax = sim.potentials.pair.x[-1]
+        rmax = sim[sim.initializer]["_potentials_rmax"]
```
mphoward, line 1008: I think this change should be reverted, will think about it.

mphoward: I think this is OK for now.

mphoward:
@clpetix this is ready for a final review

clpetix (approved):
This all looks good to me and has fixed the problem in my testing!

PR #181, opened 2023-03-03
Improve random initialization

### LINKED ISSUE #177: Random initialization fails to reach higher packing fractions
The random initialization fails to reach higher packer fraction that should still be achievable. We may be able to rely on eliminating sites through the periodic boundaries or dilate the sites so that the lattice fills the box even when it is not a multiple of the lattice.

mphoward:
This PR increases the volume available for packing particles when the boxes are not skew or have only one type.

When there is one type, the procedure for choosing the FCC lattice should guarantee there are no overlaps.

When there is more than one type, the periodic boundary conditions of SciPy's KDTree can be used to exclude overlaps through the boundaries in orthorhombic boxes. When the box is skew, this is not possible, so we have to make sure all the particles are inside the box, as before.

It is still possible that there are certain packing fractions that are difficult to reach using this procedure, and this PR does not address them. We could consider adding support for packmol in future as an alternative that could work in those cases.

Fixes #177

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,9 +110,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-      - name: Install conda JPEG workaround
-        run: |
-          sudo apt-get install -y libjpeg9
       - name: Setup Python ${{ matrix.python-version }}
         uses: conda-incubator/setup-miniconda@v2
         with:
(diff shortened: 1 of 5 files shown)


clpetix (approved):
This all looks good to me! I think considering packmol support later down the line would be a good idea, but for now this fix makes a pretty substantial change from my limited testing!

mphoward:
Thanks! Can you open an issue to propose using packmol for initialization?

PR #189, opened 2023-03-13
Refactor LAMMPS command methods as private

### LINKED ISSUE #187: Make LAMMPS `*_commands` methods private
We don't want someone calling them directly, so make them private methods.

mphoward:
Fixes #187


mphoward:
This is a straightforward find-replace, merging without requesting review.

PR #191, opened 2023-03-14
Dilute NPT simulations

clpetix:


src/relentless/simulate/dilute.py
```diff
@@ -200,12 +198,31 @@ def process(self, sim, sim_op):
...
+                B += y_a * y_b * numpy.trapz(b_ab, x=r)
+        # calculate pressure if no barostat
+        if ens.P == 0.0:
+            ens.P = kT * (rho + B * rho**2)
+            sim[self]["ensemble"] = ens
+            sim[self]["num_thermo_samples"] = None
+            sim[self]["num_rdf_samples"] = None
+        # adjust extent to maintain pressure if barostat
+        else:
+            coeffs = numpy.array([-ens.P / kT, N, B * N**2])
```
mphoward, line 214: Are these the right coefficients? I don't have pencil & paper in front of me, but I think the virial expansion is:
```
P/kT = rho + B rho^2
```
Then solve as:
```
-P/kT + rho + B rho^2 = 0
```
You could also rearrange for the molar volume `v = 1/rho`:
```
B + v + (-P/kT) v^2 = 0
```
Then you would compute `V = N/rho` or `V = N*v`.

Right now, I think you are actually getting `1/V` and not `V`.

tests/simulate/test_dilute.py
```diff
@@ -45,7 +45,7 @@ def test_run(self):
         d = relentless.simulate.Dilute(init, operations=md)
         sim = d.run(potentials=pots, directory=self.directory)
         ens_ = sim[analyzer]["ensemble"]
-        self.assertAlmostEqual(ens_.P, -207.5228556)
+        self.assertAlmostEqual(ens_.P, 194.8990634)
```
mphoward, line 48: I think these might be due to the errors above. The numbers might be a little different but shouldn't have changed sign completely, I don't think... Let me know if that's not the case.

Also, we should add a test for MD with a barostat.

mphoward (changes requested):
Thanks for working on this! I think there may be an issue with the formula that solves for the volume, so can you take a look at that? I also had some suggestions for how to make the loop that computes the virial coefficient a little more compact than what was there before.

src/relentless/simulate/dilute.py
clpetix: My understanding is:
`rho = N/V`
&
`-P/kT + rho + B rho^2 = 0`



Which gives: 
`-P/kT + N/V + B N^2 /V^2 = 0` 

Multiplying `V^2` gives:
`(-P/kT)*V^2 + N*V + BN^2 =0`

The length value output gives the right pressure when running the simulation with no barostat.

mphoward: Oh! I see, the problem is that I misread the order of coefficients for roots. I thought they would go in ascending order (a0 + a1 c + a2 x^2 + …), but they are descending.

For numerics, you may still want to solve for molar volume or density, so that the variable you are solving for is not extensive.

tests/simulate/test_dilute.py
clpetix: I had missed a negative sign in front of b_ij!

src/relentless/simulate/dilute.py
```diff
@@ -171,43 +171,54 @@ def process(self, sim, sim_op):
...
+                ens.V = extent.Cube(L)
+            elif sim.dimension == 2:
+                ens.V = extent.Square(L)
+
+            sim[self]["ensemble"] = ens
+            sim[self]["num_thermo_samples"] = None
+            sim[self]["num_rdf_samples"] = None
 
 
 class Dilute(simulate.Simulation):
```
mphoward, line 224: We should update the documentation of this class while we are here:

- [ ] After the equation of g(r), add a paragraph about how the virial equation of state is used. Give the expression for the pressure and the second virial coefficient.
- [ ] Remove the last paragraph. It is out of data, and an implementation detail that is not important anyway.

src/relentless/simulate/dilute.py
```diff
@@ -171,43 +171,54 @@ def process(self, sim, sim_op):
...
+            y_j = ens.N[j] / N
+            b_ij = -counting_factor * (geo_prefactor / 2.0) * (numpy.exp(-u / kT) - 1)
+            B += y_i * y_j * numpy.trapz(b_ij, x=r)
+        # calculate pressure if no barostat
+        if ens.P is None:
+            rho = N / ens.V.extent
+            ens.P = kT * (rho + B * rho**2)
+            sim[self]["ensemble"] = ens
+            sim[self]["num_thermo_samples"] = None
+            sim[self]["num_rdf_samples"] = None
```
mphoward, line 206: This is duplicated code, de-indent below.
```suggestion
```

src/relentless/simulate/dilute.py
```diff
@@ -171,43 +171,54 @@ def process(self, sim, sim_op):
...
+            L = V ** (1 / sim.dimension)
+
+            if sim.dimension == 3:
+                ens.V = extent.Cube(L)
+            elif sim.dimension == 2:
+                ens.V = extent.Square(L)
+
+            sim[self]["ensemble"] = ens
+            sim[self]["num_thermo_samples"] = None
+            sim[self]["num_rdf_samples"] = None
```
mphoward, line 221:
```suggestion
        sim[self]["ensemble"] = ens
        sim[self]["num_thermo_samples"] = None
        sim[self]["num_rdf_samples"] = None
```

src/relentless/simulate/dilute.py
```diff
@@ -171,43 +171,54 @@ def process(self, sim, sim_op):
...
+            if sim.dimension == 3:
+                geo_prefactor = 4 * numpy.pi * r**2
+            elif sim.dimension == 2:
+                geo_prefactor = 2 * numpy.pi * r
+            else:
+                raise ValueError("Geometric integration factor unknown for extent type")
+            y_i = ens.N[i] / N
+            y_j = ens.N[j] / N
+            b_ij = -counting_factor * (geo_prefactor / 2.0) * (numpy.exp(-u / kT) - 1)
+            B += y_i * y_j * numpy.trapz(b_ij, x=r)
```
mphoward, line 199: The meaning of `counting_factor` is clearer if defined and used in the summation of `B`, while putting the integral to get a B_ij also groups like concepts together:
```suggestion
            B_ij = - (geo_prefactor / 2.0) * numpy.trapz(numpy.exp(-u / kT) - 1, x=r)

            counting_factor = 2 if i != j else 1
            B += counting_factor * y_i * y_j * B_ij
```
I am using `B_ij` (capitalized) on purpose because in stat mech there is another integral that is `b_ij` that is almost the same thing but not quite...

src/relentless/simulate/dilute.py
```diff
@@ -171,43 +171,54 @@ def process(self, sim, sim_op):
...
-                y = (geo_prefactor / 6.0) * rho_a * rho_b * f * gr * r
-                ens.P += numpy.trapz(y, x=r)
-
-        sim[self]["ensemble"] = ens
-        sim[self]["num_thermo_samples"] = None
-        sim[self]["num_rdf_samples"] = None
+        B = 0.0
+        N = sum(ens.N[i] for i in sim.types)
+        for i, j in sim.pairs:
+            counting_factor = 2 if i != j else 1
```
mphoward, line 178: Maybe also move this below, and put it before the `B +=`, with a blank line before `counting_factor =`? It is not needed yet (I just wrote it below the for loop for compactness =)

src/relentless/simulate/dilute.py
```diff
@@ -171,43 +171,54 @@ def process(self, sim, sim_op):
...
+        if ens.P is None:
+            rho = N / ens.V.extent
+            ens.P = kT * (rho + B * rho**2)
+            sim[self]["ensemble"] = ens
+            sim[self]["num_thermo_samples"] = None
+            sim[self]["num_rdf_samples"] = None
+        # adjust extent to maintain pressure if barostat
+        else:
+            coeffs = numpy.array([-ens.P / kT, 1, B])
+            v = numpy.max(numpy.roots(coeffs))
```
mphoward, line 210:
```suggestion
            v = numpy.max(numpy.roots([-ens.P / kT, 1, B]))
```

src/relentless/simulate/dilute.py
```diff
@@ -171,43 +171,54 @@ def process(self, sim, sim_op):
...
-                        "Geometric integration factor unknown for extent type"
-                    )
-                y = (geo_prefactor / 6.0) * rho_a * rho_b * f * gr * r
-                ens.P += numpy.trapz(y, x=r)
-
-        sim[self]["ensemble"] = ens
-        sim[self]["num_thermo_samples"] = None
-        sim[self]["num_rdf_samples"] = None
+        B = 0.0
+        N = sum(ens.N[i] for i in sim.types)
```
mphoward, line 176: I like to put summation variables right before the loop with separation, but that's just a matter of style preference
```suggestion
        N = sum(ens.N[i] for i in sim.types)

        B = 0.0
```

mphoward (changes requested):
This looks great! I only have a few small things and a request to work on the docs a little, then this is good to merge.

clpetix:
This should be ready for a final review!

mphoward (approved):
Looks great, thank you!

Issue #156, opened 2023-01-27
Make configuration file I/O interoperable


mphoward:
The operations that read/write configurations/trajectories should be interoperable (e.g., HOOMD should be able to seamlessly initialize from a LAMMPS data file). To do this, we should try to leverage as many existing Python tools as possible for manipulating trajectories.

PR #194, opened 2023-03-21
Refactor EnsembleAverage interface and implementations

### LINKED ISSUE #163: Optimize `EnsembleAverage`
We should inspect the `sim_op` and data from initialization to store static properties (e.g., N, V) so we don't waste time computing them when we don't need to.

### LINKED ISSUE #188: Refactor `EnsembleAverage` arguments
We should package related options together like:
```py
EnsembleAverage(
    properties={"every": 1000, "quantities": ("temperature",)},
    rdf={"every": 5000, "spacing": 0.1},
)
```
rather than
```py
EnsembleAverage(check_thermo_every=1000, check_rdf_every=5000, rdf_dr=0.1)
```

We can also then _optionally_ do certain actions. Like, if `rdf` is not specified, this calculation can be skipped.

This should be done in conjunction with #163

mphoward:
This PR updates the `EnsembleAverage` operation in a few ways:

* Improve the user interface for specifying parameters (Fixes #188)
* Allow users to specify RDF cutoffs greater than the tabulated potential cutoff. This is a feature we would like to have for analysis. To support this, LAMMPS files are postprocessed with freud instead of computed on the fly. HOOMD could do something similar if performance becomes a bottleneck, but for now, it is done on the fly.
* Refactor the HOOMD compute operation into one callback. This is moving toward optimizing the operations.
* `lammpsio` and `freud-analysis` are required dependencies of `relentless` again. I used `freud` in both `simulate.hoomd` and `simulate.lammps`, so I didn't think it was worth making the user figure out they needed to install this. I am envisioning using `lammpsio` between these packages in future (#156), so I moved that as well now. Now users just need to know "install HOOMD" or "install LAMMPS", which is nice.
* Optimize the averaging to assume certain things are constant (Fixes #163) via the `assume_constraints` option.

### DIFF
--- a/src/relentless/simulate/dilute.py
+++ b/src/relentless/simulate/dilute.py
@@ -111,14 +111,16 @@ class RunMolecularDynamics(_Integrator):
         self.barostat = barostat
 
     def __call__(self, sim):
-        if self.barostat is not None:
-            sim[sim.initializer]["_ensemble"].P = self.barostat.P
-
         for analyzer in self.analyzers:
             analyzer.pre_run(sim, self)
 
         if self.thermostat is not None:
             sim[sim.initializer]["_ensemble"].T = self.thermostat.T
+
+        if self.barostat is not None:
+            sim[sim.initializer]["_ensemble"].P = self.barostat.P
+        else:
+            sim[sim.initializer]["_ensemble"].P = None
 
         for analyzer in self.analyzers:
             analyzer.post_run(sim, self)
@@ -168,17 +170,8 @@ class EnsembleAverage(simulate.AnalysisOperation):
         ens = sim[sim.initializer]["_ensemble"].copy()
         kT = sim.potentials.kB * ens.T
 
-        # pair distribution function is needed to get the pressure
-        # we use the potential spacing for this integral, then deal with outputting
-        # it later
-        for pair in ens.pairs:
-            u = sim.potentials.pair.energy(pair)
-            ens.rdf[pair] = ensemble.RDF(
-                sim.potentials.pair.linear_space, numpy.exp(-u / kT)
-            )
         # compute pressure
         N = sum(ens.N[i] for i in sim.types)
-
         B = 0.0
         for i, j in sim.pairs:
             r = sim.potentials.pair.linear_space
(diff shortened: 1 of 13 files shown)


mphoward:
@clpetix Could you take a look at this draft PR and see if the new user interface and modifications make sense? If yes, I can add the unit tests for `assume_constraints`, and this will be ready to merge.

clpetix:
@mphoward I like this change! This is a clean way to handle the new rdf capability!

mphoward:
Cool! This should be all ready for review, so go through it when you get a chance and we can merge.

src/relentless/simulate/dilute.py
```diff
@@ -208,7 +213,12 @@ def process(self, sim, sim_op):
         # adjust extent to maintain pressure if barostat
         else:
             coeffs = numpy.array([-ens.P / kT, 1, B])
-            v = numpy.max(numpy.roots(coeffs))
+            roots = numpy.roots(coeffs)
+            if not numpy.allclose(numpy.imag(roots), 0):
+                raise ValueError(
+                    "Unable to solve for volume, imaginary roots obtained."
+                )
+            v = numpy.max(numpy.real(roots))
```
mphoward, line 221: @clpetix I was getting a warning in my unit tests related to casting imaginary values to real, so I added this check here. I think the choice of potential parameters + pressure in the NPT dilute test may not be solvable. Could you please investigate and, if that's all that is the issue, let me know what an appropriate choice would be?

src/relentless/simulate/dilute.py
```diff
@@ -165,7 +167,10 @@ def post_run(self, sim, sim_op):
     def process(self, sim, sim_op):
         ens = sim[sim.initializer]["_ensemble"].copy()
         kT = sim.potentials.kB * ens.T
-        # pair distribution function
+
+        # pair distribution function is needed to get the pressure
```
mphoward, line 171: I think this is no longer true and so this code can be removed here.

src/relentless/simulate/dilute.py
mphoward: @clpetix I modified this test so that now it uses NVT to get the pressure for known V, then does NPT to get V back for the same P. If you have a spreadsheet somewhere, could you compute what the P that comes from the NVT simulation is supposed to be? I think it was `1.25`

clpetix: @mphoward the answer is 1.25 for this step size and cut-off of r!

src/relentless/simulate/dilute.py
```diff
@@ -163,17 +141,52 @@ def post_run(self, sim, sim_op):
...
+            else:
+                T = thermostat.T
+            N = sim[sim.initializer]["_ensemble"].N
+            V = sim[sim.initializer]["_ensemble"].V
+
+            ens = ensemble.Ensemble(T, N, V)
+        elif isinstance(sim_op, (md.RunMolecularDynamics, RunMolecularDynamics)):
+            if sim_op.thermostat is not None:
+                thermostat = sim_op.thermostat
+                if thermostat.anneal:
```
clpetix, line 166: Is temperature annealing an operation needed for dilute situations?

clpetix (approved):
Besides my question about the need for dilute annealing and adding the pressure assertion in dilute testing-this all looks good to me!

PR #197, opened 2023-04-04
Support I/O with both GSD and LAMMPS files

### LINKED ISSUE #156: Make configuration file I/O interoperable

mphoward:
This PR uses a feature I recently added to `lammpsio` to convert between GSD and LAMMPS snapshots. `InitializeFromFile` and `WriteTrajectory` should now be mostly agnostic to whether the box is HOOMD or LAMMPS.

In the process, I refactored a method for getting a temporary file name to the `Directory` to reduce code duplication.

Fixes #156

### DIFF
--- a/src/relentless/simulate/hoomd.py
+++ b/src/relentless/simulate/hoomd.py
@@ -1448,14 +1448,14 @@ class WriteTrajectory(AnalysisOperation):
         if file_format == "LAMMPS-dump":
             if mpi.world.rank_is_root:
                 dump_file = sim.directory.temporary_file(".dump")
-                with gsd.hoomd.open(self.filename) as t:
-                    snaps = [lammpsio.Snapshot.from_hoomd_gsd(s) for s in t]
+                with gsd.hoomd.open(sim.directory.file(self.filename)) as t:
+                    snaps = [lammpsio.Snapshot.from_hoomd_gsd(s)[0] for s in t]
 
-                schema = analyze.WriteTrajectory._make_dump_schema(
+                schema = analyze.WriteTrajectory._make_lammps_schema(
                     self.velocities, self.images, self.types, self.masses
                 )
                 lammpsio.DumpFile.create(dump_file, schema, snaps)
-                shutil.move(dump_file, self.filename)
+                shutil.move(dump_file, sim.directory.file(self.filename))
             mpi.world.barrier()
 
     def _get_dynamic(self):
--- a/src/relentless/simulate/hoomd.py
+++ b/src/relentless/simulate/hoomd.py
@@ -1,7 +1,8 @@
 import enum
-import os
 import warnings
 
+import gsd.hoomd
+import lammpsio
 import numpy
 from packaging import version
 
@@ -58,6 +59,7 @@ class InitializationOperation(simulate.InitializationOperation):
         # parse masses by type
         snap = sim["engine"]["_hoomd"].state.get_snapshot()
         sim.masses = self._get_masses_from_snapshot(sim, snap)
+        self._assert_dimension_safe(sim, snap)
 
         # create the potentials, defer attaching until later
         neighbor_list = hoomd.md.nlist.Tree(buffer=sim.potentials.pair.neighbor_buffer)
@@ -85,6 +87,7 @@ class InitializationOperation(simulate.InitializationOperation):
         with sim["engine"]["_hoomd"]:
             snap = sim[self]["_system"].take_snapshot(particles=True)
             sim.masses = self._get_masses_from_snapshot(sim, snap)
+            self._assert_dimension_safe(sim, snap)
 
         # attach the potentials
         def _table_eval(r_i, rmin, rmax, **coeff):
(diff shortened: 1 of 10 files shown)


clpetix:
I added a test to check that initialization works with both file types for both engines. Let me know if there are any revisions or additions I need to make!

mphoward:
Thank you! Could you also add tests that `WriteTrajectory` at least runs for both file type outputs? It would be even better if you could verify the contents, but it's OK if not. I will review all the tests at one time.

clpetix:
@mphoward this is ready for review. I fixed a few things in the WriteTrajectory conversions and added testing

clpetix:
> Thank you! Could you merge in `main` and resolve the conflicts before review?

Done!

src/relentless/simulate/hoomd.py
```diff
@@ -145,26 +149,62 @@ def _get_masses_from_snapshot(self, sim, snap):
...
 
     def _initialize_v2(self, sim):
+        gsd_filename = self._convert_to_gsd_file(sim)
         with sim["engine"]["_hoomd"]:
-            return hoomd.init.read_gsd(self.filename)
+            return hoomd.init.read_gsd(gsd_filename)
+
+    def _convert_to_gsd_file(self, sim):
+        file_format = initialize.InitializeFromFile._detect_format(
+            self.filename, self.format
```
mphoward, line 187: Does this have the same issue with not wrapping this `sim.directory.file(self.filename)`? Maybe we should add a line at the top:
```py
filename = sim.directory.file(self.filename)
```
Then use that rather than `self.filename` everywhere?

src/relentless/simulate/hoomd.py
```diff
@@ -1418,7 +1459,19 @@ def _post_run_v2(self, sim, sim_op):
         del sim[self]["_gsd"]
 
     def process(self, sim, sim_op):
-        pass
+        file_format = analyze.WriteTrajectory._detect_format(self.filename, self.format)
```
mphoward, line 1462: Same here, maybe we should do the fix above to make it clearer how the `filename` is being used consistently throughout?

src/relentless/simulate/lammps.py
```diff
@@ -261,29 +274,57 @@ def initialize_commands(self, sim):
...
+    def _convert_to_data_file(self, sim):
+        file_format = initialize.InitializeFromFile._detect_format(
+            self.filename, self.format
+        )
+        if file_format == "HOOMD-GSD":
+            if mpi.world.rank_is_root:
+                data_filename = sim.directory.temporary_file(".data")
+                with gsd.hoomd.open(self.filename) as f:
+                    frame = f[0]
+                snap = lammpsio.Snapshot.from_hoomd_gsd(frame)[0]
```
mphoward, line 299: Do we need to use the other return value to build the LAMMPS type map? I forgot about this.

src/relentless/simulate/lammps.py
```diff
@@ -1313,7 +1363,20 @@ def _post_run_commands(self, sim, sim_op):
         return cmds
 
     def process(self, sim, sim_op):
-        pass
+        file_format = analyze.WriteTrajectory._detect_format(
```
mphoward, line 1366: Same comment about `filename` here.

src/relentless/simulate/lammps.py
```diff
@@ -1313,7 +1363,20 @@ def _post_run_commands(self, sim, sim_op):
...
+            sim.directory.file(self.filename), self.format
+        )
+        if file_format == "HOOMD-GSD":
+            if mpi.world.rank_is_root:
+                gsd_file = sim.directory.temporary_file(".gsd")
+                with gsd.hoomd.open(gsd_file, "wb") as t:
+                    for snap in lammpsio.DumpFile(
+                        sim.directory.file(self.filename), sort_ids=True
+                    ):
+                        frame = snap.to_hoomd_gsd()
```
mphoward, line 1376: Do we need to override the types in the GSD frame, if there is a typemap set for LAMMPS? I think so and I again I forgot that.

mphoward (commented):
I have some comments on the code. Would you want to take a pass at addressing them, or would you like me to? I will review the tests in a second pass.

src/relentless/simulate/hoomd.py
clpetix: I don't think these are an issue. The `_detect_format()` function is just checking the filename as a string. The location of the file doesn't impact its ability to do that check

mphoward: What about down below where it reads the `DataFile` though? (That was actually what I was looking at.) I thought it might just be more readable in general to use the _same_ filename throughout the function.

clpetix: I'm not sure why it didn't cause trouble with the `DataFile` below. I agree that it would be better to use the same filename throughout!

tests/simulate/test_lammps.py
```diff
@@ -578,6 +616,39 @@ def test_writetrajectory(self):
             self.assertCountEqual(snap.typeid, [1, 1, 2, 2, 2])
             self.assertCountEqual(snap.mass, [1, 1, 1, 1, 1])
 
+    def test_writetrajectory_gsd(self):
```
mphoward, line 619: Can we simplify testing this operation by attaching both analyzers to the same simulation? Like, combine this test with what's above?

tests/simulate/test_lammps.py
```diff
@@ -578,6 +616,39 @@ def test_writetrajectory(self):
...
+        )
+        h = relentless.simulate.LAMMPS(init, operations=lgv, executable=self.executable)
+        sim = h.run(potentials=pot, directory=self.directory)
+
+        # read hoomd trajectory file
+        file = sim.directory.file("test_writetrajectory.gsd")
+        with gsd.hoomd.open(file) as traj:
+            for snap in traj:
+                self.assertEqual(snap.particles.N, 5)
+                self.assertIsNotNone(snap.particles.velocity)
```
mphoward, line 649: Is there a reason you dropped some of the other things we tested in the other test? Did it not work right?

tests/simulate/test_lammps.py
```diff
@@ -110,34 +140,59 @@ def create_file(self):
...
+        else:
+            raise ValueError("LAMMPS supports 2d and 3d simulations")
+        ens.P = 2.5
+        pot = LinPot(ens.types, params=("m",))
+        for pair in pot.coeff:
+            pot.coeff[pair].update({"m": -2.0, "rmax": 1.0})
+        pots = relentless.simulate.Potentials()
+        pots.pair = relentless.simulate.PairPotentialTabulator(
+            pot, start=1e-6, stop=2.0, num=10, neighbor_buffer=0.1
+        )
```
mphoward, line 180: I'm not quite following why this is here?

tests/simulate/test_hoomd.py
```diff
@@ -506,6 +549,36 @@ def test_writetrajectory(self):
                 self.assertIsNotNone(snap.particles.velocity)
                 self.assertCountEqual(snap.particles.typeid, [0, 0, 1, 1, 1])
 
+    def test_writetrajectory_lammpstrj(self):
```
mphoward, line 552: Same comments apply to this as in the LAMMPS file.

tests/simulate/test_hoomd.py
```diff
@@ -96,13 +97,55 @@ def create_gsd(self):
...
         ens, pot = self.ens_pot()
-        f = self.create_gsd()
+        f = self.create_gsd_file()
         op = relentless.simulate.InitializeFromFile(filename=f)
         h = relentless.simulate.HOOMD(op)
         h.run(pot, self.directory)
 
+    def test_initialize_from_lammps_file(self):
+        """Test running initialization simulation operations."""
+        if self.dim == 3:
```
mphoward, line 126: Same question here about why this is part of the test?

mphoward (commented):
I think I addressed the issues with the filenames and types! I had some questions about how the tests were implemented. If you can explain, then either you and I can work on them.

tests/simulate/test_lammps.py
clpetix: I had that there because of the type mismatch (HOOMD being A&B and LAMMPS being 1&2). Using with the `ens_pot()` was throwing an error

tests/simulate/test_lammps.py
clpetix: I had it test the same things that are tested when writing a gsd from HOOMD.

tests/simulate/test_lammps.py
clpetix: We could! I had them separate so if they fail they show the descriptive name, but you could get that info from the traceback

tests/simulate/test_lammps.py
mphoward: Oh, I see! At minimum, we could add an argument to `ens_pot` that specifies whether the types are A & B or 1 & 2. Alternatively, we could revise the LAMMPS tests to use A & B (probably it is an artifact from older code that didn't support non-numeric types in LAMMPS), but that might be a little more tedious.

tests/simulate/test_lammps.py
mphoward: Oh OK, I wonder why we didn't test for that in the HOOMD GSD?

tests/simulate/test_lammps.py
mphoward: Yeah, I think the traceback is sufficient. The unittest framework will tell us which assertion is failing, so I would rather follow the traceback than have to maintain significant duplicate code.

tests/simulate/test_hoomd.py
mphoward: Here, we do have to make the potentials because we don't currently support remapping from LAMMPS types to relentless types on initialization. We could reconsider this later.

clpetix (approved):
This all looks good to me! Thanks for cleaning up those tests; I appreciate it!

PR #201, opened 2023-04-12
Improve packaging and installation

mphoward:
This PR includes mechanisms to start publishing releaeses to PyPI. Once we have a package available on PyPI, we can make it available on conda-forge as well.

This PR also updates some of the documentation related to installation, in anticipation of this.


mphoward:
@clpetix if this all looks good to you, let's merge so that we can move onto tagging a release as discussed yesterday.

clpetix (approved):
This all looks good to me!

PR #207, opened 2023-04-14
Improve default values of collections and errors

### LINKED ISSUE #206: Give `PairMatrix` a default value
Right now it always makes an empty `dict`, but this may not be helpful. We should let the user decide.

We should also allow `default` to be a callable here and in `FixedKeyDict` so that a dict can safely be the default. Alternatively, we need to use `deepcopy`.

mphoward:
* Give PairMatrix a default
* Make all defaults copy safe
* Simplify access patterns using iterators
* Error check RDF parameters are set for relative entropy

Fixes #206


clpetix (approved):
This all looks good to me!

PR #208, opened 2023-04-14
Explicitly cast types during JSON serialization

mphoward:
This fixes a potential issue where some NumPy return types (e.g., `int64`) may not be JSON serializable. Now, we explicitly cast variable types before we write to file.

This also uses default types (`float` and `int`) instead of specific precisions in all NumPy arrays.


clpetix (approved):
This all looks good to me!

PR #212, opened 2023-04-18
Add filename argument for analyzers

### LINKED ISSUE #205: Add an optional filename to Record operation
We should add an optional filename to the Record operation that saves a file of the recorded data to the simulation directory.

mphoward:
This PR adds an optional filename argument to both `Record` and `EnsembleAverage` so that the user can control output to file when the simulation is run in an optimization, and they do not have access to the Python outputs at each iteration.

Fixes #205


clpetix (approved):
This looks good to me!

PR #223, opened 2023-09-29
Use absolute path for `InitializeFromFile`

clpetix:


mphoward:
The RTD build is failing because they changed a required entry in their configuration file. This will be fixed with PR #224, so please merge up `main` after that gets finalized.

src/relentless/simulate/initialize.py
```diff
@@ -44,6 +45,14 @@ def __init__(self, filename, format=None, dimension=None):
         self.format = format
         self.dimension = dimension
 
+    @property
+    def filename(self):
```
mphoward, line 49: Document properties:
```suggestion
    def filename(self):
        """str: Absolute path to file to be read."""
```

tests/simulate/test_hoomd.py
```diff
@@ -127,6 +128,16 @@ def test_initialize_from_gsd_file(self):
...
         h.run(pot, self.directory)
 
+        # Run in a different directory
+        with self.directory:
+            d = self.directory.directory(
+                "run", create=relentless.mpi.world.rank_is_root
+            )
+            relentless.mpi.world.barrier()
+            op = relentless.simulate.InitializeFromFile(pathlib.Path(f).name)
+            h = relentless.simulate.HOOMD(op)
```
mphoward, line 138: You might want to do this to test the setter:
```suggestion
            op.filename = pathlib.Path(f).name
```
Alternatively, you don't need a new HOOMD instance, just do:
```suggestion
            h.initializer = relentless.simulate.InitializeFromFile(pathlib.Path(f).name)
```
Maybe you actually want to do both here?

tests/simulate/test_lammps.py
```diff
@@ -164,6 +165,20 @@ def test_initialize_from_lammps_file(self):
         lmp = relentless.simulate.LAMMPS(op, executable=self.executable)
         lmp.run(potentials=pot, directory=self.directory)
 
+        # Run in a different directory
```
mphoward, line 168: Make the same changes down here as you do for `test_hoomd.py`

mphoward (changes requested):
This looks good! Here are a few small things, then this is ready to merge.

PR #232, opened 2024-01-30
Check for compatibility between box dimension and Lz for HOOMD versions

clpetix:


src/relentless/simulate/hoomd.py
```diff
@@ -212,7 +212,30 @@ def _convert_to_gsd_file(self, sim):
                 gsd_filename = None
             gsd_filename = mpi.world.bcast(gsd_filename)
         else:
-            gsd_filename = self.filename
+            if mpi.world.rank_is_root:
+                gsd_filename = self.filename
+                with gsd.hoomd.open(self.filename) as snap:
+                    frame = snap[0]
```
mphoward, line 218: I would use `trajectory` or `t` or `f` here, based on usually conventions for opening these files:
```suggestion
                with gsd.hoomd.open(self.filename) as trajectory:
                    frame = trajectory[0]
```

src/relentless/simulate/hoomd.py
```diff
@@ -212,7 +212,30 @@ def _convert_to_gsd_file(self, sim):
...
+                                with gsd.hoomd.open(gsd_filename, _gsd_write_mode) as f:
+                                    f.append(frame)
+                        else:
+                            if frame.configuration.box[2] > 0.0:
+                                pass
+                            else:
+                                frame.configuration.box[2] = 1
+                                gsd_filename = sim.directory.temporary_file(".gsd")
+                                with gsd.hoomd.open(gsd_filename, _gsd_write_mode) as f:
+                                    f.append(frame)
```
mphoward, line 235: Consolidate these into an `if-elif-else` (no empty if & pass). You can add comments about why these particular conditions are needed:
```
if _hoomd_version.major >= 3 and frame.configuration.box[2] == 0:
    # fix for HOOMD 2 style used in HOOMD 3
elif _hoomd_version.major == 2 and frame.configuration.box[2] > 0:
    # fix for HOOMD 3 style used in HOOMD 2
else:
    gsd_filename = self.filename
```

mphoward (changes requested):
Looks like the right idea to me! Here are two small things

src/relentless/simulate/hoomd.py
```diff
@@ -213,6 +213,33 @@ def _convert_to_gsd_file(self, sim):
             gsd_filename = mpi.world.bcast(gsd_filename)
         else:
             gsd_filename = self.filename
+            if mpi.world.rank_is_root:
```
mphoward, line 216: I was slightly confused here because this variable gets declared but then overridden on everything that's not the root rank. This made more sense to me:
```suggestion
            if mpi.world.rank_is_root:
                gsd_filename = self.filename
```

mphoward (changes requested):
Two small things here! Is it possible to add some sort of unit test for this? I'm not sure how we could tell it was failing before

mphoward:
I agree! Let me know when this is ready for final review.

clpetix:
This version is working and should have all the requested changes! It is ready for you whenever you get a chance @mphoward !

mphoward (approved):
Looks great, thanks!

PR #234, opened 2024-02-07
Add box resizing to MD integrators

### LINKED ISSUE #233: Add box resizing to MD integrators
We should add the ability to resize the box to MD integration methods. We can use the `barostat` argument to do this:
* If `barostat` is an `Extent`, resize the box immediately at the start of the run.
* If `barostat` is a `tuple(Extent, Extent)`, do linear interpolation between the two box sizes from the beginning to the end of the run.

clpetix:


clpetix:
@mphoward this is a working implementation of box resizing for LangevinDynamics. It is throwing a ton of errors on the unit tests, but I'm not sure how polished vs hacky we want this to be for now. 

Let me know how you'd like to proceed!

src/relentless/simulate/md.py
```diff
@@ -118,9 +118,10 @@ class RunLangevinDynamics(_Integrator):
 
     """
 
-    def __init__(self, steps, timestep, T, friction, seed, analyzers=None):
+    def __init__(self, steps, timestep, T, barostat, friction, seed, analyzers=None):
```
mphoward, line 121: Can you add the `barostat=None` right before `analyzers` so it is an optional argument? Then, update and handle this accordingly in the HOOMD backend. That may fix some of your failing unit tests.

src/relentless/simulate/hoomd.py
```diff
@@ -709,6 +719,38 @@ def _call_v2(self, sim):
...
+
+                    Lx = hoomd.variant.linear_interp([(0, Lx1), (self.steps, Lx2)])
+                    Ly = hoomd.variant.linear_interp([(0, Ly1), (self.steps, Ly2)])
+                    Lz = hoomd.variant.linear_interp([(0, Lz1), (self.steps, Lz2)])
+                    xy = hoomd.variant.linear_interp([(0, xy1), (self.steps, xy2)])
+                    xz = hoomd.variant.linear_interp([(0, xz1), (self.steps, xz2)])
+                    yz = hoomd.variant.linear_interp([(0, yz1), (self.steps, yz2)])
+
+            hoomd.update.box_resize(
+                Lx=Lx, Ly=Ly, Lz=Lz, xy=xy, xz=xz, yz=yz, scale_particles=False
```
mphoward, line 752: I would probably use `scale_particles=True` since that is also how a barostat works.

For this one that operates periodically, you also need to save it in a variable and disable & delete it after the run (see below)

src/relentless/simulate/hoomd.py
```diff
@@ -709,6 +719,38 @@ def _call_v2(self, sim):
             for analyzer in self.analyzers:
                 analyzer.pre_run(sim, self)
 
+            if isinstance(self.barostat, extent.Extent):
+                if sim.dimension == 2:
+                    Lx, Ly, xy = self.barostat.as_array("HOOMD")
+                    Lz, xz, yz = 0.0, 0.0, 0.0
```
mphoward, line 725: Unused arguments should be set to `None` so that the current value is not modified at all.

src/relentless/simulate/hoomd.py
```diff
@@ -709,6 +719,38 @@ def _call_v2(self, sim):
             for analyzer in self.analyzers:
                 analyzer.pre_run(sim, self)
 
+            if isinstance(self.barostat, extent.Extent):
```
mphoward, line 722: In this case, the period of the updater should be set so that the action is taken immediately. I forget if that is `None` or something else.

src/relentless/simulate/hoomd.py
```diff
@@ -709,6 +719,38 @@ def _call_v2(self, sim):
...
+
+                    Lz, xz, yz = 0.0, 0.0, 0.0
+                    Lx = hoomd.variant.linear_interp([(0, Lx1), (self.steps, Lx2)])
+                    Ly = hoomd.variant.linear_interp([(0, Ly1), (self.steps, Ly2)])
+                    xy = hoomd.variant.linear_interp([(0, xy1), (self.steps, xy2)])
+                else:
+                    Lx1, Ly1, Lz1, xy1, xz1, yz1 = self.barostat[0].as_array("HOOMD")
+                    Lx2, Ly2, Lz2, xy2, xz2, yz2 = self.barostat[1].as_array("HOOMD")
+
+                    Lx = hoomd.variant.linear_interp([(0, Lx1), (self.steps, Lx2)])
```
mphoward, line 744: Should this be `self.steps-1` to make sure the right box size is achieved at the end of the integration? I think that is how HOOMD's variants work.

src/relentless/simulate/hoomd.py
clpetix: I checked the 2.9.7 documentation and to start immediately then `phase=-1`. Also, `scale_particles` defaults to `True` so I have removed that argument.

mphoward: I think you actually need `period=None` to have it act immediately (i.e., resize right away). Otherwise, the default period is 1, so it would try to update over time. Concurrently, you're also giving it a constant for the sizes, so these are probably essentially the same except that the updater will keep triggering over and over again.

So in summary, I think `period=None` when we just want to change the box size immediately (a single `Extent`), otherwise `period=1` to go between two boxes.

https://hoomd-blue.readthedocs.io/en/v2.9.7/module-hoomd-update.html#hoomd.update.box_resize

clpetix:
Found a small issue where there was no check on whether a barostat was set before applying the barostat logic. This caused an error when `barostat=None`. 

This seems to be running well in the test case discussed. Ready for a review whenever you get a chance next week, @mphoward !

mphoward (changes requested):
Thanks Levi, this looks great! I had some small comments.

In order to fix the issue with the unit tests, can you add the `barostat` argument to `RunLangevinDynamics` for the other simulation methods? If you put it in their constructors, they will be able to take it when they are constructed, and then they can just do nothing with it for the time being. I think that will allow everything to pass.

clpetix:
@mphoward I think this is all good and ready for use!

clpetix:
Making a note that we should ensure the the integrator steps is strictly an int. Else, we'll throw an error when using scientific notation.

PR #236, opened 2024-02-12
Hack in bonded interactions

clpetix:


src/relentless/simulate/hoomd.py
```diff
@@ -1228,11 +1240,27 @@ def act(self, timestep):
...
+                            )
+                            if snap.bonds is not None:
+                                bonds = numpy.vstack(
+                                    [self.bonds, numpy.flip(self.bonds, axis=1)]
+                                )
+                                # list intersect method from:
+                                # https://stackoverflow.com/a/67113105
+                                filter = ~(neighbors[:, None] == bonds).all(-1).any(1)
+                            else:
+                                filter = numpy.full((len(neighbors[:])), True)
```
mphoward, line 1258: I think we can move the filter call up here so it only runs when `bonds` is present. I would also give it a longer name than `filter` so the intent is clear:
```suggestion
                                bonded_neighbor_filter = (neighbors[:, None] == bonds).all(-1).any(1)
                                neighbors.filter(~bonded_neighbor_filter)
```
This will make it easier for us to also add angle & dihedral filters later.

src/relentless/simulate/hoomd.py
```diff
@@ -1228,11 +1240,27 @@ def act(self, timestep):
...
                             # resetting when the samples are zero clears the RDF
                             self._rdf[i, j].compute(
                                 aabbs[j],
                                 snap.particles.position[type_masks[i]],
-                                neighbors=query_args,
+                                neighbors=neighbors.filter(filter),
```
mphoward, line 1263: Then use just `neighbors` here:
```suggestion
                                neighbors=neighbors,
```

src/relentless/simulate/hoomd.py
```diff
@@ -70,8 +70,8 @@ def _call_v3(self, sim):
         # parse masses by type
         snap = sim["engine"]["_hoomd"].state.get_snapshot()
         sim.masses = self._get_masses_from_snapshot(sim, snap)
+        sim[self]["_bonds"] = self._get_bonds_from_snapshot(sim, snap)
         self._assert_dimension_safe(sim, snap)
-
         # create the potentials, defer attaching until later
         neighbor_list = hoomd.md.nlist.Tree(buffer=sim.potentials.pair.neighbor_buffer)
```
mphoward, line 76: One thing we do still want to do is make sure we can specify the neighbor list exclusions, and then only do the exclusions in the thermo if they are active. I think we talked about adding that as an argument / property of the `PairPotentialTabulator`, like the `neighbor_buffer`, and then we would set it here by setting the `exclusions` option.

src/relentless/simulate/hoomd.py
```diff
@@ -1228,11 +1240,27 @@ def act(self, timestep):
...
                             query_args = dict(self._rdf[i, j].default_query_args)
                             query_args.update(exclude_ii=(i == j))
+                            neighbors = (
+                                aabbs[j]
+                                .query(
+                                    snap.particles.position[type_masks[i]], query_args
+                                )
+                                .toNeighborList()
+                            )
+                            if snap.bonds is not None:
```
mphoward, line 1250: Here you would add a check if bonds are in the exclusion list as well. Probably you need to have that forwarded to this action though, since we would normally grab it from the `sim`.

mphoward (changes requested):
Thanks! This looks like it should work. I would try to also implement the logic about the neighbor list exclusions, but it should do what Tyla wants for the time being.

src/relentless/simulate/hoomd.py
```diff
@@ -1228,11 +1240,27 @@ def act(self, timestep):
                         for i, j in self._rdf:
                             query_args = dict(self._rdf[i, j].default_query_args)
                             query_args.update(exclude_ii=(i == j))
+                            neighbors = (
+                                aabbs[j]
+                                .query(
+                                    snap.particles.position[type_masks[i]], query_args
```
clpetix, line 1246: @mphoward , I think this should be type `j` since aabbs above is calculated for `i`

src/relentless/simulate/hoomd.py
```diff
@@ -64,6 +64,8 @@ def _call_v3(self, sim):
         # parse masses by type
         snap = sim["engine"]["_hoomd"].state.get_snapshot()
         sim.masses = self._get_masses_from_snapshot(sim, snap)
+        sim[self]["_bonds"] = self._get_bonds_from_snapshot(sim, snap)
+        sim[self]["_angles"] = self._get_angles_from_snapshot(sim, snap)
         self._assert_dimension_safe(sim, snap)
         # create the potentials, defer attaching until later
         neighbor_list = hoomd.md.nlist.Tree(buffer=sim.potentials.pair.neighbor_buffer)
```
clpetix, line 71: add in exclusions=('bond', 'angle')

PR #239, opened 2024-04-02
Add support for freud version 3

clpetix:
Support freud 3 release. This made changes to the RDF interface for normalization.

### DIFF
--- a/src/relentless/simulate/hoomd.py
+++ b/src/relentless/simulate/hoomd.py
@@ -1564,7 +1564,10 @@ class HOOMD(simulate.Simulation):
 
         if not _freud_found:
             raise ImportError("freud not found.")
-        elif packaging.version.parse(freud.__version__).major != 2:
+        elif (
+            packaging.version.parse(freud.__version__).major != 2
+            or packaging.version.parse(freud.__version__).major != 3
+        ):
             raise ImportError("Only freud 2.x is supported.")
 
         super().__init__(initializer, operations)
--- a/src/relentless/simulate/hoomd.py
+++ b/src/relentless/simulate/hoomd.py
@@ -50,6 +50,8 @@ if packaging.version.Version(_gsd_version) >= packaging.version.Version("2.8.0")
 else:
     _gsd_write_mode = "wb"
 
+_freud_version = packaging.version.parse(freud.__version__)
+
 
 # initializers
 class InitializationOperation(simulate.InitializationOperation):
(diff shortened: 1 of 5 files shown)


clpetix:
@mphoward this gives us compatibility with freud version 3. 

Additionally, there was a point where freud was an optional import and we had logic to check that. We also documented that freud was required for HOOMD but not for LAMMPS. I cleaned that up to make the two files fully analogous to their treatment/documentation of freud.

src/relentless/simulate/hoomd.py
```diff
@@ -49,13 +50,6 @@ class Flags(enum.IntEnum):
 else:
     _gsd_write_mode = "wb"
 
-try:
```
mphoward, line 52: I'm good with making freud a required import above, but you should follow the pattern for the other packages of making a `_freud_version` variable here that we can access below:
```
_freud_version = packaging.version.parse(freud.__version__)
```

src/relentless/simulate/hoomd.py
```diff
@@ -1263,11 +1257,22 @@ def reset(self):
             if self.rdf_params is not None:
                 self._rdf = collections.PairMatrix(self.types)
                 for i, j in self._rdf:
-                    self._rdf[i, j] = freud.density.RDF(
-                        bins=self.rdf_params["bins"],
-                        r_max=self.rdf_params["stop"],
-                        normalize=(i == j),
-                    )
+                    if packaging.version.parse(freud.__version__).major == 2:
```
mphoward, line 1260: Two things here:
1. I always put the higher version first in an if-elif so that it's easier to drop the lower versions later on.
2. There is a lot of duplication here because we actually only need to change the `normalize` vs. `normalization` keyword. Can we use a dict to simplify this like:
```
freud_rdf_args = dict(
    bins=self.rdf_params["bins"],
    r_max=self.rdf_params["stop"],
)
if _freud_version.major == 3:
    freud_rdf_args.update(normalization_mode="finite_size" if i ==j else "exact")
elif _freud_version.major == 2:
    freud_rdf_args.update(normalize=(i==j))
self._rdf[i, j] = freud.density.RDF(**freud_rdf_args)
```

src/relentless/simulate/hoomd.py
```diff
@@ -1533,7 +1538,7 @@ class HOOMD(simulate.Simulation):
     will be automatically selected for you when the simulation is run. Both
     HOOMD 2.x and 3.x are supported.
 
-    The `freud <https://freud.readthedocs.io>`_ analysis package (version 2.x)
```
mphoward, line 1536: `freud` is actually a dependency for all of relentless, so let's just remove this discussion here. Leave the parts about hoomd.

src/relentless/simulate/lammps.py
```diff
@@ -1129,11 +1129,22 @@ def process(self, sim, sim_op):
             if mpi.world.rank_is_root:
                 rdf_ = collections.PairMatrix(sim.types)
                 for i, j in rdf_:
-                    rdf_[i, j] = freud.density.RDF(
-                        bins=sim[self]["_rdf_params"]["bins"],
-                        r_max=sim[self]["_rdf_params"]["stop"],
-                        normalize=(i == j),
-                    )
+                    if packaging.version.parse(freud.__version__).major == 2:
```
mphoward, line 1132: Apply all the same comments for this file as for `hoomd.py`

src/relentless/simulate/lammps.py
```diff
@@ -1423,6 +1434,12 @@ class LAMMPS(simulate.Simulation):
 
         relentless.simulate.LAMMPS(init, ops, executable="mpiexec -n 8 lmp_mpi")
 
+    The `freud <https://freud.readthedocs.io>`_ analysis package (version >= 2.x)
+    is also required for initialization and analysis. To use this simulation backend,
+    you will need to install both :mod:`hoomd` and :mod:`freud` into your Python
+    environment. :mod:`hoomd` is available through conda-forge or can be built
+    from source, while :mod`freud` is available through both PyPI and conda-forge.
+    Please refer to the package documentation for details of how to install these.
```
mphoward, line 1442: Remove this based on comment above.
```suggestion
```

mphoward (changes requested):
Thanks for working on this! I have a couple comments to simplify the changeset, but then this will be good to merge

clpetix:
@mphoward I've got the changes applied, but one specific mpi test is failing. I've tried it a couple times but no luck. Could you let me know if you have any idea why that is ?

mphoward (approved):
Thanks!

PR #246, opened 2024-04-25
Consider both ij and ji pairs in RDF calculations

clpetix:


clpetix:
@mphoward this should have everything ported and it seems to be working in both HOOMD and LAMMPS!

src/relentless/simulate/lammps.py
```diff
@@ -1164,22 +1164,50 @@ def process(self, sim, sim_op):
...
+                        # neighbors of j that are type i
+                        if j != i:
+                            filter_ji = numpy.logical_and(
+                                type_masks[j][neighbors[:, 0]],
+                                type_masks[i][neighbors[:, 1]],
+                            )
+                            filter_ij = numpy.logical_or(filter_ij, filter_ji)
+                        neighbors_ij.filter(filter_ij)
+                        print("types: ", i, j)
+                        print(neighbors_ij[:])
```
mphoward, line 1202: Please remove these debug statements!

mphoward (approved):
Thanks! This all looks good to me after you remove the debug prints. Did you run this on a WCA mixture to make sure everything looks good?

clpetix:
@mphoward I have a working version of the code ported to both LAMMPS and HOOMD and it replicates the results from before! 
image

clpetix:
@mphoward this is ready for a review whenever you get a chance!

clpetix:
> Looks great! Just a few small things to clean up and then we can merge. Just to confirm: did you run a 2d WCA simulation in addition to the 3d check?

Yes, I tried all three cases with both engines in both 2d and 3d boxes and they are in agreement! Here are the 2-d cases: 

image

clpetix:
@mphoward the changes have been applied and this should be ready to go!

mphoward (approved):
Looks great, thanks! Can you merge `main` into the box resize branch for Beth after I merge this PR? We should also do this for Tyla's, but that is going to cause merge conflicts that we will need to resolve so I would rather punt on that for now.

Issue #251, opened 2024-06-20
`scale` not working as documented for non-one scaling factors


clpetix:
In a relentless script the scaling factors were defined as 10 with a 0.5 step-size. The gradient was -0.009799 and the variable went from 26.8087 to 26.8577. This is equivalent to the the scale factor being `10^(1/2)` rather than 10. 

The second line should be multiplied by the scale factor: 
https://github.com/mphowardlab/relentless/blob/dacdb02c34fc84a2aba7c9b9efc7f576b4d6bfc6/src/relentless/optimize/method.py#L400-L401

We should also add tests to catch this issue.

PR #252, opened 2024-06-20
Fix scale to work as documented

clpetix:
See issue #251

### DIFF
--- a/src/relentless/optimize/method.py
+++ b/src/relentless/optimize/method.py
@@ -398,7 +398,7 @@ class SteepestDescent(Optimizer):
         cur_res = objective.compute(variables, cur_dir)
         while not self.stop.converged(cur_res) and iter_num < self.max_iter:
             grad_y = scale * cur_res.gradient
-            update = self.descent_amount(grad_y) * grad_y
+            update = scale * self.descent_amount(grad_y) * grad_y
 
             # steepest descent update
             for x in variables:
--- a/tests/optimize/test_method.py
+++ b/tests/optimize/test_method.py
@@ -178,6 +178,13 @@ class test_SteepestDescent(unittest.TestCase):
         o.max_iter = 1
         self.assertFalse(o.optimize(objective=q, variables=x))
 
+        # test result of an interation of nontrivial scalar scaling parameter
+        x.value = 9
+        o.scale = 0.5
+        o.max_iter = 1
+        o.optimize(objective=q, variables=x)
+        self.assertAlmostEqual(x.value, 8)
+
         # test with nontrivial scalar scaling parameter
         x.value = 50
         o.scale = 0.85


clpetix:
@mphoward I've fixed this bug and added a unit test to catch it in the future! I think the tests are failing related to numpy 2.0, so this should be good for you to look at when you get a chance!

mphoward:
Thanks! This should fix the issue.

While investigating this and looking at the manuscript, I realized that the `LineSearch` is using unscaled variables. This is consistent with what is written in the documentation, but I think is inconsistent with the `SteepestDescent` using scaled variables to take steps. Could you look at this to confirm and see if you think we should add `scale` as an argument of `find`? We could default it to `scale=1.0` for backward compatibility.

If you agree, I would try to fix that in this PR since it is a related issue.

clpetix:
> Thanks! This should fix the issue.
> 
> While investigating this and looking at the manuscript, I realized that the `LineSearch` is using unscaled variables. This is consistent with what is written in the documentation, but I think is inconsistent with the `SteepestDescent` using scaled variables to take steps. Could you look at this to confirm and see if you think we should add `scale` as an argument of `find`? We could default it to `scale=1.0` for backward compatibility.
> 
> If you agree, I would try to fix that in this PR since it is a related issue.

So currently `SteepestDescent` is being updated using the scaled variables but `LineSearch` is checking based on unscaled variables? I think we would just need to add in `scale=1.0` as suggested and turn: 
https://github.com/mphowardlab/relentless/blob/09505911377030e1db8786d6d8d412bbe785589e/src/relentless/optimize/method.py#L192
to 
`x.value = start.variables[x] + scale**2 * new_step * d[x]`. 
I think we need the `scale**2` because the gradient doesn't have the extra factor from using scaled variables like it does in steepest descent here:
https://github.com/mphowardlab/relentless/blob/09505911377030e1db8786d6d8d412bbe785589e/src/relentless/optimize/method.py#L400

mphoward:
I think we need to apply one `scale` factor to any gradient that is dotted against `d`, which turns it into the gradient with respect to the scaled parameters. See Eq. (18) of the draft manuscript.

I also think when we calculate `d` in the first place, we should be dividing by `scale`. That would calculate the descent direction in scaled coordinates. See Eqs. (17) and (18) of the draft manuscript.

If you scale `d` when you compute it, then I think you will need to multiply by one factor of `scale` in line 192 to make the update apply to the unscaled parameters.

clpetix:
@mphoward that all makes sense! I think I have those changes implemented as you noted!

mphoward:
Thanks, here are a few things I noticed:

* The multiplication of the gradient by scale needs to be done on the gradient directly (not outside the dot) in case scale is a KeyedArray.
* I think you may have missed some instances of dotting in find, so please search for those.
* You should copy the logic from SteepestDescent for converting the scale (as given) into a KeyedArray with the right keys, in case. Alternatively, we need to document that scale either needs to be just a float or a properly keyed KeyedArray.

clpetix:
@mphoward here are those changes!

mphoward (approved):
Looks great! Can you document `scale` in the `LineSearch`? Then this will be good to go.

mphoward (approved):
Looks good to me, thanks for the fix!

PR #253, opened 2024-06-25
Make relentless compatible with NumPy 2

mphoward:
* Update test requirements
* Create a wrapper in math module for trapezoidal integration
* Compute 2d cross product in a different way to avoid deprecation warning
* Warn about NumPy 2 with HOOMD 2 (produces errors)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,6 +80,10 @@ jobs:
           channel-priority: true
           mamba-version: "*"
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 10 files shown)


mphoward:
@clpetix this should be good to go! I spot checked the output for any other deprecation warnings. I had to manually configure the NumPy dependency for HOOMD 2 (since I got failures running with NumPy 2). We can't really stop someone from doing this, so I added a runtime warning.

clpetix (approved):
This all looks good to me! 

I had missed the 2-D cross deprecation when going through their change-log :sweat_smile:

PR #257, opened 2024-06-28
Support HOOMD 4

### LINKED ISSUE #214: Support HOOMD 4
There are some breaking changes that we need to adapt to. We should also streamline the code that selects the version code path to simplify the maintenance process.

### LINKED ISSUE #228: Add option to specify whether HOOMD runs on CPU or GPU
Currently this selection is automatic, but it could be nice to force it to run one way or the other (e.g., on CPU even when GPU is present).

mphoward:
* Adds compatibility with HOOMD 4 per [migration guide](https://hoomd-blue.readthedocs.io/en/v4.7.0/migrating.html#migrating-to-hoomd-v4). 
* Adds option to specify device to use for simulations. Default behavior to auto-select is preserved.

Fixes #214
Fixes #228

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,7 +66,10 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
-        hoomd-version: ["2.9.7", "3.11.0"]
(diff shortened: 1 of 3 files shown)


clpetix (approved):
This all makes sense and looks good to me! Thanks for taking care of this!

PR #259, opened 2024-08-06
Run GitHub Actions on a regular interval

clpetix:
We should run GitHub actions on a regular interval so that we catch if the code breaks because of our dependencies (such as what happened with Numpy 2.0). 

This will run our actions at 08:00 AM, on the first Tuesday of the month.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,8 @@ on:
     branches:
       - main
 
+  schedule:
+    - cron: "0 0 8 * 1/1 TUE#1"
   workflow_call:
 
 concurrency:


mphoward:
Thank you! I think this is a good idea, but there are some caveats on scheduled jobs we need to be careful with. Please review the list here and let’s discuss via Slack after you’ve done so:

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

clpetix:
GitHub provides the following guidance: 

- The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. If the load is sufficiently high enough, some queued jobs may be dropped. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.
  - We had discussed wanting a mid-morning run to increase the likeliness that someone is around and able to address any issues. I set these to run on the 1st and 15th of every month at 10:18 a.m. to avoid the delay mentioned above. 
- This event will only trigger a workflow run if the workflow file is on the default branch.
- Scheduled workflows will only run on the default branch.
  - Both of these are fine, we test our development branches with pushes to pull requests. 
- In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days. For information on re-enabling a disabled workflow, see "[Disabling and enabling a workflow](https://docs.github.com/en/enterprise-server@3.12/actions/using-workflows/disabling-and-enabling-a-workflow#enabling-a-workflow)."
  - With our active development this should be fine, if it does happen to go inactive you can easily re-enable it. My experience with this from MolSSI's bootcamp is that they will also send a bunch of emails warning you that it is about to go inactive. 
- When the last user to commit to the cron schedule of a workflow is removed from the organization, the scheduled workflow will be disabled. If a user with write permissions to the repository makes a commit that changes the cron schedule, the scheduled workflow will be reactivated. Note that, in this situation, the workflow is not reactivated by any change to the workflow file; you must alter the cron value and commit this change.
  - You can either change my value of 18 now or later, if I am removed from the organization, you can pick your favorite non-zero number and change the minute value to re-enable this. 
  
@mphoward, based on this I think we would be good to implement this!

mphoward:
I agree with all these comments, thanks for taking the time to go through it carefully!

Issue #260, opened 2024-08-21
Convert time steps to integer before sending to the simulation engines 


clpetix:
When specifying `steps` or `every` using scientific notation (e.g., `1e5`) the value is a float. This causes an error when running in LAMMPS. We should ensure these values are passed to the engines as integers.

mphoward:
Fixed in #263

PR #263, opened 2024-08-22
Ensure LAMMPS gets sent integer time values. 

clpetix:
Addresses Issue #260

### DIFF
--- a/src/relentless/simulate/lammps.py
+++ b/src/relentless/simulate/lammps.py
@@ -509,7 +509,7 @@ class MinimizeEnergy(SimulationOperation):
     def __init__(self, energy_tolerance, force_tolerance, max_iterations, options):
         self.energy_tolerance = energy_tolerance
         self.force_tolerance = force_tolerance
-        if max_iterations.is_integer():
+        if float(max_iterations).is_integer():
             self.max_iterations = int(max_iterations)
         else:
             raise ValueError("Max iterations should be an integer.")
@@ -517,7 +517,7 @@ class MinimizeEnergy(SimulationOperation):
 
     def _call_commands(self, sim):
         max_eval = self.options.get("max_evaluations", 100 * self.max_iterations)
-        if max_eval.is_integer():
+        if float(max_eval).is_integer():
             max_eval = int(max_eval)
         else:
             raise ValueError("Max evalulations should be an integer.")
@@ -556,7 +556,7 @@ class _Integrator(SimulationOperation):
 
     def __init__(self, steps, timestep, analyzers):
         super().__init__(analyzers)
-        if steps.is_integer():
+        if float(steps).is_integer():
             self.steps = int(steps)
         else:
             raise ValueError("Steps should be an integer.")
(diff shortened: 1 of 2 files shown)


src/relentless/simulate/lammps.py
```diff
@@ -1641,3 +1643,10 @@ def _initialize_engine(self, sim):
     _EnsembleAverage = EnsembleAverage
     _Record = Record
     _WriteTrajectory = WriteTrajectory
+
+
+def to_int(a):
```
mphoward, line 1648: Let's make this explicitly private. These modules are really part of the public API, but I don't want to promise this function will always work. You should also add a doc string.
```suggestion
def _to_int(a):
```

tests/simulate/test_lammps.py
```diff
@@ -246,6 +246,28 @@ def test_minimization(self):
         lmp.operations = emin
         lmp.run(potentials=pot, directory=self.directory)
 
+        # Whole number of iterations and evalulations as float
+        emin.max_iterations = 1000.0
```
mphoward, line 250: Did you confirm these tests fail without the bugfix?

tests/simulate/test_lammps.py
```diff
@@ -246,6 +246,28 @@ def test_minimization(self):
...
+        emin.options = {"max_evaluations": 10000.0}
+        lmp.operations = emin
+        lmp.run(potentials=pot, directory=self.directory)
+        emin.max_iterations = 1000
+        emin.options = {"max_evaluations": 10000}
```
mphoward, line 255: What is the purpose of these lines here? It wasn't clear to me.

tests/simulate/test_lammps.py
```diff
@@ -258,6 +280,18 @@ def test_brownian_dynamics(self):
         else:
             h.run(pot, self.directory)
 
+        # Whole number of steps as float
+        brn.steps = 1.0
+        h = relentless.simulate.LAMMPS(init, brn, executable=self.executable)
```
mphoward, line 285: You don't need to make new simulation engines to do a new run, remove these extra instances throughout.

mphoward (changes requested):
Thanks! Just a few small things and then this can be merged.

tests/simulate/test_lammps.py
clpetix: yes

tests/simulate/test_lammps.py
clpetix: The order of those tests are: 
1.) whole number iterations and max evaluations as floats
2.) Non whole number of iterations
3.) Non whole number of max evaluations

I thought it was best to reset it after 1 to the default integer state before testing 2 & 3.

mphoward: OK! Maybe add some comments documenting that the first time you use that pattern?

clpetix: I added comments whenever I reset something. I think that should help the readability!

mphoward (approved):
Thank you! This is a clean solution.

Issue #217, opened 2023-07-27
Scrub temporary data and files from simulation output


mphoward:
We should automatically remove any temporary files that are created during the simulation run. This could be done by checking for a `_files` data member in the simulation output that could be a string or dict, and attempting to remove anything in it from the output directory.

We should also remove any temporary data that is stashed in the simulation output at the end of a run. This could be done by removing any dictionary keys that are prefixed with a `_`.

Issue #179, opened 2023-03-03
Accept a single string for a tuple of types


mphoward:
This would be convenient to have for parametrizing potentials so we could do:
```
LennardJones("A")
```
rather than
```
LennardJones(("A",))
```

It's as simple as adding:
```
if isinstance(types, str):
    types = (types,)
```
as needed. Possibly it could be done just in the `Potential` base class even.

PR #272, opened 2024-10-02
Implement bond interactions

clpetix:
Implement bond interactions. WCA, FENE, and spline time bond interactions for both LAMMPS and HOOMD with interconversion using lammpsio.


clpetix:
At this point HOOMD bond interactions and exclusions are working. Here are the RDFs for a small simulation of dumbbells in a Langevin dynamics simulation with only a harmonic bond potential with `r0 = 2`. 
image

clpetix:
The version of the LAMMPS code is working when initializing from HOOMD GSD. 
image

clpetix:
Known tasks left to do: 

- [ ] Get the code working when coming from a LAMMPS data file
- [ ] Update the WriteTrajectory both for LAMMPS schema and when converting to HOOMD GSD
- [ ] Sort out issue with bonds crossing periodic boundary condition 
- [ ] Set Atom style appropriately in `_initialize_engine()`

mphoward:
Thanks! It would be good to also have a reference simulation run natively in HOOMD with their potentials that we can compare quantities against. For example, the distribution of bond lengths and g(r), similar to what we have done with WCA spheres in the past.

We might consider adding these types of scripts in a `validate` directory.

clpetix:
@mphoward could you have a look at this when you get a chance? I have messed something up with MPI but the tests pass on 1 core. 

I want to get a check on the implementation before I start adding lots of tests!

src/relentless/model/potential/__init__.py
```diff
@@ -29,6 +29,7 @@
 
 """
 
+from .bond import Harmonic
```
mphoward, line 32: We should be careful about this naming and import since we will quickly run into an issue when there are also harmonic angles, harmonic dihedrals, etc. For the pairs, I followed the convention that I put `Pair` somewhere in the class name if it was ambiguous. So this would change to `HarmonicBond`, but `FENE` would stay `FENE`. (Side note: don't forget to import FENE here.)

The other option would be not to import the bonded potentials into this namespace. The options would be:

* `relentless.model.potential.bond.Harmonic` vs.
* `relentless.model.potential.HarmonicBond`

I prefer the second.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+
+    where :math:`r` is the distance between two bonded particles. The parameters
+    for each type are:
+
+    +-------------+--------------------------------------------------+-----------+
+    | Parameter   | Description                                      | Initial   |
+    +=============+==================================================+===========+
+    | ``k``       | potential constant                               |           |
+    +-------------+--------------------------------------------------+-----------+
+    | ``r0``      | rest length.                                     |           |
```
mphoward, line 35: Clean up these tables to look like the pair potentials:
```suggestion
    | ``k``       | Spring constant :math:`k`.                               |           |
    +-------------+--------------------------------------------------+-----------+
    | ``r0``      | Minimum-energy length :math:`r_0`.                                     |           |
```
Given that I don't think these potentials are going to have default parameters, I think you can also remove the `Initial` column.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+    +-------------+--------------------------------------------------+-----------+
+
+    Parameters
+    ----------
+    type : str
+        Type.
```
mphoward, line 41: Check how this is documented in pair, but this potential should support multiple bond types.
```suggestion
    types: tuple[str]
        Types.
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+    Attributes
+    ----------
+    coeff : :class:`BondParameters`
+        Parameters of the potential for each type.
+
+    Examples
+    --------
+    Harmonic Bond::
+
+        >>> u = relentless.potential.bond.Harmonic("A")
```
mphoward, line 55: Does this actually work? These are supposed to be tuples of types, although extending to a single string would be a nice convenience (see #179).

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+        >>> u.coeff["A"].update({'k': 1.0, 'r0': 1.0})
```
mphoward, line 56: I like to give examples with sensible choices. A spring constant of 1 is too small for most bonds.
```suggestion
        >>> u.coeff["A"].update({'k': 1000, 'r0': 1})
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+    """
+
+    def __init__(self, types, name=None):
+        super().__init__(keys=types, params=("k", "r0"), name=name)
+
+    def energy(self, type, r):
+        """Evaluate bond energy."""
+        k = self.coeff[type]["k"]
+        r0 = self.coeff[type]["r0"]
+        return k / 2 * (r - r0) ** 2
```
mphoward, line 67: I found the divide by 2 hard to read for some reason... either use 0.5 or add parentheses
```suggestion
        return 0.5 * k * (r - r0) ** 2
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+        if var == "k":
+            return (r - r0) ** 2 / 2
+        elif var == "r0":
+            return -k * (r - r0)
+        else:
+            raise ValueError("Unknown parameter")
+
+
+class FENE(BondPotential):
+    r"""The Finite Extensible Nonlinear Elastic (FENE) bond interaction.
```
mphoward, line 88:
```suggestion
    r"""Finitely extensible nonlinear elastic (FENE) bond interaction.
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+
+    .. math::
```
mphoward, line 90: Since we have realized for Tyla's project how important exclusions can be, I am wondering if we should enable the WCA part of the FENE interaction in case someone needs 1-2 exclusions for other parts of the model. We could do this in two ways:

* Add the WCA part here, with epsilon and sigma optional arguments defaulted to 0.
* Create a separate `FENEWCA` potential that derives from `FENE` and adds the `epsilon` and `sigma` parameters.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+    +=============+==================================================+===========+
+    | ``k``       | attractive force strength.                       |           |
+    +-------------+--------------------------------------------------+-----------+
+    | ``r0``      |  size parameter.                                 |           |
+    +-------------+--------------------------------------------------+-----------+
+
+    Parameters
+    ----------
+    type : str
+        Type.
```
mphoward, line 109: Fix up as above.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+    ----------
+    coeff : :class:`BondParameters`
+        Parameters of the potential for each type.
+
+    Examples
+    --------
+    Harmonic Bond::
+
+        >>> u = relentless.potential.bond.FENE("A")
+        >>> u.coeff["A"].update({'k': 1.0, 'r0': 1.0})
```
mphoward, line 124: Same comment about type list, plus here are the typical K-G parameters for FENE:
```suggestion
    Kremer-Grest bond (no repulsion included)::

        >>> u = relentless.potential.bond.FENE("A")
        >>> u.coeff["A"].update({'k': 30, 'r0': 1.5})
```
add a second example with WCA (epsilon = 1, sigma = 1) if you add that to this class.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+            raise ValueError("Unknown parameter")
+
+
+class FENE(BondPotential):
+    r"""The Finite Extensible Nonlinear Elastic (FENE) bond interaction.
+
+    .. math::
+
+        u(r) = -\frac{k r_0^2}{2} \ln \left(1 -
+            \left(\frac{r - r_0}{r_0}\right)^2\right)
```
mphoward, line 93: This should be written like this:
```suggestion
        u(r) = -\frac{k r_0^2}{2} \ln \left[1 -
            \left(\frac{r}{r_0}\right)^2\right]
```
I will check if the equations are right below or not

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+    """
+
+    def __init__(self, types, name=None):
+        super().__init__(keys=types, params=("k", "r0"), name=name)
+
+    def energy(self, type, r):
+        """Evaluate bond energy."""
+        k = self.coeff[type]["k"]
+        r0 = self.coeff[type]["r0"]
+        return -0.5 * k * r0**2 * numpy.log(1 - (r - r0) ** 2 / r0**2)
```
mphoward, line 134:
```suggestion
        return -0.5 * k * r0**2 * numpy.log(1 - (r / r0) ** 2)
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def force(self, type, r):
+        """Evaluate bond force."""
+        k = self.coeff[type]["k"]
+        r0 = self.coeff[type]["r0"]
+        return k * r0**2 / (r0**2 - (r - r0) ** 2)
```
mphoward, line 140:
```suggestion
        return - k * r / (1 - (r / r0) ** 2)
```

Also, what happens if r >= r0? The potential should probably be inf.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def derivative(self, type, var, r):
+        r"""Evaluate bond derivative with respect to a variable."""
+        k = self.coeff[type]["k"]
+        r0 = self.coeff[type]["r0"]
+        if var == "k":
+            return -0.5 * r0**2 * numpy.log(1 - (r - r0) ** 2 / r0**2)
```
mphoward, line 147:
```suggestion
            return -0.5 * r0**2 * numpy.log(1 - (r / r0)**2)
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+        elif var == "r0":
+            return k * r0 * (r - r0) / (r0**2 - (r - r0) ** 2)
```
mphoward, line 149: This expression is not right, I don't think, because of a product rule. Check with Mathematica, but I think this would be right:
```suggestion
            r_r0_2 = (r / r0) ** 2
            x = 1 - r_r0_2
            return k * r0 * (-r_r0_2 / x - numpy.log(x))
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,425 @@
...
+        else:
+            raise ValueError("Unknown parameter")
+
+
+class BondSpline(BondPotential):
```
mphoward, line 154: This is producing a lot of duplicated code with PairSpline. It would be a good idea to create a private class or mixin somewhere for a spline potential.

src/relentless/simulate/analyze.py
```diff
@@ -73,10 +73,15 @@ def _get_rdf_params(self, sim):
             else:
                 rdf_every = self.every
 
+            if "exclude" in self.rdf:
+                rdf_exclude = self.rdf["exclude"]
```
mphoward, line 77: Probably do like this to make sure it is cast to True/False
```suggestion
                rdf_exclude = bool(self.rdf["exclude"])
```

src/relentless/simulate/hoomd.py
```diff
@@ -1188,6 +1214,8 @@ def __init__(
             self.system = system
             self.rdf_params = rdf_params
             self.constraints = constraints if constraints is not None else {}
+            self.bonds = bonds
+            self.exclusion = exclusions
```
mphoward, line 1218: This would be `exclusions` and come first, following the other code patterns
```suggestion
            self.exclusions = exclusions
            self.bonds = bonds
```

src/relentless/simulate/hoomd.py
```diff
@@ -1282,6 +1310,29 @@ def act(self, timestep):
...
+                        # bond exclusions apply regardless of order, so
+                        # consider both (i,j) and (j,i) permutations
+                        if (
+                            self.rdf_params["exclude"]
+                            and snap.bonds.N != 0
+                            and len(neighbors[:]) > 0
+                        ):
+                            bonds = numpy.vstack(
+                                [self.bonds, numpy.flip(self.bonds, axis=1)],
+                            )
```
mphoward, line 1323: This is a small optimization suggestion, but given that you need to Cantor pair the bonds now, would it be better to sort the rows of bonds and neighbors so you don't have to duplicate them in reverse order?

It would also be good to put the Cantor pairing function into a private function or method so you can reuse it, and it will be easier to read.

src/relentless/simulate/simulate.py
```diff
@@ -566,12 +566,18 @@ class PairPotentialTabulator(PotentialTabulator):
...
+    exclusions : list
+        The neighborlist nominally includes all pairs within ``rmax`` of each other.
+        This option allows for exclusions of pairs that should not be included in the
+        neighbor list. The string should be formatted as a tuple of strings, e.g.,
+        ``['bond', '1-3', 1-4]``.
 
     """
 
-    def __init__(self, potentials, start, stop, num, neighbor_buffer):
+    def __init__(self, potentials, start, stop, num, neighbor_buffer, exclusions=[]):
```
mphoward, line 577: This is one of those Python quirks you learn not to do the hard way. =) I would default this to None:
```suggestion
    def __init__(self, potentials, start, stop, num, neighbor_buffer, exclusions=None):
```
then reinterpret it as an empty list internally.

You can read here about why:

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/relentless/simulate/simulate.py
```diff
@@ -746,3 +761,29 @@ def pairwise_energy_and_force(self, types, x=None, tight=False, minimum_num=2):
...
+        be :class:`~relentless.potential.bond.BondPotential`\s.
+    start : float
+        The minimum value of ``r`` at which to tabulate.
+    stop : float
+        The maximum value of ``r`` at which to tabulate.
+    num : int
+        The number of points (value of ``r``) at which to tabulate and evaluate the
+        potential.
+    neighbor_buffer : float
+        Buffer radius used in computing the neighbor list.
```
mphoward, line 785:
```suggestion
```

mphoward (changes requested):
Thank you! This is a great first pass. Please review my comments and let me know if you have any questions.

src/relentless/simulate/simulate.py
clpetix: Thanks! I hadn't seen that before! 😅

src/relentless/model/potential/bond.py
clpetix: I kind of like the WCA+FENE since we're going to be very explicit with exclusions and default arguments. I just found HOOMD's a bit hard my first time using it with their default exclusions.

mphoward: I agree, and it will be less code to only have one class. Our documentation is pretty explicit about defaults.

clpetix:
@mphoward I think I've fixed all the issues that you pointed out and a couple more that I caught from those fixes! It should be ready for you to have another look at! I am going to start working on the tests and hopefully get this ready to merge in soon!

src/relentless/simulate/lammps.py
```diff
@@ -202,6 +203,7 @@ def _call_commands(self, sim):
         masses = mpi.world.bcast(masses)
         sim.masses.update(masses)
         dim_safe = mpi.world.bcast(dim_safe)
+        snap = mpi.world.bcast(snap)
```
mphoward, line 206: I don't think this is likely to work, since you are serializing and broadcasting a ton of data. What is it that actually needs to be shared across ranks—just the bond types? We can probably do that by making a dictionary from the label map and broadcasting that.

clpetix:
@mphoward, I got the derivative issues fixed and put unittests on the bond potentials! This should be ready for you to take a look at!

src/relentless/model/potential/__init__.py
```diff
@@ -26,9 +37,12 @@
     Parameters
     PairPotential
     PairParameters
+    BondPotential
+    BondParameters
```
mphoward, line 41: `BondedPotential` should probably go in this list of developer classes.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+    Attributes
+    ----------
+    coeff : :class:`BondParameters`
+        Parameters of the potential for each type.
+
+    Examples
+    --------
+    Harmonic Bond::
+
+        >>> u = relentless.potential.bond.Harmonic(("A"))
```
mphoward, line 56: I think this needs a comma to work, otherwise it is not a tuple.
```suggestion
        >>> u = relentless.potential.bond.Harmonic(("A",))
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+        >>> u.coeff["A"].update({'k': 1000, 'r0': 1})
+
+    """
+
+    def __init__(self, types, name=None):
+        super().__init__(keys=types, params=("k", "r0"), name=name)
+
+    def energy(self, types, r):
```
mphoward, line 64: Small thing: rename to either `type_` or `bond_type` throughout since there is only one type for a bond.
```suggestion
    def energy(self, type_, r):
```
You should not use `type` itself because that is a built-in function.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+
+        r, f, s = self._zeros(r)
+
+        f = -k * (r - r0)
+
+        if s:
+            f = f.item()
+        return f
+
+    def _derivative(self, param, r, k, r0, **params):
```
mphoward, line 92: Do any other parameters actually get passed to the bond potentials? In `PairPotential` `**params` is used to catch extra keywords like `rmax`, `shift`, etc. without having to spell them all out. Here, though, I don't think there would be any? If yes, it's better to remove.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+
+    where
+
+    .. math::
+        u_{\rm{WCA}}(r)  =
+        \begin{cases} 4 \varepsilon \left[ \left( \frac{\sigma}{r}
+                                \right)^{12} - \left( \frac{\sigma}{r}
+                                \right)^{6} \right]  + \varepsilon
+                               & r < 2^{\frac{1}{6}}\sigma\\
+        0 & r \ge 2^{\frac{1}{6}}\sigma
```
mphoward, line 125:
```suggestion
                                \right)^6 \right]  + \varepsilon
                               & r < 2^{1/6}\sigma\\
        0 & r \ge 2^{1/6}\sigma
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+        """Evaluate bond energy."""
+        params = self.coeff.evaluate(types)
+        k = params["k"]
+        r0 = params["r0"]
+        epsilon = params["epsilon"]
+        sigma = params["sigma"]
+
+        # initialize arrays
+        r, FENE, s = self._zeros(r)
+        WCA = FENE.copy()
```
mphoward, line 182: Prefer not to use all capitals for variables, as sometimes these indicate constants. This would be better:
```suggestion
        r, u_fene, s = self._zeros(r)
        u_wca = u_fene.copy()
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+
+        # set flags for FENE potential
+        fene_flag = ~numpy.greater_equal(r, r0)
```
mphoward, line 185: This should be the same but avoid the negation:
```suggestion
        fene_flag =numpy.less(r, r0)
```

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+
+        # evaluate FENE potential
+        FENE[fene_flag] = -0.5 * k * r0**2 * numpy.log(1 - (r[fene_flag] / r0) ** 2)
+        FENE[~fene_flag] = numpy.inf
+
+        # set flags for WCA potential
+        zero_flags = ~numpy.isclose(r, 0)
```
mphoward, line 192: This variable name confused me because `zero_flags` is actually the `nonzero_flags`! Either rename or remove the negation & reformulate below.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+        WCA[~zero_flags] = numpy.inf
+        WCA[~wca_flags] = 0
+
+        if s:
+            FENE = FENE.item()
+            WCA = WCA.item()
+
+        return FENE + WCA
+
+    def force(self, types, r):
```
mphoward, line 207: Same comments on this method.

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+        dWCA_dr[~zero_flags] = numpy.inf
+        dWCA_dr[~wca_flags] = 0
+
+        if s:
+            dFENE_dr = dFENE_dr.item()
+            dWCA_dr = dWCA_dr.item()
+
+        return dFENE_dr + dWCA_dr
+
+    def _derivative(self, param, r, k, r0, epsilon, sigma, **params):
```
mphoward, line 243: Same comments on this method regarding the flags. I did not check all the formulas below carefully, but I assume you have done them right in your unit tests / checked with Mathematica.

src/relentless/model/potential/potential.py
```diff
@@ -375,3 +375,111 @@ def save(self, filename):
         data = self.to_json()
         with open(filename, "w") as f:
             json.dump(data, f, sort_keys=True, indent=4)
+
+
+class BondedPotential(Potential):
+
+    def __init__(self, keys, params, container=None, name=None):
```
mphoward, line 382: Is this all the same as the base class defaults? If yes, you don't need to implement the `__init__` method since it will just pass through to the parent.

src/relentless/model/potential/potential.py
```diff
@@ -375,3 +375,111 @@ def save(self, filename):
         data = self.to_json()
         with open(filename, "w") as f:
             json.dump(data, f, sort_keys=True, indent=4)
+
+
+class BondedPotential(Potential):
```
mphoward, line 380: Class docstring missing. It should document why this class exists for developers.

src/relentless/model/potential/potential.py
```diff
@@ -375,3 +375,111 @@ def save(self, filename):
...
+
+    def __init__(self, keys, params, container=None, name=None):
+        super().__init__(keys, params, container, name)
+
+    def derivative(self, types, var, r):
+        r"""Evaluate pair derivative with respect to a variable.
```
mphoward, line 386: Update this docstring, it is copied from the pair potential.

src/relentless/model/potential/potential.py
```diff
@@ -375,3 +375,111 @@ def save(self, filename):
...
+            deriv[flags] += self._derivative(p, r[flags], **params) * dp_dvar
+
+        # coerce derivative back into shape of the input
+        if scalar_r:
+            deriv = deriv.item()
+        return deriv
+
+    @abc.abstractmethod
+    def _derivative(self, param, r, **params):
+        """Implementation of the parameter derivative function.
```
mphoward, line 462: Same here.

src/relentless/simulate/hoomd.py
```diff
@@ -79,7 +85,22 @@ def _call_v3(self, sim):
             )
             pair_potential.r_cut[(i, j)] = r[-1]
         sim[self]["_potentials"] = [pair_potential]
-        sim[self]["_potentials_rmax"] = r[-1]
+
+        sim[self]["_bonds"] = self._get_bonds_from_snapshot(sim, snap)
+        if snap.bonds.N > 0:
```
mphoward, line 90: I think these will cause problems if you try to run the simulation under MPI. The reason is that the snapshot is only valid on the root rank, and your `_get_bonds_from_snapshot` method does not broadcast.

Use `state.N_bonds` if you want access to the number of bonds on all ranks.

For the snapshot method, I would have `_get_bonds_from_snapshot` return `None` on non-root ranks, as I think only the root rank actually needs it later on to process the RDF, right?

Also a small conceptual question: if there are no bonds in the configuration but the user does specify bond potentials, should we attach them anyway? Is performance the reason not to do so?

src/relentless/simulate/lammps.py
mphoward: This comment still applies. This will need to be handled differently than broadcasting a snapshot. Let me know if you are unsure what to do here.

src/relentless/simulate/lammps.py
```diff
@@ -268,6 +303,17 @@ def pair_map(sim, pair):
         ]
         cmds += ["pair_style table linear {N}".format(N=Nr)]
 
+        # set exclusions if snap has topology
+        if snap.has_bonds() or snap.has_angles() or snap.has_dihedrals():
+            excl_12, excl_13, excl_14 = 1.0, 1.0, 1.0
+            if sim.potentials.pair.exclusions is not None:
+                if "bond" in sim.potentials.pair.exclusions:
```
mphoward, line 310: I think we said we were using 1-2 as the exclusion name, no?
```suggestion
                if "1-2" in sim.potentials.pair.exclusions:
```

src/relentless/simulate/simulate.py
```diff
@@ -78,6 +78,35 @@ def post_run(self, sim, sim_op):
     def process(self, sim, sim_op):
         pass
 
+    def _cantor_pairing(self, connection, neighbor):
+        """Compute the list intersect using the Cantor pairing function to
+        filter neighborlist.
+
+        https://en.wikipedia.org/wiki/Pairing_function
+
+        Parameters
```
mphoward, line 87: Name same as function signature.

src/relentless/simulate/simulate.py
```diff
@@ -566,12 +595,18 @@ class PairPotentialTabulator(PotentialTabulator):
         potential.
     neighbor_buffer : float
         Buffer radius used in computing the neighbor list.
+    exclusions : list
+        The neighborlist nominally includes all pairs within ``rmax`` of each other.
+        This option allows for exclusions of pairs that should not be included in the
+        neighbor list. The string should be formatted as a tuple of strings, e.g.,
+        ``['bond', '1-3', 1-4]``.
```
mphoward, line 602: Here again, I think we said 1-2 right? I would also add a bulleted list of allowed values in this documentation and what they mean.

mphoward (changes requested):
Looking good! I reviewed the main code but not the tests yet. Please address my comments there, and let me know if you have any questions. Then I can go that a quick look over again, along with the testing.

I would also really like to see the addition of some sort of simulation test where you measure distribution functions. That can be handled outside this repo for now, so long as you have a script you can run.

src/relentless/simulate/hoomd.py
clpetix: I think performance would be the reason not to, but I don't know how much of an issue that would cause. If we did that, would just a try block work? This will throw an error if `sim.potentials.bond` attribute isn't on Potentials.

clpetix:
@mphoward, I think I have all of these points addressed! I am also working in tandem to apply relevant changes to the angles & dihedral branches as well to hopefully make those reviews smoother. 

I will work on getting a simulation test measuring distribution functions. I have some that I have been using for myself, but I will clean them up to be used formally. 

This should be ready for you to look at again and tests, whenever you get a chance!

src/relentless/simulate/lammps.py
```diff
@@ -212,7 +224,19 @@ def _call_commands(self, sim):
             sim.types, x=sim.potentials.pair.squared_space, tight=True, minimum_num=2
         )
         Nr = len(r)
-        sim[self]["_potentials_rmax"] = r[-1]
+
+        if has_bonds:
+            uB = collections.FixedKeyDict(sim[self]["bonds"].type_label.types)
+            fB = collections.FixedKeyDict(sim[self]["bonds"].type_label.types)
```
mphoward, line 230: Won't `sim[self]["bonds"]` only be available on the root rank? You probably need to store the bond type map in a plain dict and broadcast it, like is done for the particle types. The array of bonds themselves can be stored in a different key & only on the root rank (None on the other ranks) like for the HOOMD snapshot.

src/relentless/simulate/lammps.py
```diff
@@ -257,6 +281,27 @@ def pair_map(sim, pair):
                         fw.write(
                             "{idx} {r} {u} {f}\n".format(idx=idx, r=r_, u=u_, f=f_)
                         )
+
+                # write bond potentials into the file
+                if has_bonds:
+                    fw.write("# LAMMPS tabulated bond potentials\n")
```
mphoward, line 287: Just to confirm: LAMMPS is fine with both the pair and bond potential being in the same file?

src/relentless/simulate/hoomd.py
mphoward: `Potentials` should have a `bond` property that defaults to `None`, like is done for `pair`. The setter should do type validation on the tabulator as well. Once that's in place, checking for the existence of bond potentials is `sim.potentials.bond is not None`.

tests/model/potential/test_bond.py
```diff
@@ -0,0 +1,679 @@
...
+
+import tempfile
+import unittest
+
+import numpy
+
+import relentless
+
+
+class test_BondParameters(unittest.TestCase):
```
mphoward, line 11: Since `BondParameters` is a pass through class, it is not necessary to duplicate a test that exists elsewhere (I'm assuming this is copied, yes?)

tests/model/potential/test_bond.py
mphoward, line 1: This test file looks good! We still need at least one basic unit test for HOOMD and LAMMPS that they can run with a bond potential attached, though.

mphoward (changes requested):
I think this is close!

src/relentless/model/potential/bond.py
```diff
@@ -0,0 +1,568 @@
...
+    +-------------+--------------------------------------------------+-----------+
+    | Parameter   | Description                                      | Initial   |
+    +=============+==================================================+===========+
+    | ``k``       | Spring constant :math:`k`.                       |           |
+    +-------------+--------------------------------------------------+-----------+
+    | ``r0``      | Minimum-energy length :math:`r_0`.               |           |
+    +-------------+--------------------------------------------------+-----------+
+    | ``epsilon`` | Interaction energy :math:`\varepsilon`.          |     0     |
+    +-------------+--------------------------------------------------+-----------+
+    | ``sigma``   | Interaction length :math:`\sigma`.               |     0     |
```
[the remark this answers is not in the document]
clpetix: Ope, that must just be a copy and past error. I think we should probably just leave everything unset and force the user to make a decision.

src/relentless/simulate/lammps.py
clpetix: My understanding is that the bond_coeff command takes the file name argument and then for a specific entry in that file just like pair. The LAMMPS examples provided being: 
```
bond_style table linear 1000
bond_coeff 1 file.table ENTRY1
```
and 
```
pair_style table linear 1000
pair_style table linear 1000 pppm
pair_style table bitmap 12
pair_coeff * 3 morse.table ENTRY1
pair_coeff * 3 morse.table ENTRY1 7.0
```
I've not had any issues with them being in the same file, but if you think that could cause a problem, we can change it!

mphoward: If it works, I'm good with it! It will make it cleaner to scrub this file later if we want to (see #217)

src/relentless/simulate/lammps.py
clpetix: I'm a bit confused on this portion. Would the bond type map be made above where the other portions are being broadcast?

tests/model/potential/test_bond.py
clpetix: What do you think would be the best method / specific things to test are here? I want to make sure we have adequate testing without too much duplication.

mphoward: In `test_hoomd.py`, etc. you need to have at least one test where the potentials are attached and you try to run a few steps, to make sure the simulation at least runs without error. Otherwise, none of the code that was added for running the simulations is actually unit tested.

To make this test stronger, you could dump a trajectory, read it back in, and make sure 2 bonded particles remain close to each other whereas they would drift apart without it.

src/relentless/simulate/lammps.py
mphoward: If you run this in MPI, does it work or does it throw an error that it can't find the `"bonds"` key? I think it will error out.

You need to make a dictionary from the type map while you have the snapshot acquired then broadcast it. You may additionally allow the user to specify a type map... take a look at how the `types` are handled currently, and you will need to do basically the same thing for the bonds, etc.

clpetix:
@mphoward this is ready for you to give another review. All tests pass on my end, but they're going to fail here until we bump lammpsio to include all the topology conversion work. 

Also, where should validation scripts be placed?

mphoward:
> @mphoward this is ready for you to give another review. All tests pass on my end, but they're going to fail here until we bump lammpsio to include all the topology conversion work.

Thanks! Is `lammpsio` ready for a release, or are there any open issues there?

> Also, where should validation scripts be placed?

If we can add some loose `pytest` checks in the script to automate these, I think in a directory `tests/validation` would be great! We could use a marker to designate them as validation tests:

https://docs.pytest.org/en/stable/example/markers.html

and I think there is a way to skip certain markers by default. If this is cumbersome, we could also do a `skipIf` based on an environment variable.

src/relentless/simulate/lammps.py
```diff
@@ -312,6 +381,20 @@ def _convert_to_data_file(self, sim):
...
+                    or snap.has_dihedrals()
+                    or snap.has_impropers()
+                ):
+                    if sim["engine"]["atom_style"] == "atomic":
+                        raise ValueError(
+                            "Atomic atom style is not compatible with topology data."
+                        )
+                # store topology data
+                if snap.has_bonds():
+                    sim[self]["bonds"] = snap.bonds
```
mphoward, line 397: Make sure `sim[self]["bonds"]` is either assigned to `None` or broadcast on other ranks. Right now it will be undefined. You may also want to make it private as `_bonds` if we don't want to guarantee it will be available as part of the interface forever.

src/relentless/simulate/lammps.py
```diff
@@ -348,6 +431,18 @@ def _convert_to_data_file(self, sim):
...
+                    if snap.has_bonds():
+                        sim[self]["bonds"] = snap.bonds
+                        # get all bond potentials
+                        types = [
+                            sim.potentials.bond.potentials[i].coeff.types
+                            for (i, j) in enumerate(sim.potentials.bond.potentials)
+                        ]
+                        # get all types from all potentials
+                        types = [t for sublist in types for t in sublist]
+                        bond_type_map = {i: types[i - 1] for i in snap.bonds.typeid}
```
mphoward, line 444: This code seems funny to me, and I can't actually figure out what all it's doing. Can you describe what are you trying to achieve here?

src/relentless/simulate/simulate.py
```diff
@@ -583,6 +630,26 @@ def neighbor_buffer(self):
...
+        - ``'1-4'``: Exclude pairs separated by three bonds.
+        """
+        return self._exclusions
+
+    @exclusions.setter
+    def exclusions(self, val):
+        if val is not None:
+            allowed = ["1-2", "1-3", "1-4"]
+            if not all([ex in allowed for ex in val]):
+                raise ValueError("Exclusions must be 'bond', '1-3', or '1-4'")
```
mphoward, line 650:
```suggestion
                raise ValueError("Exclusions must be '1-2', '1-3', or '1-4'")
```

mphoward (changes requested):
Thank you for your hardwork on this! I had a couple minor comments, then this is ready to merge.

src/relentless/simulate/lammps.py
clpetix: I wrote this to make things work but, admittedly, it is not very clear. 

`types` in line 438 is a tuple. So I looped over each tuple (called sublist here) and extracted each type (t) to make a list of all the bond types defined. Then those are used to set the `bond_type_map`. I did it this way to make sure I pulled each type even if there are multiple kinds of bond potentials each potentially with multiple types.

mphoward: Oh OK! If you want to get only the unique types, I would use a set:
```
unique_bond_types = set()
for pot in sim.potentials.bond.potentials:
    for i in pot:
        unique_bond_types.add(i)
```

For the `bond_type_map` line, is that looping over all the particles? If you only need to map the bond types to an integer, then I think you can do do:
```
bond_type_map = {i + 1: t for i, t in enumerate(unique_bond_types)}
```

clpetix: That works well! Thank you!

clpetix:
@mphoward I've made the updates and this should be ready! We are still having the issue with the OpenMPI tests, but everything else seems good!

mphoward (approved):
Thank you very much! This is a nice and much needed contribution. The errors with OpenMPI are related to a missing shared library for an MPI dependency, so I am not going to worry about fixing this now. Let's merge this in and work on the angles & dihedrals next!

Issue #278, opened 2024-11-22
Fix error in WriteTrajectory for dynamic image data in GSD


clpetix:
For HOOMD version 4.0, we have `dynamic.append("particles/images")`. The [proper setting](https://hoomd-blue.readthedocs.io/en/v4.8.2/module-hoomd-write.html#hoomd.write.GSD) is "particles/image".

PR #279, opened 2024-11-22
Fix writing GSD images.

clpetix:
Fixes Issue #278.

### DIFF
--- a/src/relentless/simulate/hoomd.py
+++ b/src/relentless/simulate/hoomd.py
@@ -1620,7 +1620,7 @@ class WriteTrajectory(AnalysisOperation):
             if self.velocities is True:
                 dynamic.append("particles/velocity")
             if self.images is True:
-                dynamic.append("particles/images")
+                dynamic.append("particles/image")
         else:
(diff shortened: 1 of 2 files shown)


mphoward:
Thanks! Can we add some short unit test that checks this?

clpetix:
@mphoward, do you have any insights on why this added tests would cause the HOOMD 2 version to fail?

mphoward:
I am not sure why this would fail in HOOMD 2. It is writing images out that are not just zeros, but they don't seem to be the right ones. Maybe try writing the GSD files with high frequency and watch the movies (HOOMD 2 vs. 4) to see if you spot a difference.

If it is some quirk of HOOMD 2, I move to drop it sooner rather than later.

clpetix:
Thanks! I'll try that and see, but I am in favor of dropping support for HOOMD 2 pretty soon.

clpetix:
@mphoward, dropping HOOMD 2 fixed these failing tests!

mphoward (approved):
Looks good, thanks!

PR #282, opened 2024-12-11
Implement Angle Interactions

clpetix:
Implement harmonic, cosine, cosine squared, and spline angles in LAMMPS and HOOMD.


clpetix:
@mphoward, this should be ready for you to review!

mphoward (changes requested):
This looks great! I had a couple comments mostly related to class names and inheritance, but otherwise this is almost good to go.

src/relentless/simulate/lammps.py
```diff
@@ -393,8 +444,12 @@ def _convert_to_data_file(self, sim):
                         "Atomic atom style is not compatible with topology data."
                     )
                 # store topology data
+                sim[self]["_bonds"] = None
                 if snap.has_bonds():
                     sim[self]["_bonds"] = snap.bonds
+                sim[self]["_angles"] = None
+                if snap.has_angles():
+                    sim[self]["_angles"] = snap.angles
```
clpetix, line 452: While testing the inheritance, I discovered that in LAMMPS we ran into an issue where if `"exclude":True` for `rdf` without having that type of connection in the snapshot, then it would raise an error because the key was never defined. This addresses that problem.

src/relentless/simulate/lammps.py
```diff
@@ -1279,7 +1350,7 @@ def process(self, sim, sim_op):
                     # consider both (i,j) and (j,i) permutations
                     if (
                         sim[self]["_rdf_params"]["exclude"]
-                        and sim[sim.initializer]["_bonds"].N != 0
+                        and sim[sim.initializer]["_bonds"] is not None
```
clpetix, line 1353: This change was also necessary to address the above comment.

src/relentless/model/potential/potential.py
```diff
@@ -490,3 +490,293 @@ def _derivative(self, param, r, **params):
...
+    The spline potential is setup from a tabulated potential instead of
+    specifying knot parameters directly::
+
+        spline = relentless.potential.bond.BondSpline(types=("bondA",), num_knots=3)
+        spline.from_array(("bondA"),[0,1,2],[4,2,0])
+
+    However, the knot variables can be iterated over and manipulated directly::
+
+        for r,k in spline.knots("bondA"):
+            k.value = 1.0
```
clpetix, line 529: How would you hand this example given that this is now the base class?

clpetix:
@mphoward, I added in your suggestions. While testing the inheritance problem pointed out, I found a bug that slipped through the cracks from the bonds PR and addressed that as noted above. 

Aside from a question above, this should be good for you have another look at!

src/relentless/model/potential/potential.py
mphoward: The derived classes actually still need a docstring, so you can put specific examples using the right class in each of those. This class should probably get a note saying that it should not be instantiated directly, rather than an example, or else this example should be generic and use "A" as the type.

Small thing to note above: it should be `potential.BondedSpline`

src/relentless/simulate/lammps.py
mphoward: Makes sense! This is an appropriate fix.

src/relentless/simulate/lammps.py
mphoward: Also agreed.

src/relentless/model/potential/potential.py
```diff
@@ -490,3 +490,293 @@ def _derivative(self, param, r, **params):
 
         """
         pass
+
+
+class BondedSpline(BondedPotential):
```
mphoward, line 495: Don't forget to link this into the documentation.

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,306 @@
...
+            d = 1 + numpy.cos(theta)
+        else:
+            raise ValueError("Unknown parameter")
+
+        if s:
+            d = d.item()
+        return d
+
+
+class AngleSpline(potential.BondedSpline, AnglePotential):
```
mphoward, line 287: Following with the other comment: here we want to give the class a full docstring describing its constructor and examples. The description of the constructor arguments can be basically copy–pasted from the base class, then add specific examples.

I think you then don't really need the methods below, and can just give an empty class body with `pass` because all of these are just calling the parent methods. The class will inherit the docstrings of those parents, which are probably more detailed.

Same comment applies to the `BondSpline`.

clpetix: The methods below change the `r` input to `theta`. Thinking more on this the rs are hardcoded in as parameters as "dr-", so this doesn't eliminate all of them. 

If we just pass, are we fine having just using r in all splines?

mphoward: Oh good point. I think in the base potential tabulator, we use x as a generic coordinate so you could follow the same practice here. I’ll leave it up to you whether you want to just use x for all derived classes, or rename the arguments like you’ve done here.

If you rename the arguments, you can also rename the string name of those parameters by adding a class variable like `_space_coord_name = “x”` in the base class , then override it in the deriving classes to r, theta, etc. You would use this variable to format the parameter names as dr, dtheta, etc.

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,397 @@
...
+            d = d.item()
+        return d
+
+
+class HarmonicCosineAngle(AnglePotential):
+    r"""Harmonic cosine angle potential.
+
+    .. math::
+
+        u(\theta) = \frac{k}{2} (cos(\theta) - cos(\theta_0))^2
```
mphoward, line 124:
```suggestion
        u(\theta) = \frac{k}{2} (\cos \theta - \cos \theta_0)^2
```

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,397 @@
...
+    The angle between particles :math:`i`, :math:`j`, and :math:`k` is defined as:
+
+    .. math::
+
+        \theta_{ijk} = \arccos\left(\frac{\mathbf{r}_{ij} \cdot
+        \mathbf{r}_{jk}}{|\mathbf{r}_{ij}||\mathbf{r}_{jk}|}\right)
+
+    where :math:`\mathbf{r}_{ij}` and :math:`\mathbf{r}_{jk}` are the vectors
+    between particles :math:`i` and :math:`j`, and particles :math:`j` and
+    :math:`k`, respectively.
```
mphoward, line 25:
```suggestion
    from particle *i* to particle *j* and from particle *j* to particle *k*,
    respectively.
```

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,397 @@
...
+            d = d.item()
+        return d
+
+
+class CosineAngle(AnglePotential):
+    r"""Cosine angle potential.
+
+    .. math::
+
+        u(\theta) = k (1 + cos(\theta))
```
mphoward, line 211:
```suggestion
        u(\theta) = k (1 + \cos \theta)
```

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,397 @@
...
+        Returns
+        -------
+        float or numpy.ndarray
+            The force evaluated at ``theta``. The return type is consistent
+            with ``theta``.
+
+        """
+        return super().force(type_=type_, x=theta)
+
+    def _derivative(self, param, r, **params):
```
mphoward, line 395: This should also be `theta`
```suggestion
    def _derivative(self, param, theta, **params):
```
The `derivative` function also needs to have a similar pass-through implementation.

src/relentless/model/potential/potential.py
```diff
@@ -490,3 +490,319 @@ def _derivative(self, param, r, **params):
...
+        differences between neighboring knots are manipulated for all other
+        knots. Defaults to ``'diff'``.
+    name : str
+        Unique name of the potential. Defaults to ``__u[id]``, where ``id`` is the
+        unique integer ID of the potential.
+
+    """
+
+    valid_modes = ("value", "diff")
+    _space_coord_name = "x"
```
mphoward, line 525: I noticed that the parent `BondedPotential` class uses `r` as its space variable for `derivative`. I would go replace that by `x` to make this consistent in our code base.

mphoward (approved):
Looks great! Please address the small comments here, then this is ready to merge.

clpetix:
@mphoward, I think I've fully sorted out making sure that `r` & `theta` are used appropriately. Let me know if you catch anything else!

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,471 @@
...
+        \mathbf{r}_{jk}}{|\mathbf{r}_{ij}||\mathbf{r}_{jk}|}\right)
+
+    where :math:`\mathbf{r}_{ij}` and :math:`\mathbf{r}_{jk}` are the vectors
+    from particle *i* to particle *j* and from particle *j* to particle *k*,
+    respectively.
+
+    """
+
+    def derivative(self, type_, var, theta):
+        r"""Evaluate bond derivative with respect to a variable.
```
mphoward, line 30:
```suggestion
        r"""Evaluate derivative with respect to a variable.
```

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,471 @@
...
+
+        if s:
+            d = d.item()
+        return d
+
+
+class AngleSpline(potential.BondedSpline, AnglePotential):
+    """Spline angle potentials.
+
+    The angleed spline potential is defined by interpolation through a set of
```
mphoward, line 326:
```suggestion
    The angle spline potential is defined by interpolation through a set of
```

src/relentless/model/potential/angle.py
```diff
@@ -0,0 +1,471 @@
...
+        -------
+        float or numpy.ndarray
+            The force evaluated at ``theta``. The return type is consistent
+            with ``theta``.
+
+        """
+        return super().force(type_=type_, x=theta)
+
+    def derivative(self, type_, var, theta):
+        r"""Evaluate bond derivative with respect to a variable.
```
mphoward, line 432:
```suggestion
        r"""Evaluate derivative with respect to a variable.
```

mphoward (approved):
Thanks! I will commit these small edits and merge.

PR #286, opened 2025-01-13
Implement dihedral interactions

clpetix:


clpetix:
@mphoward, this should be ready for you to take a look whenever you get a chance!

src/relentless/model/potential/__init__.py
```diff
@@ -38,6 +38,15 @@
...
 
+Dihedral potentials
+===================
+
+.. autosummary::
+    :toctree: generated/
+
+    OPLSDihedral
+    RyckaertBellemansDihedral
+    DihedralSpline
```
mphoward, line 49: I noticed we are inconsistent between bond, angle, and dihedral in how we are ordering these. Should they all be alphabetical, or logically ordered?

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,482 @@
...
+
+        \mathbf{n}_{A} = \mathbf{r}_{ij} \times \mathbf{r}_{jk}
+
+    and
+
+    .. math::
+
+        \mathbf{n}_{B} = \mathbf{r}_{jk} \times \mathbf{r}_{k\ell}
+
+    where :math:`\mathbf{r}_{ij}`,  :math:`\mathbf{r}_{jk}`, :math:`\mathbf{r}_{jk}`
```
mphoward, line 37:
```suggestion
    where :math:`\mathbf{r}_{ij}`,  :math:`\mathbf{r}_{jk}`, :math:`\mathbf{r}_{k\ell}`
```

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,482 @@
...
+        unique integer ID of the potential.
+
+    Attributes
+    ----------
+    coeff : :class:`DihedralParameters`
+        Parameters of the potential for each type.
+
+    Examples
+    --------
+    OPLS Dihedral::
```
mphoward, line 122: Is this a specific type of dihedral? It might help someone to state which one it represents and include the reference to the paper somewhere in this description.
```suggestion
    OPLS dihedral::
```

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,482 @@
...
+        else:
+            raise ValueError("Unknown parameter")
+
+        if s:
+            d = d.item()
+        return d
+
+
+class RyckaertBellemansDihedral(DihedralPotential):
+    r"""Cosine squared dihedral potential.
```
mphoward, line 195: Fix up the short class description, maybe this?
```suggestion
    r"""Ryckaert–Bellemans dihedral potential.
```

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,482 @@
...
+        unique integer ID of the potential.
+
+    Attributes
+    ----------
+    coeff : :class:`DihedralParameters`
+        Parameters of the potential for each type.
+
+    Examples
+    --------
+    Harmonic Dihedral::
```
mphoward, line 237: Same here, state what this is an example for, possibly with a reference.

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,482 @@
...
+
+        psi = phi - numpy.pi
+
+        u = (
+            c0
+            + c1 * numpy.cos(psi)
+            + c2 * numpy.cos(psi) ** 2
+            + c3 * numpy.cos(psi) ** 3
+            + c4 * numpy.cos(psi) ** 4
+            + c5 * numpy.cos(psi) ** 5
```
mphoward, line 276: Performance isn't a HUGE concern for us, but I would at least store `cos_phi` in a variable and use it to evaluate these powers, so the cosine isn't calculated 6 times.

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,482 @@
...
+
+        for r,k in spline.knots("dihedralA"):
+            k.value = 1.0
+
+    """
+
+    _space_coord_name = "phi"
+
+    def __init__(self, types, num_knots, mode="diff", name=None):
+        super().__init__(types=types, num_knots=num_knots, mode=mode, name=name)
```
mphoward, line 375: I'm not sure if I missed this in prior PRs, but if all of this is directly passed through I think there is no need to write this function out explicitly. (It will inherit.)

src/relentless/simulate/lammps.py
```diff
@@ -260,6 +264,21 @@ def _call_commands(self, sim):
...
+                phiD, uD[i], fD[i] = (
+                    sim.potentials.dihedral.linear_space,
+                    sim.potentials.dihedral.energy(key=i),
+                    sim.potentials.dihedral.force(key=i),
+                )
+                if numpy.any(numpy.isinf(uD[i])):
+                    raise ValueError(
+                        "Dihedral potential/force is infinite at evaluated r"
+                    )
+            NphiD = len(phiD)
```
mphoward, line 280: Do you need to error check that this has at least 2 elements in it, since you are going to drop one later? We do something like this for the pair potentials, I think.

src/relentless/simulate/simulate.py
```diff
@@ -98,12 +98,12 @@ def _cantor_pairing(self, connection, neighbor):
...
-        ) / 2 + connection[:, 1]
-        pi_neighbor = (neighbor[:, 0] + neighbor[:, 1]) * (
-            neighbor[:, 0] + neighbor[:, 1] + 1
-        ) / 2 + neighbor[:, 1]
+        pi_connection = (connection[:, 0] + connection[:, -1]) * (
+            connection[:, 0] + connection[:, -1] + 1
+        ) / 2 + connection[:, -1]
+        pi_neighbor = (neighbor[:, 0] + neighbor[:, -1]) * (
+            neighbor[:, 0] + neighbor[:, -1] + 1
+        ) / 2 + neighbor[:, -1]
```
mphoward, line 106: Since the cantor pairing is really only meant to work on 2-tuples, I would prefer that you fix the code that calls this method to only use the first and last index, rather than change the cantor pairing function. I think when you call it, you could do like:
```
_cantor_pairing(dihedrals[:, (0, -1)], neighbors)
```
to pull the first and last columns of every row.

src/relentless/simulate/simulate.py
```diff
@@ -870,3 +882,28 @@ class AnglePotentialTabulator(PotentialTabulator):
...
+    stop : float
+        The maximum value of :math:`\theta` at which to tabulate.
+    num : int
+        The number of points (value of :math:`\theta`) at which to tabulate and
+        evaluate the potential.
+
+    """
+
+    def __init__(self, potentials, num):
+        super().__init__(potentials=potentials, start=-numpy.pi, stop=numpy.pi, num=num)
```
mphoward, line 909: I wanted to confirm: is this the IUPAC definition of the range of a dihedral angle? You should follow that, and also document the domain of the angle when you define the math of it above. (Didn't scroll back to check, so apologies if it's already there.)

mphoward (changes requested):
This looks great Levi. Here are a few comments to cleanup, then this should be ready to go.

src/relentless/simulate/simulate.py
clpetix: Yes. IUPAC defines the dihedral to have an absolute value less than 180 degrees with the sign depending on how the bonds are rotated. 

https://goldbook.iupac.org/terms/view/T06406

src/relentless/model/potential/__init__.py
clpetix: I reordered these to be alphabetic which probably makes the most sense when searching for something.

clpetix:
@mphoward, I've addressed these comments and this should be ready for you again. Thanks!

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,488 @@
...
+            If any value in ``phi`` is negative.
+        TypeError
+            If the parameter with respect to which to take the derivative
+            is not a :class:`~relentless.variable.Variable`.
+
+        """
+
+        return super().derivative(type_=type_, var=var, x=phi)
+
+    pass
```
mphoward, line 82: Since this is not an empty class, pass is not needed.
```suggestion
```

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,488 @@
...
+
+class OPLSDihedral(DihedralPotential):
+    r"""OPLS dihedral potential.
+
+    .. math::
+
+        u(\phi) = \frac{1}{2} \left( k_1 (1+\cos \phi) + k_2 (1+\cos 2\phi)
+        + k_3 (1+ \cos 3\phi) + k_4 (1+ \cos 4\phi) \right)
+
+    where :math:`\phi` is the dihedral between four bonded particles. The dihedral
```
mphoward, line 93:
```suggestion
    where :math:`\phi` is the dihedral between four bonded particles. The potential
```

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,488 @@
...
+        return f
+
+    def _derivative(self, param, phi, k1, k2, k3, k4):
+        r"""Evaluate dihedral derivative with respect to a variable."""
+        phi, d, s = self._zeros(phi)
+
+        if param == "k1":
+            d = 0.5 * (1 + numpy.cos(phi))
+        elif param == "k2":
+            d = 0.5 * (-1 - numpy.cos(2 * phi))
```
mphoward, line 184:
```suggestion
            d = -0.5 * (1 + numpy.cos(2 * phi))
```

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,488 @@
...
+        elif param == "k3":
+            d = 0.5 * (1 + numpy.cos(3 * phi))
+        elif param == "k4":
+            d = 0.5 * (-1 - numpy.cos(4 * phi))
```
mphoward, line 188:
```suggestion
            d = -0.5 * (1 + numpy.cos(4 * phi))
```

src/relentless/model/potential/dihedral.py
```diff
@@ -0,0 +1,488 @@
...
+class RyckaertBellemansDihedral(DihedralPotential):
+    r"""Ryckaert-Bellemans dihedral potential.
+
+    .. math::
+
+        u(\phi) = c_0 + c_1 (\cos (\phi - \pi)) + c_2 (\cos (\phi - \pi))^2 +
+        c_3 (\cos (\phi - \pi))^3 + c_4 (\cos (\phi - \pi))^4 +
+        c_5 (\cos (\phi - \pi))^5
+
+    where :math:`\phi` is the dihedral between four bonded particles. The dihedral
```
mphoward, line 206:
```suggestion
    where :math:`\phi` is the dihedral between four bonded particles. The potential
```

src/relentless/simulate/simulate.py
```diff
@@ -870,3 +882,28 @@ class AnglePotentialTabulator(PotentialTabulator):
...
+
+    Parameters
+    ----------
+    potentials : :class:`~relentless.potential.dihedral.DihedralPotential` or array_like
+        The dihedral potential(s) to be tabulated. If array_like, all elements must
+        be :class:`~relentless.potential.dihedral.DihedralPotential`\s.
+    start : float
+        The minimum value of :math:`\theta` at which to tabulate.
+    stop : float
+        The maximum value of :math:`\theta` at which to tabulate.
```
mphoward, line 901:
```suggestion
```

src/relentless/simulate/simulate.py
```diff
@@ -870,3 +882,28 @@ class AnglePotentialTabulator(PotentialTabulator):
...
+    num : int
+        The number of points (value of :math:`\theta`) at which to tabulate and
```
mphoward, line 903:
```suggestion
        The number of points (value of :math:`\phi`) at which to tabulate and
```

src/relentless/simulate/simulate.py
```diff
@@ -870,3 +882,28 @@ class AnglePotentialTabulator(PotentialTabulator):
 
     def __init__(self, potentials, num):
         super().__init__(potentials=potentials, start=0.0, stop=numpy.pi, num=num)
+
+
+class DihedralPotentialTabulator(PotentialTabulator):
+    r"""Tabulate one or more dihedral potentials.
+
+    Enables evaluation of energy, force, and derivative at different
+    dihedral values (i.e. :math:`\theta`).
```
mphoward, line 891:
```suggestion
    dihedral values (i.e. :math:`\phi`).
```

src/relentless/simulate/simulate.py
```diff
@@ -870,3 +882,28 @@ class AnglePotentialTabulator(PotentialTabulator):
 
     def __init__(self, potentials, num):
         super().__init__(potentials=potentials, start=0.0, stop=numpy.pi, num=num)
+
+
+class DihedralPotentialTabulator(PotentialTabulator):
```
mphoward, line 887: Not sure if the angle tabulator is also missing this, but it should be noted here that the tabulation runs from -pi to pi, so that someone knows how num relates to the spacing.

mphoward (changes requested):
This looks great! Just a couple small things and then we can merge.

src/relentless/simulate/simulate.py
clpetix: That's my bad! I did document the range on the angle potential, but missed it on this one.

clpetix:
@mphoward I implemented all these changes. Thanks for the help getting all these ready to go!

mphoward (approved):

PR #290, opened 2025-01-21
Fix failing MPI tests

clpetix:


clpetix:
@mphoward the tests still fail with your first set of suggestions but changing it to MPICH let them pass.

mphoward:
Thanks! It must be installing the OpenMPI version, but then some local dependency is not met. It might be possible to circumvent that by installing an MPI library locally before using conda, but I think I had issues with that in the past as well. If that solution works, let's stick with this!

clpetix:
@mphoward, this seems to be working! I don't think we fully cleaned out 29 Sep 2021 version, so I caught a check on RunBrownianDynamics that needed to go as well. This should be ready for your review!

mphoward (approved):

Issue #292, opened 2025-02-03
Add ensemble average calculation method


clpetix:
We should implement the ability to calculate the relative entropy using ensemble averages from Equation 4 of [the relentless paper](https://doi.org/10.1063/5.0233683). 

This method will give a path to designing bonded interactions without use of bonded distributions functions and should be more robust without sensitivity to histogramming in g(r) calculations. Long term we could probably end support for the g(r) method.

mphoward:
This work was completed in #293 and #300.

Issue #184, opened 2023-03-06
Use variable name rather than variable object


mphoward:
In places where we use the variable object as a key, we should consider replacing it with the variable name. This will be helpful for serializing to file and reconstructing objects from file. However, there could be ramifications where places where we use `is` to test for objects, and we need to consider whether that will be less convenient. For example:
```
if x in gradient
```
vs.
```
if x.name in gradient
```
It might be possible to specially define the `__getitem__` method to reinterpret variables to use their name, though, and retain the first behavior.

PR #293, opened 2025-02-03
Implement ensemble average calculation method. 

clpetix:
Addresses Issue #292.

### DIFF
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -368,10 +368,13 @@ class RelativeEntropy(ObjectiveFunction):
 
         # compute gradient and result
         # relative entropy *value* is None
-        if self._use_trajectory(self.target, self.thermo):
-            gradient = self._compute_gradient_direct_average(sim_ens, variables)
-        else:
-            gradient = self.compute_gradient(sim_ens, variables)
+        gradient = None
+        if mpi.world.rank_is_root:
+            if self._use_trajectory(self.target, self.thermo):
+                gradient = self._compute_gradient_direct_average(sim_ens, variables)
+            else:
+                gradient = self.compute_gradient(sim_ens, variables)
+        mpi.world.bcast(gradient)
         result = ObjectiveFunctionResult(
             variables, None, gradient, directory if not directory_is_tmp else None
         )
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -402,12 +402,78 @@ class RelativeEntropy(ObjectiveFunction):
         # load the target ensemble trajectory file
         traj_tgt = self.target
         dvars = variable.graph.check_variables_and_types(variables, variable.Variable)
+
+        with gsd.hoomd.open(traj_tgt, "r") as traj:
+            pair_types_tgt = traj[0].particles.types
+
+            bond_types_tgt = None
+            if traj[0].bonds.N != 0:
+                bond_types_tgt = traj[0].bonds.types
+
+            angle_types_tgt = None
+            if traj[0].angles.N != 0:
+                angle_types_tgt = traj[0].angles.types
+
+            dihedral_types_tgt = None
+            if traj[0].dihedrals.N != 0:
+                dihedral_types_tgt = traj[0].dihedrals.types
+
+            V_tgt = 0
+            for snap in traj:
+                Lx, Ly, Lz, xy, xz, yz = snap.configuration.box
+                V_tgt += extent.TriclinicBox(Lx, Ly, Lz, xy, xz, yz).extent / len(traj)
+
+        with gsd.hoomd.open(sim_traj, "r") as traj:
+            pair_types_sim = traj[0].particles.types
+
+            bond_types_sim = None
+            if traj[0].bonds.N != 0:
+                bond_types_sim = traj[0].bonds.types
+
+            angle_types_sim = None
+            if traj[0].angles.N != 0:
+                angle_types_sim = traj[0].angles.types
+
+            dihedral_types_sim = None
+            if traj[0].dihedrals.N != 0:
+                dihedral_types_sim = traj[0].dihedrals.types
+
+            V_sim = 0
+            for snap in traj:
+                Lx, Ly, Lz, xy, xz, yz = snap.configuration.box
+                V_sim += extent.TriclinicBox(Lx, Ly, Lz, xy, xz, yz).extent / len(traj)
+
+        if pair_types_tgt != pair_types_sim:
+            raise ValueError(
+                "Particle types in target and simulation trajectories do not match."
+            )
+        if bond_types_tgt != bond_types_sim:
+            raise ValueError(
+                "Bond types in target and simulation trajectories do not match."
+            )
+        if angle_types_tgt != angle_types_sim:
+            raise ValueError(
+                "Angle types in target and simulation trajectories do not match."
+            )
+        if dihedral_types_tgt != dihedral_types_sim:
+            raise ValueError(
+                "Dihedral types in target and simulation trajectories do not match."
+            )
+
+        if not numpy.isclose(V_tgt, V_sim):
+            raise ValueError(
+                "Volume of target and simulation trajectories do not match."
+            )
+
+        # normalization to extensive or intensive as specified
+        norm_factor = V_tgt if not self.extensive else 1.0
+
         gradient = math.KeyedArray(keys=dvars)
         for var in dvars:
             gradient[var] = (
                 self._calc_ensemble_average_dvar_dlambda(traj_tgt, dvars)[var]
                 - self._calc_ensemble_average_dvar_dlambda(sim_traj, dvars)[var]
-            )
+            ) / norm_factor
         return gradient
 
     def _calc_ensemble_average_dvar_dlambda(self, trajectory, variables):
@@ -487,12 +553,16 @@ class RelativeEntropy(ObjectiveFunction):
 
                 for i in snap.particles.types:
                     for j in snap.particles.types:
+                        filter_j_gt_i = neighbors[:, 1] > neighbors[:, 0]
+                        neighbors.filter(filter_j_gt_i)
+
                         filter_ij = numpy.logical_and(
                             type_masks[i][neighbors[:, 0]],
                             type_masks[j][neighbors[:, 1]],
                         )
 
                     neighbors.filter(filter_ij)
+
                     distances = neighbors.distances[filter_ij]
 
                     for var in variables:
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -585,13 +585,13 @@ class RelativeEntropy(ObjectiveFunction):
                         pos_2 = pos[bonds[:, 1]]
 
                         # Calculate distances
-                        distances = numpy.linalg.norm(pos_2 - pos_1, axis=1)
+                        dr = numpy.linalg.norm(pos_2 - pos_1, axis=1)
 
                         for var in variables:
                             rs = self.potentials.bond.linear_space
                             dus = self.potentials.bond.derivative(i, var)
                             dudvar = math.AkimaSpline(rs, dus)
-                            gradient[var] += numpy.sum(dudvar(distances)) / len(traj)
+                            gradient[var] += numpy.sum(dudvar(dr)) / len(traj)
 
                 # angle contributions to the gradient
                 if snap.angles.N != 0:
@@ -609,7 +609,7 @@ class RelativeEntropy(ObjectiveFunction):
                         r_12 = pos_2 - pos_1
                         r_23 = pos_3 - pos_2
 
-                        angles = numpy.arccos(
+                        dtheta = numpy.arccos(
                             numpy.dot(r_12, r_23)
                             / (numpy.linalg.norm(r_12) * numpy.linalg.norm(r_23))
                         )
@@ -618,7 +618,7 @@ class RelativeEntropy(ObjectiveFunction):
                             rs = self.potentials.angle.linear_space
                             dus = self.potentials.angle.derivative(i, var)
                             dudvar = math.AkimaSpline(rs, dus)
-                            gradient[var] += numpy.sum(dudvar(angles)) / len(traj)
+                            gradient[var] += numpy.sum(dudvar(dtheta)) / len(traj)
 
                 # dihedral contributions to the gradient
                 if snap.dihedrals.N != 0:
@@ -641,7 +641,7 @@ class RelativeEntropy(ObjectiveFunction):
                         cross_12_23 = numpy.cross(r_12, r_23)
                         cross_23_34 = numpy.cross(r_23, r_34)
 
-                        dihedrals = numpy.arccos(
+                        dphi = numpy.arccos(
                             numpy.dot(cross_12_23, cross_23_34)
                             / (
                                 numpy.linalg.norm(cross_12_23)
@@ -653,7 +653,7 @@ class RelativeEntropy(ObjectiveFunction):
                             rs = self.potentials.dihedral.linear_space
                             dus = self.potentials.dihedral.derivative(i, var)
                             dudvar = math.AkimaSpline(rs, dus)
-                            gradient[var] += numpy.sum(dudvar(dihedrals)) / len(traj)
+                            gradient[var] += numpy.sum(dudvar(dphi)) / len(traj)
         return gradient
 
     def compute_gradient(self, ensemble, variables):
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -374,7 +374,7 @@ class RelativeEntropy(ObjectiveFunction):
                 gradient = self._compute_gradient_direct_average(sim_ens, variables)
             else:
                 gradient = self.compute_gradient(sim_ens, variables)
-        mpi.world.bcast(gradient)
+        gradient = mpi.world.bcast(gradient)
         result = ObjectiveFunctionResult(
             variables, None, gradient, directory if not directory_is_tmp else None
         )
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -645,6 +645,7 @@ class RelativeEntropy(ObjectiveFunction):
                             dphi[dphi < -numpy.pi],
                         )
                         print(f"objective.py phi before tabulator: {dphi[0]:.20}")
+                        print(f"objective.py pi: {numpy.pi:.20}")
                         print("objective.py: gt ", numpy.greater(dphi[0], numpy.pi))
                         print("objective.py: lt", numpy.less(dphi[0], -numpy.pi))
                         for var in variables:
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -618,6 +618,7 @@ class RelativeEntropy(ObjectiveFunction):
                                 * numpy.linalg.norm(cross_23_34)
                             )
                         )
+                        print(dphi)
                         for var in variables:
                             gradient[var] += numpy.sum(
                                 self.potentials.dihedral.derivative(i, var, x=dphi)
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -545,11 +545,11 @@ class RelativeEntropy(ObjectiveFunction):
                         bonds = snap.bonds.group[bond_type_filter]
 
                         # Get positions for all pairs of bonded particles
-                        pos_1 = box.wrap(pos[bonds[:, 0]])
-                        pos_2 = box.wrap(pos[bonds[:, 1]])
+                        pos_1 = pos[bonds[:, 0]]
+                        pos_2 = pos[bonds[:, 1]]
 
                         # Calculate distances
-                        dr = numpy.linalg.norm(pos_2 - pos_1, axis=1)
+                        dr = numpy.linalg.norm(box.wrap(pos_2 - pos_1), axis=1)
 
                         for var in variables:
                             gradient[var] += numpy.sum(
@@ -567,13 +567,13 @@ class RelativeEntropy(ObjectiveFunction):
                         angles = snap.angles.group[angle_type_filter]
 
                         # get positions for all triplets of bonded particles
-                        pos_1 = box.wrap(pos[angles[:, 0]])
-                        pos_2 = box.wrap(pos[angles[:, 1]])
-                        pos_3 = box.wrap(pos[angles[:, 2]])
+                        pos_1 = pos[angles[:, 0]]
+                        pos_2 = pos[angles[:, 1]]
+                        pos_3 = pos[angles[:, 2]]
 
                         # calculate angles
-                        r_12 = pos_2 - pos_1
-                        r_23 = pos_3 - pos_2
+                        r_12 = box.wrap(pos_2 - pos_1)
+                        r_23 = box.wrap(pos_3 - pos_2)
 
                         # use einsum for row-wise dot product
                         dtheta = numpy.arccos(
@@ -599,15 +599,15 @@ class RelativeEntropy(ObjectiveFunction):
                         dihedrals = snap.dihedrals.group[dihedral_type_filter]
 
                         # get positions for all quadruplets of bonded particles
-                        pos_1 = box.wrap(pos[dihedrals[:, 0]])
-                        pos_2 = box.wrap(pos[dihedrals[:, 1]])
-                        pos_3 = box.wrap(pos[dihedrals[:, 2]])
-                        pos_4 = box.wrap(pos[dihedrals[:, 3]])
+                        pos_1 = pos[dihedrals[:, 0]]
+                        pos_2 = pos[dihedrals[:, 1]]
+                        pos_3 = pos[dihedrals[:, 2]]
+                        pos_4 = pos[dihedrals[:, 3]]
 
                         # calculate dihedrals
-                        r_12 = pos_2 - pos_1
-                        r_23 = pos_3 - pos_2
-                        r_34 = pos_4 - pos_3
+                        r_12 = box.wrap(pos_2 - pos_1)
+                        r_23 = box.wrap(pos_3 - pos_2)
+                        r_34 = box.wrap(pos_4 - pos_3)
 
                         cross_12_23 = numpy.cross(r_12, r_23)
                         cross_23_34 = numpy.cross(r_23, r_34)
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -632,7 +632,7 @@ class RelativeEntropy(ObjectiveFunction):
 
                         # wrap dihedral angles to [-pi, pi]
                         print(
-                            "objetive.py trigger before",
+                            "objetive.py out of range before wrapping: ",
                             dphi[dphi > numpy.pi],
                             dphi[dphi < -numpy.pi],
                         )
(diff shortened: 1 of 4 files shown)


src/relentless/optimize/objective.py
```diff
@@ -322,10 +323,13 @@ def compute(self, variables, directory=None):
             The result, which has unknown value ``None`` and known gradient.
 
         """
-        if self.thermo.rdf is None:
-            raise ValueError(
-                "EnsembleAverage needs to compute RDF, specify parameters."
-            )
+        if isinstance(self.target, Ensemble) and isinstance(
+            self.thermo, EnsembleAverage
+        ):
```
clpetix, line 328: Do you think this needs a hidden attribute to reduce duplication? It's used more than once but not a ton, so I wasn't sure.

src/relentless/optimize/objective.py
```diff
@@ -353,7 +357,12 @@ def compute(self, variables, directory=None):
         # run simulation and use result to compute gradient
         try:
             sim = self.simulation.run(self.potentials, directory)
-            sim_ens = sim[self.thermo]["ensemble"]
+            if isinstance(self.target, Ensemble) and isinstance(
+                self.thermo, EnsembleAverage
+            ):
+                sim_ens = sim[self.thermo]["ensemble"]
+            else:
+                sim_ens = sim[self.thermo]["_gsd"].filename
```
clpetix, line 365: This lets us very easily pull the current iterations gsd file for analysis! So this should be pretty straightforward to implement!

src/relentless/optimize/objective.py
```diff
@@ -375,6 +384,65 @@ def compute(self, variables, directory=None):
...
+        traj_sim, traj_tgt = 0, 0
+        update = traj_sim + traj_tgt
+
+        for var in dvars:
+            update = 0
+            types = collections.PairMatrix(
+                keys=self.potentials.pair.potentials[0].coeff.types
+            )
+            for i, j in types:
+                gradient[var] = update
```
clpetix, line 442: Here I think we should pull the types from the loaded files and make a pair matrix and keyed array for all the bonded types then loop through evaluating. Probably also need a check to make sure the the types are the same between the two files. 

Let me know if you think there is a better way to implement this.

src/relentless/optimize/objective.py
```diff
@@ -375,6 +384,65 @@ def compute(self, variables, directory=None):
...
+        else:
+            return self._compute_gradient_new(ensemble, variables)
+
+    def _compute_gradient_new(self, ensemble, variables):
+        r"""Computes the relative entropy gradient for an ensemble.
+
+        Parameters
+        ----------
+        ensemble : :class:`string`
+            Path to gsd trajectory file.
```
clpetix, line 415: We talked about this, but I just wanted to confirm we're good to just require this use GSD since it should contain all the type information. I don't think that is a problem since we interconvert for users.

clpetix:
@mphoward, I think I have a very very barebones implementation of isolating the new and old method and getting the necessary data to calculate with the new method. Could you please just give the overall structure a look and my comments / questions about implementation? I'm trying to plan this out a bit more to hopefully avoid any major revision in code review!

src/relentless/optimize/objective.py
mphoward: Yes, I think that it would make sense to extract the types from the trajectory files for all interactions (pairs, bonds, etc.). Line 439 above will have an issue if you don't do this because it's possible for different potentials to be keyed on different types.

src/relentless/optimize/objective.py
mphoward: Yes, I think it's OK to say (at least for now) that it has to be a GSD file.

clpetix:
@mphoward, could you take a brief look at this? It seems to be working with topology and in agreement with the old method on pair calculations! 

Next steps are adding in UnitTests and I think it would be a good idea to try to estimate the uncertainty between the two methods using a simple test case.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +383,286 @@ def compute(self, variables, directory=None):
...
+
+                    neighbors.filter(filter_ij)
+
+                    distances = neighbors.distances[filter_ij]
+
+                    for var in variables:
+                        rs = self.potentials.pair.linear_space
+                        dus = self.potentials.pair.derivative((i, j), var)
+                        dudvar = math.AkimaSpline(rs, dus)
+                        gradient[var] += numpy.sum(dudvar(distances)) / len(traj)
```
mphoward, line 580: Can't this be done by specifying x directly, rather than passing through an intermediate interpolation?
```suggestion
                        gradient[var] += numpy.sum(self.potentials.pair.derivative((i, j), var, x=distances))
```
Side note: I would normalize the whole thing by the number of frames at the end

Same comments apply below.

src/relentless/optimize/objective.py
```diff
@@ -492,6 +781,44 @@ def target(self):
 
     @target.setter
     def target(self, value):
-        if value.V is None or value.N is None:
-            raise ValueError("The target ensemble must have both V and N set.")
+        if isinstance(value, Ensemble):
+            if value.V is None or value.N is None:
+                raise ValueError("The target ensemble must have both V and N set.")
```
mphoward, line 786: Consider adding an error check that the target is a GSD file, if that is a hard requirement?

src/relentless/optimize/objective.py
```diff
@@ -353,15 +352,25 @@ def compute(self, variables, directory=None):
         # run simulation and use result to compute gradient
         try:
             sim = self.simulation.run(self.potentials, directory)
-            sim_ens = sim[self.thermo]["ensemble"]
+            if self._use_ensemble_average(self.target, self.thermo):
+                sim_ens = sim.directory.file(self.thermo.filename)
```
mphoward, line 356: Error check needed somewhere to ensure this is a GSD file? Otherwise I think the calculation method will choke when it tries to open it.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +383,286 @@ def compute(self, variables, directory=None):
...
+                    raise ValueError(
+                        "Temperature not provided and cannot be calculated "
+                        "from trajectory."
+                    )
+                T_avg = 0
+                for snap in traj:
+                    T_avg += numpy.sum(
+                        numpy.linalg.norm(snap.particles.velocity, axis=1) ** 2
+                        / (3 * snap.particles.mass * snap.particles.N)
+                    ) / len(traj)
```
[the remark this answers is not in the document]
clpetix: You're right! I flubbed p^2 / m --> mv^2 :sweat_smile:

src/relentless/optimize/objective.py
```diff
@@ -374,6 +383,286 @@ def compute(self, variables, directory=None):
...
+        :class:`~relentless.math.KeyedArray`
+            The gradient, keyed on the ``variables``.
+
+        """
+
+        # load the target ensemble trajectory file
+        traj_tgt = self.target
+        dvars = variable.graph.check_variables_and_types(variables, variable.Variable)
+
+        with gsd.hoomd.open(traj_tgt, "r") as traj:
```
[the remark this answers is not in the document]
clpetix: I didn't see where we had that in. How would you recommend implementing that? My first attempt broke the tests haha!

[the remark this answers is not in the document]
mphoward: The closest example would be in the LAMMPS ensemble analyzer, since that opens a trajectory to compute the RDF. In HOOMD, the equivalent in the ensemble analyzer would be only accessing the snapshot on the root rank.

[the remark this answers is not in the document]
clpetix: Since this will go in the new method, I'm going to leave this unresolved until I have unit tests in then come back and do the MPI safeguard.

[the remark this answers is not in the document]
mphoward: Sounds good, let me know when you are ready to do so! It might be possible to just put this check outside where we call the method itself and broadcast the gradient it returns.

clpetix:
@mphoward, I have the unittests with no exclusions but freud is throwing an error about using the unwrap function. Do you have any ideas what this may be? The tests are passing on my local machine.

clpetix:
@mphoward, this should be ready for you to review again now that #297 is merged in! The base tests are all in and passing. 

I've also applied your recommendation from above of broadcasting the gradient for MPI and that seems to have worked on MPICH, but is causing issues in openPMI. Could you please let me know if you have any ideas on what is tripping it up?

clpetix:
@mphoward, I tried some things and then reverted and now MPICH is also failing.

src/relentless/optimize/objective.py
```diff
@@ -353,15 +353,29 @@ def compute(self, variables, directory=None):
...
         # relative entropy *value* is None
-        gradient = self.compute_gradient(sim_ens, variables)
+        gradient = None
+        if mpi.world.rank_is_root:
+            if self._use_trajectory(self.target, self.thermo):
+                gradient = self._compute_gradient_direct_average(sim_ens, variables)
+            else:
+                gradient = self.compute_gradient(sim_ens, variables)
+        mpi.world.barrier()
+        gradient = mpi.world.bcast(gradient)
```
mphoward, line 378: I think the problem is that you're broadcasting a dict that uses variable objects (not their names) as keys and so are actually different on different ranks. (A poor design choice we made long ago... #184).

It's a bit clunky, but what you can do is turn the `gradient` object into a list, broadcast the list, then reassign the list to the right keys.
```suggestion
        gradient_list = None
        if mpi.world.rank_is_root:
            if self._use_trajectory(self.target, self.thermo):
                gradient = self._compute_gradient_direct_average(sim_ens, variables)
            else:
                gradient = self.compute_gradient(sim_ens, variables)
            gradient_list = [gradient[var] for var in variables]
        gradient_list = mpi.world.bcast(gradient_list)
        gradient = {var: gradient_list[i] for i, var in enumerate(variables)}
```

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+                        "from trajectory."
+                    )
+                T_avg = 0
+                for snap in traj:
+                    T_avg += numpy.sum(
+                        snap.particles.mass
+                        * numpy.linalg.norm(snap.particles.velocity, axis=1) ** 2
+                        / (3 * snap.particles.N * self.potentials.kB)
+                    ) / len(traj)
+                self.T = T_avg
```
mphoward, line 462: Unless I'm missing something, the calculation of `V_tgt` and of `T` are the only parts of these two code blocks that we need?

Also, you should use `T_avg` directly (rather than storing in `self.T` here) because `self.T` is what the user specified. Put an else branch to assign `T_avg = self.T` if it is set.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+        return gradient
+
+    def _calc_ensemble_average_dvar_dlambda(self, trajectory, variables):
+        gradient = math.KeyedArray(keys=variables)
+        for var in variables:
+            gradient[var] = 0
+
+        with gsd.hoomd.open(trajectory, "r") as traj:
+            # loop through the trajectory and calculate ensemble average
+            box = freud.box.Box.from_box(traj[0].configuration.box)
```
mphoward, line 496: You will need to move the box inside the loop, it will change if it is NPT.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+                for i in snap.particles.types:
+                    for j in snap.particles.types:
+                        filter_ij = numpy.logical_and(
+                            type_masks[i][neighbors[:, 0]],
+                            type_masks[j][neighbors[:, 1]],
+                        )
+                        for var in variables:
+                            gradient[var] += numpy.sum(
+                                self.potentials.pair.derivative(
+                                    (i, j), var, x=neighbors.distances[filter_ij]
```
mphoward, line 575: The distance freud reports respect the PBCs, right?

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+                                )
+                            ) / len(traj)
```
mphoward, line 577: It will be very easy to forget to do this normalization for each contribution. How about you just wait until after the loop over the trajectory is complete and divide once?

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+                        bond_type_filter = bond_type_map[i] == snap.bonds.typeid
+
+                        bonds = snap.bonds.group[bond_type_filter]
+
+                        # Get positions for all pairs of bonded particles
+                        pos_1 = pos[bonds[:, 0]]
+                        pos_2 = pos[bonds[:, 1]]
+
+                        # Calculate distances
+                        dr = numpy.linalg.norm(pos_2 - pos_1, axis=1)
```
mphoward, line 592: I think you need to apply PBCs before you do this calculation. That would apply for all vectors in bonded interactions.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+
+                        for var in variables:
+                            rs = self.potentials.bond.linear_space
+                            dus = self.potentials.bond.derivative(i, var)
+                            dudvar = math.AkimaSpline(rs, dus)
+                            gradient[var] += numpy.sum(dudvar(dr)) / len(traj)
```
mphoward, line 598: This should be evaluated directly using the derivative method of the tabulated bond potential, not interpolation. (Like you did for the pairs above)

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+
+                # angle contributions to the gradient
```
mphoward, line 600: Same comments on PBC and how to evaluate derivative as for bond.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+                            / (numpy.linalg.norm(r_12) * numpy.linalg.norm(r_23))
+                        )
+
+                        for var in variables:
+                            rs = self.potentials.angle.linear_space
+                            dus = self.potentials.angle.derivative(i, var)
+                            dudvar = math.AkimaSpline(rs, dus)
+                            gradient[var] += numpy.sum(dudvar(dtheta)) / len(traj)
+
+                # dihedral contributions to the gradient
```
mphoward, line 631: Same comments on PBC and how to evaluate derivative as for bond.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+
+                        # calculate dihedrals
+                        r_12 = pos_2 - pos_1
+                        r_23 = pos_3 - pos_2
+                        r_34 = pos_4 - pos_3
+
+                        cross_12_23 = numpy.cross(r_12, r_23)
+                        cross_23_34 = numpy.cross(r_23, r_34)
+                        dphi = numpy.arccos(
+                            numpy.einsum("ij,ij->i", cross_12_23, cross_23_34)
```
mphoward, line 657: Same comment about how to compute dot product.

src/relentless/optimize/objective.py
```diff
@@ -374,6 +389,284 @@ def compute(self, variables, directory=None):
...
+                "Bond types in target and simulation trajectories do not match."
+            )
+        if angle_types_tgt != angle_types_sim:
+            raise ValueError(
+                "Angle types in target and simulation trajectories do not match."
+            )
+        if dihedral_types_tgt != dihedral_types_sim:
+            raise ValueError(
+                "Dihedral types in target and simulation trajectories do not match."
+            )
```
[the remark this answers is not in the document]
clpetix: I think we discussed that we should require that the user provide a target file that matches their simulation file in terms of types. Should I go ahead and remove this requirement?

src/relentless/optimize/objective.py
clpetix: Yes

src/relentless/optimize/objective.py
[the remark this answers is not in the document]
clpetix: I think we discussed in the context of Tyla's coarse-graining. Requiring the user to clean up the fine-grained file to match the types to what will be output from relentless. 

I'm good with whatever you prefer here though!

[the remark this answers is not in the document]
mphoward: Based on how the rest of the code is written to select types, I can't see much drawback from using the types that are supposed to be there for purposes of computing the gradient, which are the ones that are in the potential objects (neither the target nor simulated trajectory). It would be reasonable to put it on the user not to supply garbage for either.

Do the potential tabulators currently have a mechanism to give the set of types its child potentials contain? If not, I would favor adding a property that does that, then using those types in this method rather than checking the trajectories.

mphoward:
Yes, I think this would be appropriate. Alternatively, you could take the angle modulo 2 pi since you know the dihedral must be periodic (something a little bigger than pi will be wrapped to something a little bigger than -pi).

src/relentless/optimize/objective.py
clpetix: We would need to unwrap these coordinates, right? Then there should be a requirement that we have the images in the trajectory?

mphoward: No, you shouldn't need to *unwrap* any coordinates for these calculations. All these interactions are computed under the minimum image convention, so you will need to *wrap* them.

I would wrap the particle positions when you load them so you know they lie in the freud box, then I would wrap the difference vectors to apply the minimum image convention.

src/relentless/optimize/objective.py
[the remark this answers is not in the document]
clpetix: I added in the property of types to the tabulators and used that in the ensemble average.

clpetix:
@mphoward, I've made these suggested changes. This should ready for you to take a look at again!

src/relentless/optimize/objective.py
```diff
@@ -374,6 +392,242 @@ def compute(self, variables, directory=None):
...
+                        pos_3 = pos[angles[:, 2]]
+
+                        # calculate angles
+                        r_12 = box.wrap(pos_2 - pos_1)
+                        r_23 = box.wrap(pos_3 - pos_2)
+
+                        # use einsum for row-wise dot product
+                        dtheta = numpy.arccos(
+                            -numpy.sum(r_12 * r_23, axis=1)
+                            / (numpy.linalg.norm(r_12) * numpy.linalg.norm(r_23))
```
mphoward, line 583: Do these norms need to be across `axis=1`?

src/relentless/optimize/objective.py
```diff
@@ -374,6 +392,242 @@ def compute(self, variables, directory=None):
...
+                        r_23 = box.wrap(pos_3 - pos_2)
+                        r_34 = box.wrap(pos_4 - pos_3)
+
+                        cross_12_23 = numpy.cross(r_12, r_23)
+                        cross_23_34 = numpy.cross(r_23, r_34)
+                        dphi = numpy.arccos(
+                            -numpy.sum(cross_12_23 * cross_23_34, axis=1)
+                            / (
+                                numpy.linalg.norm(cross_12_23)
+                                * numpy.linalg.norm(cross_23_34)
```
mphoward, line 620: Same question here

tests/optimize/test_objective.py
```diff
@@ -266,6 +281,581 @@ def test_directory(self):
             y["rdf"]["('1', '1')"], sim_ens.rdf["1", "1"].table.tolist()
         )
 
+
+class test_RelativeEntropyDirectAverage(unittest.TestCase):
```
mphoward, line 285: These look quite comprehensive! Only thing I noticed (sorry if I missed it): is there anything to check the calculation of the temperature when it is not given?

mphoward (commented):
Overall, this looks great! Let's make sure we also update the `RelativeEntropy` documentation to explain the two options and which one we recommend (the new version).

tests/optimize/test_objective.py
clpetix: Oops, good catch! I've added in a test that should cover this and the case where temp cannot be computed

clpetix:
@mphoward, I've addressed all the comments and expanded the testing. This should be ready for you to take another look at!

src/relentless/optimize/objective.py
```diff
@@ -288,12 +292,13 @@ class RelativeEntropy(ObjectiveFunction):
 
     """
 
-    def __init__(self, target, simulation, potentials, thermo, extensive=False):
+    def __init__(self, target, simulation, potentials, thermo, extensive=False, T=None):
```
mphoward, line 295: The docstring for this class needs to be updated to reflect the new arguments that can be accepted, and to explain the differences in calculations (and what is supported or not).

src/relentless/optimize/objective.py
```diff
@@ -374,6 +392,262 @@ def compute(self, variables, directory=None):
...
+
+                        bonds = snap.bonds.group[bond_type_filter]
+
+                        # Get positions for all pairs of bonded particles
+                        pos_1 = pos[bonds[:, 0]]
+                        pos_2 = pos[bonds[:, 1]]
+
+                        # Calculate distances
+                        dr = numpy.linalg.norm(
+                            numpy.asarray(box.wrap((pos_2 - pos_1)), dtype=float),
```
mphoward, line 564:
```suggestion
                            numpy.asarray(box.wrap(pos_2 - pos_1), dtype=float),
```

src/relentless/optimize/objective.py
```diff
@@ -353,15 +359,27 @@ def compute(self, variables, directory=None):
...
         # compute gradient and result
         # relative entropy *value* is None
-        gradient = self.compute_gradient(sim_ens, variables)
+        if mpi.world.rank_is_root:
+            if self._use_trajectory(self.target, self.thermo):
+                gradient = self._compute_gradient_direct_average(
+                    sim.directory.file(self.thermo.filename), variables
+                )
+            else:
+                gradient = self.compute_gradient(
```
mphoward, line 375: Since this method will not work properly if there are design variables in the bonded potentials, should we add a couple lines in that method that compute the derivatives of any bonded potentials with respect to the design variables, and yell if they aren't all zero?

Additionally, I think we are not doing neighbor list exclusions in the old version, right? I'm good with that, but then we should yell if there are exclusions turned on in the nonbonded potentials because the RDF used to do the calculations will be incorrect.

That will allow us to still support special cases where there are bonded potentials and you are only designing the simplest nonbonded interactions. We can document that any other cases should use the new approach.

You should probably add some small unit tests that confirm these exceptions are raised.

mphoward (changes requested):
This looks great! There is one code suggestion to help prevent user error, and otherwise, this just needs documentation before merging.

src/relentless/optimize/objective.py
clpetix: > Since this method will not work properly if there are design variables in the bonded potentials, should we add a couple lines in that method that compute the derivatives of any bonded potentials with respect to the design variables, and yell if they aren't all zero?

I agree with this. 

> Additionally, I think we are not doing neighbor list exclusions in the old version, right? I'm good with that, but then we should yell if there are exclusions turned on in the nonbonded potentials because the RDF used to do the calculations will be incorrect.

I think this one is trickier because we do support calculating the RDF with exclusions, so you can pseudo implement bonded exclusions like we did with Tyla but it's error-prone. Should this be a UserWarning that this method is not well suited to designs with bonded exclusions and a suggestion to use the other method?

mphoward: > I think this one is trickier because we do support calculating the RDF with exclusions, so you can pseudo implement bonded exclusions like we did with Tyla but it's error-prone.

Oh, you are right! I would document that this is possible in the docstring, but we recommend using the new method.

clpetix:
@mphoward, these mpi tests are failing, but I can't see an obvious reason. Do you think this a case of the MPI causing deadlock and not actually triggering the error?

src/relentless/optimize/objective.py
```diff
@@ -246,24 +250,56 @@ class RelativeEntropy(ObjectiveFunction):
...
+        \left\langle  \frac{\partial U}{\partial \mathbf{x}} \right\rangle
+        \right)
+
+    where :math:`\langle\cdot\rangle_0` and :math:`\langle\cdot\rangle` are ensemble
+    averages in the target and model ensembles, respectively,
+    :math:`\beta=1/(k_{\rm B}T)`, and :math:`U` is the total potential energy.
+
+    It is recommended to use the direct ensemble average method in most cases,
+    as it allows for design of :class:`~relentless.potential.pair.PairPotential` and
+    bonded interactions. This method also easily supports design of pair-potentials
```
mphoward, line 274:
```suggestion
    bonded interactions. This method also easily supports design of pair potentials
```

src/relentless/optimize/objective.py
```diff
@@ -246,24 +250,56 @@ class RelativeEntropy(ObjectiveFunction):
...
-    Currently, the :class:`RelativeEntropy` objective function supports only
-    :class:`~relentless.potential.pair.PairPotential` interactions. These interactions
-    are characterized by :math:`g_{ij}(r)`, an :class:`~relentless.ensemble.RDF`
-    for each pair of interacting types :math:`(i,j)` in each
-    :class:`~relentless.ensemble.Ensemble`. The gradient of :math:`S_{\rm rel}` is then:
+
+    The gradient of the relative entropy can be computed through direct ensemble
+    averaging or through integration of the radial distribution function (rdf)
+    of the target and model ensembles. When ``target`` is a :class:`str` and
+    ``thermo`` is a :class:`~relentless.simulate.analyze.WriteTrajectory`
```
mphoward, line 257: We should try to add some suborganization on this. I think you can do it like:

```suggestion
    of the target and model ensembles.
    
    .. rubric:: Direct ensemble average
    
    When ``target`` is a :class:`str` and ``thermo`` is a
    :class:`~relentless.simulate.analyze.WriteTrajectory`
```

src/relentless/optimize/objective.py
```diff
@@ -246,24 +250,56 @@ class RelativeEntropy(ObjectiveFunction):
...
+    where :math:`\langle\cdot\rangle_0` and :math:`\langle\cdot\rangle` are ensemble
+    averages in the target and model ensembles, respectively,
+    :math:`\beta=1/(k_{\rm B}T)`, and :math:`U` is the total potential energy.
+
+    It is recommended to use the direct ensemble average method in most cases,
+    as it allows for design of :class:`~relentless.potential.pair.PairPotential` and
+    bonded interactions. This method also easily supports design of pair-potentials
+    with bonded exclusions.
+
+    The :class:`RelativeEntropy` objective function supports a method for designing
```
mphoward, line 277:
```suggestion

    .. rubric:: Radial distribution function

    The :class:`RelativeEntropy` objective function supports a method for designing
```

src/relentless/optimize/objective.py
```diff
@@ -272,28 +308,32 @@ class RelativeEntropy(ObjectiveFunction):
...
-    thermo : :class:`~relentless.simulate.simulate.SimulationOperation`
-        The thermodynamic analyzer operation for the simulation ensemble and rdf
-        (usually :class:`~relentless.simulate.simulate.AddEnsembleAnalyzer`).
+    thermo : :class:`~relentless.simulate.analyze.EnsembleAverage` or
+            :class:`~relentless.simulate.analyze.WriteTrajectory`
+        The thermodynamic analyzer operation for the simulation ensemble.
         The model ensemble will be extracted from this operation.
     extensive : bool
         Specification of whether the relative entropy is extensive (defaults to
         ``False``).
```
mphoward, line 326: T argument needs to be added and documented

clpetix:
@mphoward, I've addressed these small documentation changes! This should be good to go!

mphoward:
Hooray!!

Issue #270, opened 2024-09-13
Support Python 3.13


mphoward:
This will be released October 1, 2024 and needs to be tested.

mphoward:
Resolved in #294

PR #294, opened 2025-02-10
Add support for Python 3.13.

clpetix:
Addresses Issue #270.


clpetix:
@mphoward, this looks to be working! It looks like no version of LAMMPS currently supports Python 3.13. I put it in the matrix and excluded them so that when one releases that does it will be tested.

mphoward:
Thank you!

PR #297, opened 2025-03-18
Add coordinate validation to potential classes.

clpetix:
Adds validation of the coordinate to to ensure it respects the proper bounds to all bonded potential classes. This check was already done in the pair potential class.


mphoward:
@clpetix ping me when this is ready for review after our discussion!

clpetix:
@mphoward, this should be good! Just making sure the tests pass!

src/relentless/model/potential/potential.py
```diff
@@ -462,6 +467,10 @@ def derivative(self, type_, var, x):
             deriv = deriv.item()
         return deriv
 
+    def _validate_coordinate(self, x):
+        """Validate the coordinate ``x`` is within the valid range."""
+        pass
+
```
mphoward, line 473: I don't think you want to have this here, since it is making an empty implementation of the method (allowing children not to override it).
```suggestion
```

src/relentless/model/potential/angle.py
```diff
@@ -61,9 +61,17 @@ def derivative(self, type_, var, theta):
             is not a :class:`~relentless.variable.Variable`.
 
         """
+        # Validate theta
+        theta = self._validate_coordinate(theta)
```
mphoward, line 66: I noticed this is repeated in all of these children. I think you can just call `self._validate_coordinate` at the start of `BondedPotential.derivative`, and it will use the right one for the child class you're working with.
```suggestion
```

src/relentless/model/potential/angle.py
```diff
@@ -61,9 +61,17 @@ def derivative(self, type_, var, theta):
...
 
         return super().derivative(type_=type_, var=var, x=theta)
 
+    def _validate_coordinate(self, theta):
+        """Validate the angle ``theta`` is between 0 and pi."""
```
mphoward, line 70: Please provide a complete docstring.

src/relentless/model/potential/angle.py
```diff
@@ -116,6 +124,9 @@ def energy(self, type_, theta):
 
         theta, u, s = self._zeros(theta)
 
+        # Validate theta
+        theta = self._validate_coordinate(theta)
```
mphoward, line 128: It isn't necessary to catch the return value, if you don't want. Also, Python recommends against commenting things that are easily understood from the code
```suggestion
        self._validate_coordinate(theta)
```

src/relentless/model/potential/bond.py
```diff
@@ -47,9 +47,17 @@ def derivative(self, type_, var, r):
             is not a :class:`~relentless.variable.Variable`.
 
         """
+        # validate r
+        r = self._validate_coordinate(r)
```
mphoward, line 52:
```suggestion
```

tests/model/potential/test_angle.py
```diff
@@ -223,12 +223,22 @@ def test_energy(self):
...
+        # test invalid theta less than zero
         theta_input = numpy.array([0.0, 0.5, 1.0])
         u_actual = numpy.array([500.0, 125.0, 0.0])
         u = harmonic_angle.energy(type_=("1"), theta=theta_input)
         numpy.testing.assert_allclose(u, u_actual)
 
+        # test invalid theta less than zero
+        theta_input = -1.0
+        with self.assertRaises(ValueError):
+            harmonic_angle.energy(type_=("1"), theta=theta_input)
```
mphoward, line 235: FYI: these probably all need a trailing comma
```suggestion
        with self.assertRaises(ValueError):
            harmonic_angle.energy(type_=("1",), theta=-1.0)
```
since otherwise the parentheses are ignored and it's like you've just passed the string. It works here because the string is interpreted as a list of characters, but would be a problem if you had multi-character types.

I would also just put the value of theta in the call, there isn't a reason to separate it out.

tests/simulate/test_simulate.py
```diff
@@ -43,6 +43,9 @@ def derivative(self, key, var, x):
             d = d.item()
         return d
 
+    def _validate_coordinate(self, x):
+        pass
```
mphoward, line 47: If you are implementing this as empty, it needs to return `x` based on the signature you're using
```suggestion
        return x
```

mphoward (changes requested):
Thanks! Overall this looks good, and it just needs some find-replace edits

tests/model/potential/test_angle.py
clpetix: Putting the trailing comma causes as an error since the key gets set as ("1",). 

>   File "/Users/levipetix/Documents/research/projects/software/relentless/tests/model/potential/test_angle.py", line 234, in test_energy
    harmonic_angle.energy(type_=("1",), theta=-1.0)
  File "/Users/levipetix/Documents/research/projects/software/relentless/src/relentless/model/potential/angle.py", line 117, in energy
    params = self.coeff.evaluate(type_)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/levipetix/Documents/research/projects/software/relentless/src/relentless/model/potential/potential.py", line 73, in evaluate
    if self[key][p] is not None:
       ~~~~^^^^^
  File "/Users/levipetix/Documents/research/projects/software/relentless/src/relentless/model/potential/potential.py", line 96, in __getitem__
    return self._data[key]
           ~~~~~~~~~~^^^^^
  File "/Users/levipetix/Documents/research/projects/software/relentless/src/relentless/collections.py", line 98, in __getitem__
    key = self._check_key(key)
          ^^^^^^^^^^^^^^^^^^^^
  File "/Users/levipetix/Documents/research/projects/software/relentless/src/relentless/collections.py", line 94, in _check_key
    raise KeyError("Key {} is not in dictionary.".format(key))
KeyError: "Key ('1',) is not in dictionary."

tests/simulate/test_simulate.py
clpetix: Given the change above to not have it pass the value back, I think this is now fine.

tests/model/potential/test_angle.py
mphoward: Oh sorry, you are right: there just shouldn't be any parentheses here!

src/relentless/model/potential/potential.py
```diff
@@ -344,6 +344,11 @@ def _zeros(cls, x):
             raise TypeError("Potential coordinate must be 1D array.")
         return x, numpy.zeros_like(x), s
 
+    @abc.abstractmethod
+    def _validate_coordinate(self, x):
+        """Validate the coordinate ``x`` is within the valid range."""
```
mphoward, line 349: This should get a proper docstring too, even though it is the base method.

src/relentless/model/potential/pair.py
```diff
@@ -474,6 +471,12 @@ def _derivative(self, param, r, **params):
         """
         pass
 
+    def _validate_coordinate(self, r):
+        """Validate the coordinate for pair potentials."""
```
mphoward, line 475: Fix up docstring

src/relentless/model/potential/pair.py
```diff
@@ -474,6 +471,12 @@ def _derivative(self, param, r, **params):
...
+        if numpy.any(r < 0):
+            raise ValueError("Pair distance must be non-negative")
+        return r
```
mphoward, line 478: This shouldn't return anymore
```suggestion
```

mphoward (approved):
Thanks! Please address these small things, then this can be merged.

clpetix:
@mphoward, I've applied those changes so this should be ready!

mphoward (approved):

PR #300, opened 2025-05-06
Fix direct ensemble average bugs

clpetix:
This addresses the two known issues mentioned in Issue #292.

### DIFF
--- a/src/relentless/optimize/objective.py
+++ b/src/relentless/optimize/objective.py
@@ -372,12 +372,11 @@ class RelativeEntropy(ObjectiveFunction):
             The result, which has unknown value ``None`` and known gradient.
 
         """
-        if self.thermo.rdf is None and not self._use_trajectory(
-            self.target, self.thermo
-        ):
-            raise ValueError(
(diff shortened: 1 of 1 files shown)


src/relentless/optimize/objective.py
```diff
@@ -372,12 +372,11 @@ def compute(self, variables, directory=None):
...
 
         """
-        if self.thermo.rdf is None and not self._use_trajectory(
-            self.target, self.thermo
-        ):
-            raise ValueError(
-                "EnsembleAverage needs to compute RDF, specify parameters."
-            )
+        if not self._use_trajectory(self.target, self.thermo):
+            if self.thermo.rdf is None:
```
mphoward, line 376: Just FYI: if you wanted this to stay as a one-liner, it would work if you reverse the order like this:
```suggestion
        if not self._use_trajectory(self.target, self.thermo) and self.thermo.rdf is None:
```
Because the first condition will be false if `self.thermo` has the wrong type, the second condition does not get checked. I'm fine leaving it the way you've written it though.

mphoward (approved):
Looks good, thanks!

Issue #301, opened 2025-05-13
Only do calculation of bonded contributions to the gradient if the tabulator has that type of potential


clpetix:
https://github.com/mphowardlab/relentless/blob/9a4a6f112178746f28011de27283701cf58357ea/src/relentless/optimize/objective.py#L601-L606

We should only try to calculate `self.potentials.bond is not None` for each type of bonded interactions. As it stands, this causes an issue if the GSD has bonds but the simulation does not, e.g., using an Einstein crystal as a target for self-assembly.

mphoward:
Yes, I agree. The same check should also be done for the pair potentials now that you have bonded interactions. For example, you could design the bonds in an ideal gas of dumbbells. Only enter the calculation for each type of interaction if its tabulator is not None.

PR #302, opened 2025-05-15
Only do gradient contribution calculation if the tabulator has interaction type

clpetix:
Resolves Issue #301.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,3 +48,12 @@ or from conda-forge using `conda`:
 Contributions are welcomed and appreciated! Fork and create a pull request on
 [GitHub](https://github.com/mphowardlab/relentless). We value the input and
 experiences all users and contributors bring to `relentless`.
+
+## Citing relentless
+
+When using `relentless` in published work, please cite:
+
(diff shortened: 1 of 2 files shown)


src/relentless/optimize/objective.py
```diff
@@ -585,21 +585,22 @@ def _calc_ensemble_average_dvar_gradient(self, trajectory, variables):
...
+                            for var in variables:
+                                gradient[var] += numpy.sum(
+                                    self.potentials.pair.derivative(
+                                        (i, j), var, x=neighbors.distances[filter_ij]
+                                    )
+                                )
 
                 # bond contributions to the gradient
-                if snap.bonds.N != 0:
+                if self.potentials.bond is not None:
```
mphoward, line 603: Do you want to do both checks? This code path only does something if both are true:
```suggestion
                if self.potentials.bond is not None and snap.bonds.N != 0:
```
Similarly for the angles and dihedrals

mphoward (commented):
Thanks! One small question to address. Should there be any unit tests to catch this?

mphoward:
@clpetix I realized this PR was still open. Do you need anything from me to finish it off?

Issue #303, opened 2025-07-02
MPI communicator needs to be wrapped in a HOOMD communicator when using HOOMD-blue


clpetix:
This communicator needs to be wrapped in a `hoomd.communicator.Communicator` object to prevent errors when running under MPI. 

https://github.com/mphowardlab/relentless/blob/c807258668ebbc02b0ada83a1758efcbee8bf54a/src/relentless/simulate/hoomd.py#L1394

PR #304, opened 2025-07-02
Wrap MPI communicator in HOOMD communicatory

clpetix:
Fixes Issue #303


mphoward:
Did you confirm this worked on Easley? If yes, it is a straightforward change so we can merge.

clpetix:
@mphoward, yes, this fixed the issue on Easley!

PR #307, opened 2025-08-27
Fix failing errors regarding lammpsio box. 

clpetix:
This addresses the errors first seen in PR #306.


mphoward (approved):
Makes sense to me! Do we know if something changed in `lammpsio` to trigger this?

mphoward:
Hmm, maybe some change in another dependency has caused this branch to trip now then... will merge!

PR #317, opened 2026-05-07
Add Adam Optimizer

clpetix:
This PR implements the Adam Optimizer. We've had enough success with this in several projects that I think it is worth adding it as an option!

### DIFF
--- a/src/relentless/optimize/__init__.py
+++ b/src/relentless/optimize/__init__.py
@@ -63,5 +63,5 @@ from .criteria import (
     Tolerance,
     ValueTest,
 )
-from .method import FixedStepDescent, LineSearch, Optimizer, SteepestDescent
+from .method import AdamOptimizer, FixedStepDescent, LineSearch, Optimizer, SteepestDescent
 from .objective import ObjectiveFunction, ObjectiveFunctionResult, RelativeEntropy
--- a/src/relentless/optimize/__init__.py
+++ b/src/relentless/optimize/__init__.py
@@ -22,6 +22,7 @@ Algorithms
     SteepestDescent
     FixedStepDescent
     LineSearch
+    AdamOptimizer
 
 Convergence criteria
 ====================
--- a/src/relentless/optimize/__init__.py
+++ b/src/relentless/optimize/__init__.py
@@ -22,7 +22,7 @@ Algorithms
     SteepestDescent
     FixedStepDescent
     LineSearch
-    AdamOptimizer
+    Adam
 
 Convergence criteria
 ====================
@@ -65,7 +65,7 @@ from .criteria import (
     ValueTest,
 )
 from .method import (
-    AdamOptimizer,
+    Adam,
     FixedStepDescent,
     LineSearch,
     Optimizer,
--- a/src/relentless/optimize/__init__.py
+++ b/src/relentless/optimize/__init__.py
@@ -63,5 +63,11 @@ from .criteria import (
     Tolerance,
     ValueTest,
 )
-from .method import AdamOptimizer, FixedStepDescent, LineSearch, Optimizer, SteepestDescent
+from .method import (
+    AdamOptimizer,
+    FixedStepDescent,
+    LineSearch,
+    Optimizer,
+    SteepestDescent,
+)
 from .objective import ObjectiveFunction, ObjectiveFunctionResult, RelativeEntropy
--- a/src/relentless/optimize/method.py
+++ b/src/relentless/optimize/method.py
@@ -591,3 +591,261 @@ class FixedStepDescent(SteepestDescent):
         for i in k:
             k[i] = self.step_size
         return k / gradient.norm()
+
+class AdamOptimizer(Optimizer):
+    r"""Adam optimization algorithm.
+
+    For an :class:`~relentless.optimize.objective.ObjectiveFunction`
+    :math:`f\left(\mathbf{x}\right)`, the Adam algorithm seeks to approach
+    a minimum of the function.
+
+    The optimization is performed using scaled variables :math:`\mathbf{y}`.
+    Define :math:`\mathbf{X}` as the scaling parameters for each variable such
+    that :math:`y_i=x_i/X_i`. (A variable can be left unscaled by setting
+    :math:`X_i=1`).
+
+    Adam maintains first and second moment estimates of the gradient to adapt
+    the step size for each parameter individually. Let :math:`\mathbf{g}_t`
+    be the gradient at iteration :math:`t`, and let :math:`\mathbf{m}_t` and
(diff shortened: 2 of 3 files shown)


src/relentless/optimize/method.py
```diff
@@ -591,3 +591,285 @@ def descent_amount(self, gradient):
...
+    max_iter : int
+        The maximum number of optimization iterations allowed.
+    step_size : float
+        The step size hyperparameter (:math:`\alpha`).
+    beta1 : float
+        The exponential decay rate for the first moment estimates
+        (defaults to ``0.9``).
+    beta2 : float
+        The exponential decay rate for the second moment estimates
+        (defaults to ``0.999``).
```
mphoward: I would prefer using a subscript for readability if you can push this change throughout:
```suggestion
    beta_1 : float
        The exponential decay rate for the first moment estimates
        (defaults to ``0.9``).
    beta_2 : float
        The exponential decay rate for the second moment estimates
        (defaults to ``0.999``).
```

tests/optimize/test_method.py
```diff
@@ -334,5 +334,169 @@ def test_run(self):
...
+
+        # test dictionary of scaling parameters
+        o.scale = {x: 0.3}
+        self.assertEqual(o.stop, t)
+        self.assertEqual(o.max_iter, 1000)
+        self.assertAlmostEqual(o.step_size, 0.25)
+        self.assertAlmostEqual(o.beta1, 0.9)
+        self.assertAlmostEqual(o.beta2, 0.999)
+        self.assertAlmostEqual(o.epsilon, 1e-8)
+        self.assertEqual(o.scale, {x: 0.3})
```
mphoward, line 369: This makes the intent of the test clearer:
```suggestion
        # test scalar scaling parameter
        o.scale = 0.5
        self.assertAlmostEqual(o.scale, 0.5)

        # test dictionary of scaling parameters
        o.scale = {x: 0.3}
        self.assertEqual(o.scale, {x: 0.3})
```

src/relentless/optimize/__init__.py
```diff
@@ -22,6 +22,7 @@
     SteepestDescent
     FixedStepDescent
     LineSearch
+    Adam
```
mphoward, line 25: Is it intentional to put this last in the list? I'm OK with that, just wanted to check.

src/relentless/optimize/method.py
```diff
@@ -591,3 +591,285 @@ def descent_amount(self, gradient):
...
+            {\sqrt{\hat{\mathbf{v}}_n}+\epsilon}
+
+    The gradient of the function with respect to the scaled variables is:
+
+    .. math::
+
+        \nabla f\left(\mathbf{y}\right) =
+            \left[X_1 \frac{\partial f}{\partial x_1},
+            \cdots,
+            X_n \frac{\partial f}{\partial x_n}\right]
```
mphoward: I would reorder like this. Additionally, can you please explicitly state that n starts at 1 (I assume this is the case)?
```suggestion
    :math:`X_i=1`). The gradient of the function with respect to the scaled variables is:

    .. math::

        \mathbf{g} = \nabla f\left(\mathbf{y}\right) =
            \left[X_1 \frac{\partial f}{\partial x_1},
            \cdots,
            X_n \frac{\partial f}{\partial x_n}\right]

    Define :math:`\alpha` as the descent step size hyperparameter. Adam
    iteratively minimizes the function by taking steps based on exponentially
    weighted first and second moment estimates of the gradient. Let
    :math:`\mathbf{g}_n` be the gradient at iteration :math:`n`, and let
    :math:`\mathbf{m}_n` and :math:`\mathbf{v}_n` be the first and second moment
    estimates. If the scaled variables are :math:`\mathbf{y}_n` at iteration
    :math:`n`, the next value of the variables is:

    .. math::

        \mathbf{m}_n &= \beta_1 \mathbf{m}_{n-1}
            + \left(1-\beta_1\right)\mathbf{g}_n \\
        \mathbf{v}_n &= \beta_2 \mathbf{v}_{n-1}
            + \left(1-\beta_2\right){\mathbf{g}_n}^2 \\
        \hat{\mathbf{m}}_n &= \frac{\mathbf{m}_n}{1-{\beta_1}^n} \\
        \hat{\mathbf{v}}_n &= \frac{\mathbf{v}_n}{1-{\beta_2}^n} \\
        \mathbf{y}_{n+1} &= \mathbf{y}_n-\alpha
            \frac{\hat{\mathbf{m}}_n}
            {\sqrt{\hat{\mathbf{v}}_n}+\epsilon}

```

src/relentless/optimize/method.py
```diff
@@ -591,3 +591,285 @@ def descent_amount(self, gradient):
         for i in k:
             k[i] = self.step_size
         return k / gradient.norm()
+
+
+class Adam(Optimizer):
+    r"""Adam optimization algorithm.
+
+    For an :class:`~relentless.optimize.objective.ObjectiveFunction`
+    :math:`f\left(\mathbf{x}\right)`, the Adam algorithm seeks to approach
```
mphoward: Is there a link to a reference on this? A paper and/or one of the standard ML frameworks?

mphoward (changes requested):
Looks great! Here are a couple minor requests.

src/relentless/optimize/__init__.py
clpetix: Yes, I thought it made sense that FixedStep followed SteepestDescent and it didn't make sense for Adam to be before regular steepest descent.

src/relentless/optimize/method.py
clpetix: Added a link to the arxiv paper presenting Adam!

clpetix:
@mphoward, whenever you have a chance could you look at these failing jobs and see if you have any ideas? I don't really understand what is happening but I think it has to just be a GitHub actions issue 😅

mphoward:
It is surprisingly hard to spot, but based on this error:
```
Defaulting to user installation because normal site-packages is not writeable
```
I suspect that the problem is that a different `pip` is being used (ex: the system one), and so the tests are choking on those dependencies. I am not sure if `relentless` is even installed in the test environment!

I'm not sure why this is broken all of a sudden. As a quick fix, you could try explicitly adding `pip` as the last package installed when HOOMD / LAMMPS are installed via conda:

https://www.anaconda.com/docs/getting-started/working-with-conda/packages/pip-install#using-pip-install-in-a-conda-environment

If that doesn't work, I have been considering moving to `pixi` because it is able to solve conda-forge and PyPI dependencies simultaneously (even on my desktop). That will be more work though.

clpetix:
I've addressed all your comments, and this should be about ready to go!

mphoward (approved):
Thank you, looks great!

PR #328, opened 2026-07-21
Enforce SPEC-0 for dependency support

### LINKED ISSUE #321: Enforce SPEC 0
We started this discussion here:

https://github.com/mphowardlab/relentless/issues/267

but never followed up on it. I am now fully supportive of strictly adopting SPEC 0:

https://scientific-python.org/specs/spec-0000/

Maintaining backwards compatibility with Python is less of an issue given how many different package managers are available for installing specific versions. It is also a lot of work to test so many combinations and make sure they work.

SPEC 0 would mean:
* python >= 3.12
* freud >= 3.2
* gsd >= 3.3
* networkx >= 3.4
* numpy >= 2.1
* scipy >= 1.15
* HOOMD >= 4.9

LAMMPS is trickier because of how they name their packages. The LAMMPS conda-forge recipe also has not been updated in a long time, so I am not sure how we will test modern versions. Hence, we may leave LAMMPS pinned at the versions we currently support.

mphoward:
Fixes #321

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v7
         with:
-          python-version: "3.9"
+          python-version: "3.14"
       - name: Install dependencies
         run: |
           python -m pip install pip --upgrade pip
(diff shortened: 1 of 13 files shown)


clpetix (approved):

mphoward:
Merging to avoid branch getting stale, as dependencies are self-imposed / documented. We can add some more notes later if we like!

